Issue #34, opened 2026-08-01
TypeScript: port R004 (tools/list order is not deterministic)


dheerajjha:
Port **R004** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** tools/list order is not deterministic (`advisory`)
**Python implementation to mirror:** [`r004_tool_ordering.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r004_tool_ordering.py)
**Spec:** Deterministic tool ordering (SHOULD)

### What makes this one specific

`server.setRequestHandler(ListToolsRequestSchema, ...)` and `server.tool(...)` registrations. Check whether the returned array is sorted.

### How to do it

1. Add `languages = ("python", "typescript")` to the rule class.
2. Split `check()` on `project.language`, exactly like the two reference ports:
   - [`r001_session_id_removed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r001_session_id_removed.py) — identifier match *and* a string-literal match anchored to a header access
   - [`r006_sse_transport_deprecated.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r006_sse_transport_deprecated.py) — the simpler shape
3. Pick your search mode deliberately. This is the single easiest thing to get wrong:
   - `search_code()` — skips comments **and** strings. For code identifiers.
   - `search_wire()` — skips comments and docstrings, **keeps** string literals. For removed JSON-RPC method names, which only ever appear inside strings.
   - `search()` — raw, matches prose too. Almost never right.
4. Add tests to `tests/test_typescript.py`: one that fires on a legacy server, one that stays silent on a migrated one, and one that stays silent when the thing is only *named in a comment*.

### Done when

`mcp-migrate check` reports R004 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

Ask in the thread if anything is unclear — happy to talk it through before you write code, and PRs get reviewed the same day.

atiqur-rahman-pro:
Like to take R004!

dheerajjha:
Closing the loop on this, late: @atiqur-rahman-pro claimed this on 1 August and I never replied. They did it anyway — #61, merged — and have since taken #65, the brace-depth follow-up.

[...]

PR #51, opened 2026-08-01
feat(rules): port R003 (MissingRoutingHeaders) to TypeScript (fixes #33)

### LINKED ISSUE #33: TypeScript: port R003 (Custom HTTP client does not send Mcp-Method (or Mcp-Name where required))
Port **R003** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Custom HTTP client does not send Mcp-Method (or Mcp-Name where required) (`advisory`)
**Python implementation to mirror:** [`r003_routing_headers.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r003_routing_headers.py)

Look for `fetch(`/`axios(`/`got(` calls with a `headers` object that omits `Mcp-Method`. The Python rule requires MCP protocol surface in the same file before it fires -- keep that guard, it is what stopped this rule reporting every REST call in mcp-atlassian.

### How to do it

### Done when

`mcp-migrate check` reports R003 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

atiqur-rahman-pro:
## What kind of PR is this?

- [ ] Something else

## Cookbook recipe

## Rule

## Fixer

      nothing the second time

## Board entry

## Anything else reviewers should know?

Ported R003 (`MissingRoutingHeaders`) to TypeScript.

- Added `languages = ("python", "typescript")` to `MissingRoutingHeaders` rule class.
- Implemented `_check_ts` using `search_wire` to detect unheadered `fetch( / axios( / got(` calls in code while ignoring comments/docstrings.
- Gated checks on TypeScript files referencing MCP protocol surface (`@modelcontextprotocol`, `tools/call`, `tools/list`, `resources/read`, `prompts/get`, `jsonrpc`).
- Added unit tests in `tests/test_typescript.py` covering legacy client detection, migrated client silence, and comment-only mentions.

Fixes #33

### DIFF
--- a/src/mcp_migrate/rules/r003_routing_headers.py
+++ b/src/mcp_migrate/rules/r003_routing_headers.py
@@ -26,6 +26,11 @@ MCP_METHOD_RX = re.compile(
 # "missing Mcp-Name" was the bug: it isn't missing, it was never required.
 NAME_REQUIRED_RX = re.compile(r"tools/call|resources/read|prompts/get")
 
+TS_HTTP_CALL = r"\b(?:fetch|axios|got)\s*\(|\.(?:post|request)\s*\("
+TS_MCP_SURFACE_RX = re.compile(
+    r"@modelcontextprotocol|tools/call|tools/list|resources/read|prompts/get|jsonrpc",
+    re.IGNORECASE,
+)
 
 # Submodules that are still "the SDK owns the transport entirely", the same
 # way `mcp.server.fastmcp` is -- importing them is configuration of the
@@ -93,8 +98,14 @@ class MissingRoutingHeaders(Rule):
         "is tools/call, resources/read, or prompts/get -- those are the only methods "
         "that carry a name to route on. Or move to an SDK that sets both for you."
     )
+    languages = ("python", "typescript")
 
     def check(self, project: Project) -> list[Finding]:
+        if project.language == "typescript":
+            return self._check_ts(project)
+        return self._check_python(project)
+
+    def _check_python(self, project: Project) -> list[Finding]:
         # Gate on the project actually importing a raw HTTP client -- without
         # that, ".post(" is almost certainly unrelated (an ORM, a queue, a
         # dataclass method) and flagging it would just be noise.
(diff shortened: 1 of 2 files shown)


dheerajjha (approved):
Merging this — thank you, and welcome. This is the first external contribution to the project.

A few things you did that I want to call out, because they weren't obvious from the issue:

**You kept the MCP-surface gate.** That's the part that actually matters in R003 and the easiest thing to drop when porting. Without it this rule flags every `fetch(`/`axios(` call in the codebase — an earlier version of the Python rule did exactly that and hit `mcp-atlassian` 19 times, costing it 475 penalty points on plain Jira REST calls. Your `TS_MCP_SURFACE_RX` check is the thing that stops that happening in TypeScript, and you carried it across without being told twice.

**You picked `search_wire` and were right.** `search_code` would have found nothing (the header names are string literals), plain `search` would have flagged the comment in your own third test. That three-way split is the least obvious part of the codebase and you got it first time.

**You noticed the coverage denominator.** Bumping `2 of 21` → `3 of 21` is exactly right and I didn't think to mention it in the issue. That number is load-bearing — it's what decides when TypeScript stops being "partial" and starts carrying a grade.

I also ran it against `browserbase/mcp-server-browserbase` as a false-positive check: no R003 findings, which is the correct answer for an SDK-based server.

---

**One apology.** If you ran the suite locally and saw `test_partial_coverage_is_stated_with_a_denominator` and `test_check_on_a_mixed_tree_declares_its_coverage` fail, **that was our bug, not yours.** Those tests assert on printed output, and `rich` colours it when `FORCE_COLOR` is set in your environment — the escape codes land inside the captured string, in the middle of the number, so `"3 of 21"` is really `"\x1b[1;36m3\x1b[0m of ..."`. It's invisible in CI and reproducible on a laptop, which is the worst possible place for it. Fixed on main in 25eb714. Sorry if that cost you time.

---

Two things, if you're up for them:

1. **What annoyed you?** Setup, docs, the issue description, anything that took longer than it should have. I'd rather hear it bluntly and file it than have the next person hit the same thing.
2. **Want another?** 18 rules left, same shape: #34 (R004), #38 (R009), #40 (R011) are the closest to what you just did. #46 (R017) is a single literal if you want a quick one. Comment on whichever you take so nobody doubles up.

atiqur-rahman-pro:
@dheerajjha!- Thank you so much. You Honored me and I Really appreciated your detailed feedback and testing against Browserbase! Looking forward to making more contributions.

dheerajjha:
Late reply, but worth closing the loop: you have gone on to land #61 (R004) as well, and you have claimed #65 — the brace-depth follow-up — which is assigned to you with implementation notes.

One thing from that first exchange I should have said sooner. You commented on #34 asking to take R004 and I never answered; you went ahead and did it anyway twelve hours later. That worked out, but it should not have required you to guess. Claims get answered from here.

Two merged, one assigned. Thanks for sticking with it.

PR #52, opened 2026-08-01
feat(rules): port R005 (NoExtensionsDeclared) to TypeScript

### LINKED ISSUE #35: TypeScript: port R005 (Server capabilities declare no extensions map)
Port **R005** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Server capabilities declare no extensions map (`advisory`)
**Python implementation to mirror:** [`r005_extensions.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r005_extensions.py)
**Spec:** extensions field on ServerCapabilities

`ServerCapabilities` in the SDK's TS types, and the capabilities object passed to `new Server({...}, {capabilities: {...}})`. Careful: match the SDK's own names, not a bare `capabilities` -- that identifier is far too common, and it was a 91% false-positive rate on the Python side before it was tightened.

### How to do it

### Done when

`mcp-migrate check` reports R005 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

MasRama:
## What kind of PR is this?

- [x] Rule — adds/changes a rule in `src/mcp_migrate/rules/`

## Rule

Ports R005 to TypeScript, following the pattern established by R001/R003/R006.

Two detection signals:

1. **Explicit `ServerCapabilities` type reference** — `search_code` skips comments and string literals, so a JSDoc mention or a string is not treated as a declaration.
2. **`capabilities: {}` property in a file that imports `@modelcontextprotocol/sdk`** — the common TS pattern where the type is inferred from the `Server` constructor and `ServerCapabilities` never appears as a name. Gated on an SDK import to keep the bare `capabilities` false-positive rate down (91% on the Python side before it was tightened). Uses `search_wire` so both identifier and string-key forms are caught.

Both signals check the file for `extensions` and stay silent if it appears anywhere, same as the Python rule. The `seen_files` set prevents duplicate findings when both signals match the same file.

Tests added to `tests/test_typescript.py`:
- fires on a TS server with `ServerCapabilities` type, no `extensions`
- silent on a migrated server that declares `extensions: {}`
- silent when `ServerCapabilities` is only named in a comment
- fires on the constructor pattern (`new Server` + `capabilities: {}` + SDK import, no type reference)
- silent on a non-MCP file with a bare `capabilities: {}` property

The partial-coverage denominator test is updated from 3 to 4 of 21, since R005 now reads TypeScript.

Closes #35

### DIFF
--- a/src/mcp_migrate/rules/r005_extensions.py
+++ b/src/mcp_migrate/rules/r005_extensions.py
@@ -18,6 +18,21 @@ from .base import Finding, Project, Rule
 CAPS_RX = re.compile(r"\bServerCapabilities\b|\bserver_capabilities\b")
 EXTENSIONS_RX = re.compile(r"\bextensions\b")
 
+# --- TypeScript -----------------------------------------------------------
+#
+# ServerCapabilities is the SDK's own type name in TypeScript, same as in
+# Python. The other place capabilities are declared is the options object
+# passed to `new Server({...}, {capabilities: {...}})`, where the type is
+# inferred from the constructor and ServerCapabilities never appears as a
+# name. A bare `capabilities` identifier is far too common to match on its
+# own (91% false-positive rate on the Python side before it was tightened),
+# so the constructor signal is gated on an @modelcontextprotocol/sdk import.
+TS_CAPS_TYPE_RX = r"\bServerCapabilities\b"
+TS_SDK_IMPORT_RX = re.compile(r"@modelcontextprotocol/sdk")
+TS_CAPS_PROP_RX = r"\bcapabilities\s*:\s*\{"
+
+MESSAGE = "Capabilities are declared but `extensions` is absent."
+
 
 class NoExtensionsDeclared(Rule):
     id = "R005"
@@ -28,8 +43,14 @@ class NoExtensionsDeclared(Rule):
         "Optional features now negotiate through `extensions` on ServerCapabilities. "
         "Declare an empty map if you support none -- it tells clients you speak 2026-07-28."
     )
+    languages = ("python", "typescript")
 
     def check(self, project: Project) -> list[Finding]:
+        if project.language == "typescript":
+            return self._check_ts(project)
+        return self._check_python(project)
+
+    def _check_python(self, project: Project) -> list[Finding]:
         out: list[Finding] = []
         seen_files = set()
         # search_code: a comment/docstring that merely mentions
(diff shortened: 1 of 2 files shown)


dheerajjha (approved):
Merging — thank you, and welcome. Second external contribution to the project.

You did the thing I was most worried someone would miss on this particular rule. R005 has the worst history in the codebase: on the Python side `capabilities\s*=` had no word boundary and no MCP context, so it fired on an LLM backend's class attribute, an OpenAI-compat shim's `required_capabilities=`, and a plugin registry's `update_capabilities`. **91% false positives** — 56 servers flagged, 5 real. Porting that naively to TypeScript, where `capabilities: {` is an even more ordinary thing to write, would have been worse.

Gating signal 2 on an `@modelcontextprotocol/sdk` import is exactly the right call, and your `test_r005_stays_silent_on_non_mcp_typescript` is the test that proves it. That's the one I'd have asked for if it hadn't been there.

Two signals rather than one is also correct and I hadn't thought it through when I wrote the issue: in TypeScript the type is usually inferred from the `Server` constructor, so `ServerCapabilities` genuinely never appears as a name in a lot of real servers. Signal 1 alone would have missed most of them.

**Verified against a real server before merging.** `browserbase/mcp-server-browserbase` now reports R005 at `src/index.ts:134`:

```ts
const server = new McpServer({
  name: "Browserbase MCP Server",
  ...
  capabilities: {
    resources: { subscribe: true, listChanged: true },
    tools: {},
  },
});
```

Real `McpServer`, real capabilities declaration, no `extensions` key — a true positive that the Python-only version could never have found. That's the first finding TypeScript support has produced on a real repo beyond R001.

190 tests pass. TypeScript coverage is now 4 of 21.

---

Two small things, both non-blocking and neither worth another round trip — I'll tidy them:

- The `r005_extensions` import sits after `r006_...` in the test file, so it's out of alphabetical order.
- A couple of extra blank lines around the new test section.

If you want another: #34 (R004), #38 (R009) and #40 (R011) are the closest in shape to what you just did. **#40 has the same trap you just handled well** — `ping` appears in health-check endpoints constantly, so it needs the same kind of gating.

And the same question I asked the last contributor: what annoyed you? Setup, the issue description, anything that took longer than it should have. I'd rather file it than have the next person hit it.

MasRama:
Honestly the issue description was great — the "done when" checklist and the two reference ports made the shape of the thing obvious. The one thing that tripped me up: `test_partial_coverage_is_stated_with_a_denominator` hardcodes the coverage count (`3 of 21`). I did not know about it until the test failed after adding `typescript` to the rule's `languages`. Not a big deal, but it is the kind of thing that is easy to miss if someone is only running the tests for their new rule and not the full suite. Might be worth computing that count dynamically or at least mentioning it in the issue template for port tasks.

Thanks for the pointers — I will take #40 next. The `ping` / health-check overlap sounds like the same gating problem, so I will be careful with it.

dheerajjha:
I owe you a reply and an acknowledgement, eleven hours late.

You wrote this on 1 August:

> The one thing that tripped me up: `test_partial_coverage_is_stated_with_a_denominator` hardcodes the coverage count (`3 of 21`)... Might be worth computing that count dynamically or at least mentioning it in the issue template for port tasks.

**You were exactly right, and I did not answer.** What happened next is the part worth telling you.

Over the following day, eleven port PRs landed. **Every single one** of them bumped that literal from `"4 of 21"` to `"5 of 21"` — individually correct, collectively impossible. The first merges and `main` sits at 5; the second asserts 5 when the real count is 6, `main` goes red, and the contributor who merged second gets a failure they did not cause. It turned a set of genuinely independent tasks into a queue where everyone but the first person is punished.

I hit it this morning while merging and wrote it up as my own design bug. It was, but you had already found it, diagnosed it, and proposed the fix — *compute the count dynamically* — before a single collision happened. The fix I shipped in `61b0782` is your suggestion, close to verbatim:

```python
```

That report was worth more than the port it came attached to, and it sat unread while the exact failure you predicted played out eleven times. Sorry.

Your second suggestion — putting it in the issue template for port tasks — is now moot for this specific line, since nobody needs to touch it again. But the general point stands and I am taking it: any test that encodes a count of things contributors add is a trap, and I will look for others.

You also said you would take #40 (`ping`), and you did — #53 is merged, and the `/ping` health-check test in it is the best false-positive guard anyone has written here. Thank you for both halves of this.

PR #53, opened 2026-08-01
feat(rules): port R011 (PingRemoved) to TypeScript

### LINKED ISSUE #40: TypeScript: port R011 (Implements the removed ping request/response)
Port **R011** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Implements the removed ping request/response (`breaking`)
**Python implementation to mirror:** [`r011_ping_removed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r011_ping_removed.py)

`PingRequestSchema` and the literal `"ping"` as a method name. Beware: `ping` appears in health-check endpoints constantly, so require MCP context.

### How to do it

### Done when

`mcp-migrate check` reports R011 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

MasRama:
## What kind of PR is this?

- [x] Rule — adds/changes a rule in `src/mcp_migrate/rules/`

## Rule

Ports R011 to TypeScript, following the pattern from R001/R003/R005/R006.

Two detection signals:

1. **`PingRequest` identifier** — `\bPingRequest\w*` via `search_code`. Broadened from the Python `\bPingRequest\b` because the TS SDK exports `PingRequestSchema` (Zod) alongside the `PingRequest` type. No MCP context gate needed — the identifier is distinctive enough that it essentially never appears outside MCP code.

2. **`"ping"` as a JSON-RPC method dispatch** — `method === "ping"`, `case "ping":`, `{"ping": ...}` via `search_code`. Gated on MCP context (SDK import or MCP method names in the file) because `ping` is one of the most overloaded tokens in server code. The dispatch patterns handle TS syntax: `===` operator, template literals, and `switch case`. The `{`/`,` anchor trick keeps the match start on the code side of the string literal so `search_code` does not drop it.

The gate is the key part: a `/ping` health-check route in a file that also imports the MCP SDK stays silent, and a `message === "ping"` comparison in a file with no MCP context stays silent too.

Tests added to `tests/test_typescript.py`:
- fires on `PingRequestSchema` import + usage (2 findings)
- fires on `method === "ping"` dispatch with MCP context
- fires on `case "ping":` dispatch with MCP context
- silent on a migrated server
- silent when `PingRequest` is only named in a comment
- silent on a `/ping` health-check route in a file with MCP context
- silent on `message === "ping"` in a file without MCP context

The partial-coverage denominator test is updated from 4 to 5 of 21.

Closes #40

### DIFF
--- a/src/mcp_migrate/rules/r011_ping_removed.py
+++ b/src/mcp_migrate/rules/r011_ping_removed.py
@@ -3,8 +3,10 @@ import re
 from .base import Finding, Project, Rule
 
 # `PingRequest` is the MCP SDK's own model name -- distinctive, essentially
-# never appears outside MCP code.
-PING_CODE_RX = re.compile(r"\bPingRequest\b")
+# never appears outside MCP code. In TypeScript the SDK exports
+# `PingRequestSchema` (Zod) and `PingRequest` (inferred type), so we match
+# any identifier starting with PingRequest.
+PING_CODE_RX = re.compile(r"\bPingRequest\w*")
 
 # A bare `"ping"` string is one of the most overloaded tokens in server
 # code (health checks, keepalives, load-balancer probes have nothing to do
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged — thanks, this is the strongest port yet.

The reason I say that: `test_r011_stays_silent_on_health_check_endpoint`. `app.get("/ping", ...)` in a file that *also* imports the MCP SDK is the exact case that would have made this rule useless — every Express server on earth has that route, and the SDK import means a naive context gate would not save you. You wrote the test for it before anyone asked. That is the difference between a port and a correct port.

Gating signal 2 on `TS_MCP_SURFACE_RX` while leaving signal 1 ungated is also the right call, and matches how R005 handles the same problem.

One thing to flag, not to fix — you also widened the **Python** pattern:

```python
```

That is correct for TypeScript, where the SDK exports `PingRequestSchema`. On the Python side it now also matches `PingRequestHandler`, `PingRequestParams`, and anything else prefixed that way, and there is no Python test covering the change. It is almost certainly harmless — I could not find a Python false positive for it in the 620-server scan — but it is a Python behaviour change riding along in a TypeScript PR, and I would rather say so out loud than let it land silently.

If you want another: **#40 is done, so R007 (#36) or R009 (#38)** are the interesting ones left. R009 is the one where a docstring was being reported as *breaking*, so it has the same "a mention is not a use" shape you just handled well here.

Issue #50, opened 2026-08-01
TypeScript: port R021 (Pins an older JSON Schema dialect than 2020-12)


dheerajjha:
Port **R021** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Pins an older JSON Schema dialect than 2020-12 (`advisory`)
**Python implementation to mirror:** [`r021_json_schema_2020_12_required.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r021_json_schema_2020_12_required.py)
**Spec:** SEP-2106 https://modelcontextprotocol.io/specification/2026-07-28/changelog

### What makes this one specific

`$schema` pins naming a JSON Schema draft older than 2020-12, typically in a zod-to-json-schema config or a literal schema object.

### How to do it

1. Add `languages = ("python", "typescript")` to the rule class.
2. Split `check()` on `project.language`, exactly like the two reference ports:
   - [`r001_session_id_removed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r001_session_id_removed.py) — identifier match *and* a string-literal match anchored to a header access
   - [`r006_sse_transport_deprecated.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r006_sse_transport_deprecated.py) — the simpler shape
3. Pick your search mode deliberately. This is the single easiest thing to get wrong:
   - `search_code()` — skips comments **and** strings. For code identifiers.
   - `search_wire()` — skips comments and docstrings, **keeps** string literals. For removed JSON-RPC method names, which only ever appear inside strings.
   - `search()` — raw, matches prose too. Almost never right.
4. Add tests to `tests/test_typescript.py`: one that fires on a legacy server, one that stays silent on a migrated one, and one that stays silent when the thing is only *named in a comment*.

### Done when

`mcp-migrate check` reports R021 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

Ask in the thread if anything is unclear — happy to talk it through before you write code, and PRs get reviewed the same day.

PR #54, opened 2026-08-02
fix: enable R017 for TypeScript

### LINKED ISSUE #46: TypeScript: port R017 (Uses the old -32002 resource-not-found error code)
Port **R017** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Uses the old -32002 resource-not-found error code (`breaking`)
**Python implementation to mirror:** [`r017_resource_not_found_code_changed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r017_resource_not_found_code_changed.py)

The literal `-32002`. Mechanical, and probably the easiest one here.

### How to do it

### Done when

`mcp-migrate check` reports R017 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

PuvaanRaaj:
## Summary

- enable rule R017 for TypeScript projects
- add positive, migrated, comment-only, and unrelated-number regression coverage
- update the reported TypeScript rule-coverage count

Fixes #46

## Validation

- `uv run --extra dev pytest tests/test_typescript.py -v` (26 passed)
- `uv run --extra dev pytest -v` (194 passed)
- `uv run mcp-migrate rules`
- `git diff --check`

## AI assistance

Prepared with GPT-5.6 assistance and reviewed against the existing Python rule and TypeScript scanning behavior.

### DIFF
--- a/src/mcp_migrate/rules/r017_resource_not_found_code_changed.py
+++ b/src/mcp_migrate/rules/r017_resource_not_found_code_changed.py
@@ -22,6 +22,7 @@ class ResourceNotFoundCodeChanged(Rule):
         "The resource-not-found error code changed from -32002 to -32602 (Invalid "
         "params). Update whatever raises or checks for -32002 in this context."
     )
+    languages = ("python", "typescript")
 
     def check(self, project: Project) -> list[Finding]:
         out: list[Finding] = []
--- a/tests/test_typescript.py
+++ b/tests/test_typescript.py
@@ -25,6 +25,9 @@ from mcp_migrate.rules.r003_routing_headers import MissingRoutingHeaders
 from mcp_migrate.rules.r004_tool_ordering import NondeterministicToolOrder
 from mcp_migrate.rules.r005_extensions import NoExtensionsDeclared
 from mcp_migrate.rules.r006_sse_transport_deprecated import DeprecatedSSETransport
+from mcp_migrate.rules.r017_resource_not_found_code_changed import (
+    ResourceNotFoundCodeChanged,
+)
 from mcp_migrate.rules.r011_ping_removed import PingRemoved
 from mcp_migrate.rules.r015_result_type_required import RequiredResultTypeMissing
 from mcp_migrate.rules.r016_cacheable_result_required import (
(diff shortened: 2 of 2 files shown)


dheerajjha:
R017, reviewed and verified locally — this is ready as far as the code goes.

You made the right call declaring `languages = ("python", "typescript")` and writing **no** `_check_ts` at all. A JSON-RPC error code is a bare numeric literal in every language, so the existing scan already works and a TypeScript branch would have been duplicated logic pretending to be a port. Several of the other rules genuinely do need a split; this one does not, and spotting the difference is the harder skill.

`test_r017_ignores_unrelated_typescript_negative_number` is the test I would have asked for — `TIMEOUT_SENTINEL = -32002` is exactly how this rule would start crying wolf.

---

```python
```

dheerajjha:
Thanks for these — and for rebasing all seven. **#55 (R019), #57 (R012) and #59 (R018) are merged.** TypeScript coverage went from 9 of 21 to 14 of 21 this morning and three of those five are yours.

This one I can't take yet, and Codex was right about why. I reproduced it:

```ts
  // old -32002 resource not found, migrated to -32602
  return { code: -32602 };
}
```

R017 reports a `breaking` finding on line 2. That file is correct — it's already on `-32602` and the old code appears only in prose.

The cause is subtle and worth knowing, because it will bite any rule written this way. `RESOURCE_NOT_FOUND_RX` is built **entirely of lookaheads**, so the regex consumes nothing and `match.start()` is always `0`. `search_wire` then asks "is column 0 inside a comment?" — and column 0 is `  ` (indentation), which is code. The comment starts at column 2, so the check passes and the finding escapes. Your test only covers a `//` at column 0, which is exactly the case that happens to work.

Two ways out. Either anchor the match on something that actually consumes the token you care about:

```python
RESOURCE_NOT_FOUND_RX = re.compile(r"-32002\b")
```

and keep the surrounding conditions as separate lookaheads applied to the line — or, if you need the lookahead form, report at the column of the real signal rather than the match start.

I'd take the first. Add the indented case to the test and I'll merge it same day. This is a good catch on a real sharp edge, not a knock on the port.

dheerajjha:
@PuvaanRaaj — checking in rather than quietly reassigning, because I'd rather ask than assume.

You have four PRs open — #54 (R017), #56 (R021), #58 (R013), #60 (R014) — and they've been sitting for two days since my review. All four now conflict with `main`, which isn't your doing: the TypeScript backend moved underneath them when v0.1.3 shipped.

**Three of your ports are already merged and in that release** — R012, R018 and R019 went out in v0.1.3 yesterday and are credited to you in the changelog. So this isn't a "your work isn't wanted" message.

The four open ones each have one verified false positive that I reproduced on your branches. Short version, on #54's shape:

```ts
  // old -32002 resource not found, migrated to -32602
  return { code: -32602 };
}
```

Reports `breaking`. The file is already correct — the old code appears only in a comment explaining the migration.

**What I'd like to know is simply whether you still want them.** Three honest options:

1. **You keep them.** Say so and I'll leave them assigned to you with no deadline. If the rebase is the annoying part, say that too and I'll do the conflict resolution for you and push to your branches — the conflicts are all in `tests/test_typescript.py` and I've resolved that same file a dozen times this week.
2. **You keep some.** Pick the ones you want, I'll close the rest and reopen the issues.
3. **You'd rather move on.** Completely fine, no explanation needed. I'll close them and note on the issues that the approach was sound and where the false positive is, so whoever picks them up starts from your work rather than from scratch.

The only outcome I want to avoid is the fourth one, where they sit open indefinitely and #42, #43, #46 and #50 stay blocked because nobody can tell whether they're live.

**No reply within a few days and I'll take option 3** — close them with credit and reopen the issues. Not a deadline, just so you know what silence means rather than having to guess.

dheerajjha:
@PuvaanRaaj — closing **this one only**. #56, #58 and #60 stay open and are still wanted.

#46 (R017 → TypeScript) ended up with three PRs against it: yours, @li2631026381-alt's #74, and @waterlemonnn's #122. I merged #122 this morning. Yours and #74 had been `CONFLICTING` since `main` moved under them.

**Your diagnosis was right and it was the same one that merged.** You worked out that R017 needs no TypeScript branch at all — the qualifying context is a numeric literal near ordinary English, not a language-specific identifier — so enabling the language is the whole change:

```python
languages = ("python", "typescript")
```

That is exactly what #122 does. It arrived three days later, with tests, against a `main` your branch had fallen behind. That's the only difference between them.

**The three that are still live matter more than this one did.** R013, R014 and R021 are three of the four rules that still don't read TypeScript — the whole remaining gap is `R002, R013, R014, R021`, and you have PRs open for three of them. Nobody else is working on any of them; I checked the open PR bodies for `Closes #N`, not just assignees.

They need a rebase onto current `main`, which has moved a long way this week (v0.1.3 shipped, and `main` is at 377 tests). **I'll do the rebase for you if you want** — say the word here and I'll rebase all three onto `main`, push to your fork, and leave your commits and authorship intact. Or if you'd rather do it yourself, that's completely fine.

No rush and no deadline attached to this. You have three merged PRs here and I'd rather wait than lose the last three ports to a stale branch.

PR #55, opened 2026-08-02
fix: enable R019 for TypeScript

### LINKED ISSUE #48: TypeScript: port R019 (Uses removed tasks/list or the removed blocking tasks/result)
Port **R019** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Uses removed tasks/list or the removed blocking tasks/result (`breaking`)
**Python implementation to mirror:** [`r019_tasks_polling_replaces_blocking_result.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r019_tasks_polling_replaces_blocking_result.py)
**Spec:** SEP-2663 https://modelcontextprotocol.io/specification/2026-07-28/changelog

`tasks/list` and blocking `tasks/result` literals.

### How to do it

### Done when

`mcp-migrate check` reports R019 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

PuvaanRaaj:
## Summary

- enable rule R019 for TypeScript projects
- detect only the removed `tasks/list` and blocking `tasks/result` wire literals

Fixes #48

## Validation

- `uv run mcp-migrate rules`
- `git diff --check`

## AI assistance

Prepared with GPT-5.6 assistance and reviewed against the existing Python implementation and TypeScript comment-aware scanner.

### DIFF
--- a/src/mcp_migrate/rules/r019_tasks_polling_replaces_blocking_result.py
+++ b/src/mcp_migrate/rules/r019_tasks_polling_replaces_blocking_result.py
@@ -18,8 +18,24 @@ class TasksPollingReplacesBlockingResult(Rule):
         "tasks/get + tasks/update. Tasks itself moved out of core into the "
         "io.modelcontextprotocol/tasks extension -- declare it there instead."
     )
+    languages = ("python", "typescript")
 
     def check(self, project: Project) -> list[Finding]:
+        if project.language == "typescript":
+            # The removed task methods are JSON-RPC wire names, so they
+            # appear as string literals. search_wire keeps those literals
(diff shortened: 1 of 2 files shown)


dheerajjha:
R019, reviewed and verified locally.

`search_wire` is the correct choice and your comment says why in one line — `tasks/list` cannot be a bare identifier, so it only ever exists inside a string literal, and `search_code` would never see it. That is the distinction the whole three-way search split exists for and you applied it without being told.

---

```python
```

PR #56, opened 2026-08-02
fix: enable R021 for TypeScript

### LINKED ISSUE #50: TypeScript: port R021 (Pins an older JSON Schema dialect than 2020-12)

PuvaanRaaj:
## Summary

- enable rule R021 for TypeScript projects
- detect explicit legacy JSON Schema dialect pins while ignoring comments

Fixes #50

## Validation

- `uv run mcp-migrate rules`
- `git diff --check`

## AI assistance

Prepared with GPT-5.6 assistance and reviewed against the existing Python implementation and TypeScript comment-aware scanner.

### DIFF
--- a/src/mcp_migrate/rules/r021_json_schema_2020_12_required.py
+++ b/src/mcp_migrate/rules/r021_json_schema_2020_12_required.py
@@ -26,8 +26,19 @@ class OldJSONSchemaDialect(Rule):
         "outputSchema. If you pin an older draft explicitly, move it to 2020-12 (or "
         "drop the pin and let a modern validator pick the default)."
     )
+    languages = ("python", "typescript")
+
+    MESSAGE = "Pins an older JSON Schema dialect; 2026-07-28 requires 2020-12 support."
 
     def check(self, project: Project) -> list[Finding]:
+        if project.language == "typescript":
+            # A TypeScript schema pin is a string literal. search_wire keeps
+            # the literal but ignores comments that only describe the upgrade.
+            return [
+                self.finding(self.MESSAGE, f, line, text)
+                for f, line, text in project.search_wire(OLD_DIALECT_RX.pattern)
+            ]
+
         out: list[Finding] = []
         # Raw text, not search_code: a `$schema` dialect pin is always a
         # URL/date string (`"http://json-schema.org/draft-07/schema#"`,
@@ -35,8 +46,5 @@ class OldJSONSchemaDialect(Rule):
         # starts inside a STRING token -- search_code would silently never
         # find it (see the notifications/initialized note in r009).
         for f, line, text in project.search_wire(OLD_DIALECT_RX.pattern):
-            out.append(self.finding(
-                "Pins an older JSON Schema dialect; 2026-07-28 requires 2020-12 support.",
-                f, line, text,
-            ))
+            out.append(self.finding(self.MESSAGE, f, line, text))
         return out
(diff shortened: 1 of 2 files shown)


dheerajjha:
R021, reviewed and verified locally.

Same correct `search_wire` reasoning as #55 — a `$schema` dialect URL is a string literal by definition. Hoisting the message to `self.MESSAGE` so both branches share it is a good instinct; syf2211 did the same thing independently in #62.

---

```python
```

dheerajjha:
Thanks — and **#55, #57 and #59 are merged**, so three of your seven are in. Coverage went 9 of 21 → 14 of 21 this morning.

This one needs one more pass. Reproduced:

```ts
export const releaseDate = "2019-09-30";
export const schema = { $schema: "https://json-schema.org/draft/2020-12/schema" };
```

R021 fires `breaking` on line 1. That file pins the **correct** dialect — the only problem is that an unrelated date happens to start with `2019-09`.

`search_wire` deliberately keeps ordinary string literals (that's what makes it the right mode for wire values), so every date string in the project is in scope, and `OLD_DIALECT_RX`'s word boundary is happy to match the `2019-09` prefix of `2019-09-30`.

The fix is to require the match be a dialect, not a date — either the full URL form, or `2019-09` only when it isn't followed by another `-` and a day:

```python
OLD_DIALECT_RX = re.compile(
    r"json-schema\.org/draft[/-](?:0?[47]|2019-09)\b"
    r"|(?<![\d-])2019-09(?![-\d])"
)
```

…plus a `$schema`-context requirement if you want to be stricter. Worth a test with a bare ISO date in it, since that's the shape that'll show up in real servers.

A false positive is the one failure mode this project treats as worse than missing a finding — a wrong `breaking` verdict costs someone their badge. That's the only reason this is waiting.

dheerajjha:

dheerajjha:
Closing — superseded by #128, which merged. **This one was a genuine tie on behaviour**, so the tiebreak deserves stating.

Both produce identical output on the same corpus: fires on a `draft-07` pin in a `.ts` file, silent on a `2020-12` pin, silent on a comment-only mention. Neither catches an uppercase `DRAFT-07` — `OLD_DIALECT_RX` carries no `re.IGNORECASE` on the Python path either, so that gap predates both PRs and neither introduced it.

What decided it:

1. **#128 is smaller.** It adds `"typescript"` to `languages` and nothing else, on the reasoning that a `$schema` pin is always a URL string — never a language-specific identifier or wire name — so the existing `search_wire` pass already works unmodified. This PR adds an explicit `if project.language == "typescript"` branch that calls `search_wire(OLD_DIALECT_RX.pattern)` — the same call the Python path below it already makes. Correct, but a duplicated branch that has to be kept in sync with the one above it.
2. **Mergeability.** This has been `CONFLICTING` since 2026-08-05.

Your `MESSAGE` constant extraction was the right instinct and I would have taken that half on its own.

One thing this PR quietly demonstrates, worth flagging since it bit R017 for real: `search_wire(OLD_DIALECT_RX.pattern)` passes only the pattern string, so any flags on the compiled object are dropped. Harmless here because `OLD_DIALECT_RX` has none — but that exact line shape was #123.

PR #57, opened 2026-08-02
fix: enable R012 for TypeScript

### LINKED ISSUE #41: TypeScript: port R012 (Implements the removed logging/setLevel request)
Port **R012** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Implements the removed logging/setLevel request (`breaking`)
**Python implementation to mirror:** [`r012_logging_set_level_removed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r012_logging_set_level_removed.py)

`SetLevelRequestSchema` and the literal `logging/setLevel`.

### How to do it

### Done when

`mcp-migrate check` reports R012 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

PuvaanRaaj:
## Summary

- enable rule R012 for TypeScript projects
- detect the removed schema handler and `logging/setLevel` wire method

Fixes #41

## Validation

- `uv run mcp-migrate rules`
- `git diff --check`

## AI assistance

Prepared with GPT-5.6 assistance and reviewed against the existing Python rule and TypeScript scanning paths.

### DIFF
--- a/src/mcp_migrate/rules/r012_logging_set_level_removed.py
+++ b/src/mcp_migrate/rules/r012_logging_set_level_removed.py
@@ -5,6 +5,7 @@ from .base import Finding, Project, Rule
 # `SetLevelRequest`/`SetLevelRequestParams` are the MCP SDK's own model
 # names for this request -- distinctive, no false-positive risk.
 SET_LEVEL_CODE_RX = re.compile(r"\bSetLevelRequest\b|\bSetLevelRequestParams\b")
+TS_SET_LEVEL_CODE_RX = re.compile(r"\bSetLevelRequest\w*")
 
 
 class LoggingSetLevelRemoved(Rule):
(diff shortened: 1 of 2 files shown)


dheerajjha:
R012, reviewed and verified locally.

Splitting the messages into `CODE_MESSAGE`/`WIRE_MESSAGE` class attributes so the two branches cannot drift is good, and keeping the original comment about why `logging/setLevel` needs `search_wire` rather than deleting it on the way past is the kind of thing that keeps a codebase legible.

**One real gap, and it applies to #57 and #58 both.** You match only the `...Schema` suffix:

```python
TS_SET_LEVEL_CODE_RX = re.compile(r"\bSetLevelRequestSchema\b")
```

The TypeScript SDK exports *two* names per request: the Zod schema (`SetLevelRequestSchema`) and the inferred type (`SetLevelRequest`), plus `...Params` in places. A file annotating `let r: SetLevelRequest` is a genuine use of a removed feature and goes unreported. I checked against your branches:

```
PR #57  Schema suffix -> 1 finding    bare type -> 0    Params type -> 0
PR #58  Schema suffix -> 1 finding    bare type -> 0    Params type -> 0
```

@MasRama hit exactly this in #53 (merged) and solved it in one character:

```python
```

`\bSetLevelRequest\w*` and `\bSubscribeRequest\w*` / `\bUnsubscribeRequest\w*` would close it. Worth reading #53 before you rebase — it is the closest sibling to these two.

---

```python
```

PR #58, opened 2026-08-02
fix: enable R013 for TypeScript

### LINKED ISSUE #42: TypeScript: port R013 (Uses resources/subscribe or resources/unsubscribe, replaced by subscriptions/listen)
Port **R013** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Uses resources/subscribe or resources/unsubscribe, replaced by subscriptions/listen (`breaking`)
**Python implementation to mirror:** [`r013_subscriptions_replaced.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r013_subscriptions_replaced.py)

`SubscribeRequestSchema`, `UnsubscribeRequestSchema`, and the literals `resources/subscribe` / `resources/unsubscribe`.

### How to do it

### Done when

`mcp-migrate check` reports R013 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

PuvaanRaaj:
## Summary

- enable rule R013 for TypeScript projects
- detect removed subscription schemas and wire methods

Fixes #42

## Validation

- `uv run mcp-migrate rules`
- `git diff --check`

## AI assistance

Prepared with GPT-5.6 assistance and reviewed against the existing Python rule and TypeScript scanning paths.

### DIFF
--- a/src/mcp_migrate/rules/r013_subscriptions_replaced.py
+++ b/src/mcp_migrate/rules/r013_subscriptions_replaced.py
@@ -5,6 +5,7 @@ from .base import Finding, Project, Rule
 # `SubscribeRequest`/`UnsubscribeRequest` are the MCP SDK's own model names
 # -- distinctive, no false-positive risk.
 SUBSCRIBE_CODE_RX = re.compile(r"\bSubscribeRequest\b|\bUnsubscribeRequest\b")
+TS_SUBSCRIBE_CODE_RX = re.compile(r"\bSubscribeRequest\w*|\bUnsubscribeRequest\w*")
 
 
 class ResourceSubscriptionsReplaced(Rule):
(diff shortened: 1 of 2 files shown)


dheerajjha:
R013, reviewed and verified locally.

Nice detail: `\bSubscribeRequestSchema\b` does **not** double-count inside `UnsubscribeRequestSchema`, because the `b` of `Unsub` is a word character so the leading `\b` fails. That is easy to get wrong in the other direction and your test would have caught it.

**One real gap, and it applies to #57 and #58 both.** You match only the `...Schema` suffix:

```python
TS_SET_LEVEL_CODE_RX = re.compile(r"\bSetLevelRequestSchema\b")
```

The TypeScript SDK exports *two* names per request: the Zod schema (`SetLevelRequestSchema`) and the inferred type (`SetLevelRequest`), plus `...Params` in places. A file annotating `let r: SetLevelRequest` is a genuine use of a removed feature and goes unreported. I checked against your branches:

```
PR #57  Schema suffix -> 1 finding    bare type -> 0    Params type -> 0
PR #58  Schema suffix -> 1 finding    bare type -> 0    Params type -> 0
```

@MasRama hit exactly this in #53 (merged) and solved it in one character:

```python
```

`\bSetLevelRequest\w*` and `\bSubscribeRequest\w*` / `\bUnsubscribeRequest\w*` would close it. Worth reading #53 before you rebase — it is the closest sibling to these two.

---

```python
```

dheerajjha:
Thanks for this — **#55, #57 and #59 are merged**, three of your seven in, coverage now 14 of 21.

This one is the one I'd most like fixed, because it fires three times on a clean file. Reproduced:

```ts
class SubscribeRequester {}
export const ROUTE = "/resources/subscribers";
export const cache = new SubscribeRequestCache();
```

Three `breaking` findings, all wrong. Two separate causes:

**1. The `\w*` suffix is unbounded.** `SubscribeRequest\w*` is meant to catch `SubscribeRequestSchema` (the Zod schema is what servers actually write, and `\bSubscribeRequest\b` can't match inside it). But it also swallows `SubscribeRequester` and `SubscribeRequestCache`. Bound it to the names that exist:

```python
SUBSCRIBE_NAMES_RX = re.compile(
    r"\b(?:Un)?[Ss]ubscribeRequest(?:Params|Schema)?\b"
)
```

**2. The wire pattern has no end boundary.** `resources/subscribe` matches inside `"/resources/subscribers"`. `search_wire` keeps all string literals and gives you no JSON-RPC context, so an HTTP route reads the same as a method name. Needs `(?![\w/-])` on the end, or an anchor requiring the literal be the complete method.

Both are one-line changes and both deserve a test — the route case especially, since `/resources/subscribers` is a plausible REST endpoint for a server that has correctly migrated to `subscriptions/listen`.

dheerajjha:

dheerajjha:
Closing — superseded by #126, which merged. Unlike #60 (where you were right about something the merged PR got wrong), **here the measurement went the other way**, and it is worth saying plainly so the reasoning is on the record.

Both ports ran against the same corpus:

```ts
// r013_false.ts
class SubscribeRequester { send() {} }
const x = new UnsubscribeRequestBuilderFactory();
```

| | this PR | merged #126 |
|---|---|---|
| real `SubscribeRequestSchema` / `resources/subscribe` | finds | finds |
| `class SubscribeRequester` | **false positive** | silent |
| `UnsubscribeRequestBuilderFactory` | **false positive** | silent |

`TS_SUBSCRIBE_CODE_RX = r"\bSubscribeRequest\w*|\bUnsubscribeRequest\w*"` — the unbounded `\w*` is #87, which has been open the whole time. #126 uses a bounded `(?:Schema|Params)` suffix and cites #87 for why.

That matters more than usual for R013 because a false positive here tells someone to rip out a class that has nothing to do with resource subscriptions. **A false positive is worse than a missed finding** is the rule this repo runs on.

None of which is a knock on the work — `\w*` is the obvious way to write it, three other rules shipped with the same shape, and that is exactly why #87 exists as an issue rather than a note in someone's head. Your R012 port and two fixers are in 0.1.3.

PR #59, opened 2026-08-02
fix: enable R018 for TypeScript

### LINKED ISSUE #47: TypeScript: port R018 (Uses a server-initiated request replaced by Multi Round-Trip Requests)
Port **R018** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Uses a server-initiated request replaced by Multi Round-Trip Requests (`breaking`)
**Python implementation to mirror:** [`r018_multi_round_trip_replaces_server_initiated.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r018_multi_round_trip_replaces_server_initiated.py)
**Spec:** SEP-2322 https://modelcontextprotocol.io/specification/2026-07-28/changelog

`createMessage`, `listRoots`, `elicitInput` -- the server-initiated requests replaced by MRTR.

### How to do it

### Done when

`mcp-migrate check` reports R018 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

PuvaanRaaj:
## Summary
- enable R018 (Multi Round-Trip Requests) for TypeScript
- match specific legacy TypeScript SDK request schemas and legacy wire-method literals
- add positive, migrated, and comment-only coverage

Fixes #47

## Validation
- uv run --extra dev pytest tests/test_typescript.py -v (25 passed)
- uv run --extra dev pytest -v (193 passed)
- uv run mcp-migrate rules
- git diff --check

Assisted by GPT-5.6; implementation and validation were reviewed by the author.

### DIFF
--- a/src/mcp_migrate/rules/r018_multi_round_trip_replaces_server_initiated.py
+++ b/src/mcp_migrate/rules/r018_multi_round_trip_replaces_server_initiated.py
@@ -31,6 +31,15 @@ FEATURES_LITERAL = {
     r"notifications/elicitation/complete": "notifications/elicitation/complete",
 }
 
+# The TypeScript SDK exports Zod schemas for request handling. These names
+# are specific to the removed MCP server-initiated requests, unlike generic
+# callback names such as `createMessage` or `listRoots`.
+TS_FEATURES_CODE = {
+    r"\bListRootsRequestSchema\b": "Server-initiated roots/list",
+    r"\bCreateMessageRequestSchema\b": "Server-initiated sampling/createMessage",
+    r"\bElicitRequestSchema\b": "Server-initiated elicitation/create",
+}
+
 
 class MultiRoundTripReplacesServerInitiated(Rule):
     id = "R018"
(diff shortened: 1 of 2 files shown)


dheerajjha:
R018, reviewed and verified locally.

This comment is the reason I trust the rest of the PR:

```python
# These names are specific to the removed MCP server-initiated requests,
# unlike generic callback names such as `createMessage` or `listRoots`.
```

You worked out on your own that `createMessage` and `listRoots` are too generic to match bare and restricted yourself to the `...RequestSchema` forms. That is the same reasoning that made R005 usable after a 91% false-positive rate, and R007 was hand-audited for a near-identical mistake. Reaching it unprompted is the good sign here.

The `...Schema`-only concern I raised on #57 and #58 is much weaker for this rule precisely *because* the bare names are ambiguous — restricting to the schema suffix is defensible here in a way it is not there.

---

```python
```

PR #60, opened 2026-08-02
fix: enable R014 for TypeScript

### LINKED ISSUE #43: TypeScript: port R014 (Implements SSE resumability (Last-Event-ID / event redelivery))
Port **R014** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Implements SSE resumability (Last-Event-ID / event redelivery) (`breaking`)
**Python implementation to mirror:** [`r014_sse_resumability_removed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r014_sse_resumability_removed.py)

`Last-Event-ID` header and event-store/resumability plumbing on the SSE transport.

### How to do it

### Done when

`mcp-migrate check` reports R014 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

PuvaanRaaj:
## Summary
- enable R014 (SSE resumability removal) for TypeScript
- distinguish executable identifiers and Last-Event-ID wire-header accesses while ignoring comments
- add positive, migrated, and comment-only coverage

Fixes #43

## Validation
- uv run --extra dev pytest tests/test_typescript.py -v (25 passed)
- uv run --extra dev pytest -v (193 passed)
- uv run mcp-migrate rules
- git diff --check

Assisted by GPT-5.6; implementation and validation were reviewed by the author.

### DIFF
--- a/src/mcp_migrate/rules/r014_sse_resumability_removed.py
+++ b/src/mcp_migrate/rules/r014_sse_resumability_removed.py
@@ -1,4 +1,20 @@
+import re
+
 from .base import Finding, Project, Rule
+
+# --- TypeScript -----------------------------------------------------------
+
+# TypeScript servers commonly keep the header value in a camel-cased local
+# before passing it to their replay store. The identifiers are specific enough
+# to be code-only matches, while the wire name itself must be kept in a string
+# literal and is therefore matched only in a header access.
+TS_IDENT_RX = r"\blastEventId\b|\blast_event_id\b|\bLAST_EVENT_ID\b"
+TS_HEADER_RX = (
+    r"headers?\s*(?:\[|\.get\s*\(|\.set\s*\(|\.delete\s*\()\s*"
+    r"[\"'`]last-event-id[\"'`]"
+    r"|[\"'`]last-event-id[\"'`]\s*\]"
+    r"|setHeader\s*\(\s*[\"'`]last-event-id[\"'`]"
+)
 
 
 class SSEResumabilityRemoved(Rule):
@@ -11,8 +27,14 @@ class SSEResumabilityRemoved(Rule):
         "event store / replay logic -- a dropped connection is just a dropped connection "
         "now, the client issues a fresh request."
     )
+    languages = ("python", "typescript")
 
     def check(self, project: Project) -> list[Finding]:
+        if project.language == "typescript":
+            return self._check_ts(project)
+        return self._check_python(project)
+
+    def _check_python(self, project: Project) -> list[Finding]:
         out: list[Finding] = []
         # search_code: a comment noting "we don't support Last-Event-ID"
         # (like the comment_only_mentions fixture pattern for R001) isn't a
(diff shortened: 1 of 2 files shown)


dheerajjha:
R014, reviewed and verified locally.

The dedup on `(path, line)` matters more than it looks: `const lastEventId = req.headers.get("Last-Event-ID")` matches the identifier pattern *and* the header pattern on the same line, and without the `seen` set every such line would report twice. You anticipated it and said so in the comment.

I checked the case-sensitivity question, since the canonical HTTP spelling is `Last-Event-ID` and your pattern is lowercase:

```
lowercase  "last-event-id"        -> 1 finding
canonical  "Last-Event-ID"        -> 1 finding
.get("Last-Event-ID")             -> 1 finding
bare camelCase lastEventId        -> 1 finding
```

All four land, so no problem — I am noting it only because it is not obvious from reading the regex alone and someone will wonder later.

---

```python
```

dheerajjha:
Thanks — **#55, #57 and #59 are merged.** Three of seven in, coverage 9 of 21 → 14 of 21.

This one has a false positive I can reproduce:

```ts
const source = new EventSource("/feed");
source.onmessage = (event) => log(event.lastEventId);
```

R014 reports `breaking`. But `lastEventId` is a **standard `MessageEvent` property** in the DOM — that snippet is ordinary browser code consuming any SSE stream, with no MCP transport and no replay logic anywhere near it. Plenty of MCP servers ship a web client in the same repo.

The identifier alone isn't enough signal. It needs to co-occur with something that makes it about resumability — a `Last-Event-ID` header read, a replay/redelivery call, or an event-store lookup — rather than firing on every executable occurrence.

Codex also flagged the mirror-image gap, which I confirmed: the header read is **missed** when a framework accessor is used rather than a `headers` object.

```ts
const from = req.get("Last-Event-ID");        // Express — not detected
const from = c.req.header("Last-Event-ID");   // Hono   — not detected
```

That one is only a miss, so it's a nice-to-have, not a blocker — silence is an acceptable output here. The false positive is the blocker.

Tightening the identifier to require resumability context would fix the first and naturally make room for the second.

dheerajjha:

dheerajjha:
Closing this, and I want to be precise about why, because **this PR was measurably right about something the one that merged got wrong.**

@waterlemonnn opened an independent R014 port (#127) three days after yours. Rather than pick on seniority I ran both against the same corpus:

| input | your #60 | merged #127 |
|---|---|---|
| `const lastEventId = req.headers.get("last-event-id")` | finds | finds |
| `res.setHeader("last-event-id", cursor)` | finds | finds |
| `const headers = { "Last-Event-ID": cursor };` | **misses** | finds |
| `(lastEventID) => replay(lastEventID)` | **finds** | misses |

Two real differences, one each way. Your `TS_HEADER_RX` requires an access form (`headers[…]`, `.get(`, `setHeader(`), so a plain object-literal header map slips past. But you passed `flags=re.IGNORECASE` to the identifier search and #127 does not — so `lastEventID` with a capital D is a live false negative on `main` right now, and you had it right.

**I filed that as #143 with your name on it.** It is the same class of bug as #123: a flag that never reaches the search because only `.pattern` gets passed.

What decided it was mergeability, not quality. #56/#58/#60 have been `CONFLICTING` since 2026-08-05 and you have been offline since 2026-08-04, and I was not willing to hold a working port waiting on a rebase that might not come. If that reads as unfair, say so and I will reopen — the code is not gone.

You also have #56 and #58 in the same position; I have commented on both. **Your R012 port and both fixers shipped in 0.1.3, and you have four merged PRs here. This is not a brush-off — it is three ports arriving second by two days.**

PR #61, opened 2026-08-02
feat(rules): port R004 (NondeterministicToolOrder) to TypeScript (fixes #34)

### LINKED ISSUE #34: TypeScript: port R004 (tools/list order is not deterministic)

atiqur-rahman-pro:
…es #34)

## What kind of PR is this?

- [ ] Something else

## Cookbook recipe

## Rule

## Fixer

      nothing the second time

## Board entry

## Anything else reviewers should know?

Ported R004 (`NondeterministicToolOrder`) to TypeScript.

- Added `languages = ("python", "typescript")` to `NondeterministicToolOrder` rule class.
- Implemented `_check_ts` to scan TypeScript handler definitions (`ListToolsRequestSchema`, `listTools`, `tools/list`) and verify explicit tool sorting (`.sort()`, `.toSorted()`, `sortBy`, `orderBy`).
- Added unit tests in `tests/test_typescript.py` covering unsorted handler detection and sorted handler silence.
- Updated coverage fraction assertion to `5 of 21`.

Fixes #34

### DIFF
--- a/src/mcp_migrate/rules/r004_tool_ordering.py
+++ b/src/mcp_migrate/rules/r004_tool_ordering.py
@@ -46,6 +46,10 @@ def _body_bounds(lines: list[str], line_no: int) -> tuple[int, int]:
     return body_start, end
 
 
+TS_HANDLER_RX = re.compile(r"ListToolsRequestSchema|listTools|tools/list", re.IGNORECASE)
+TS_SORT_RX = re.compile(r"\.(?:sort|toSorted)\s*\(|\bsorted\s*\(|\bsortBy\b|\borderBy\b", re.IGNORECASE)
+
+
 class NondeterministicToolOrder(Rule):
     id = "R004"
     title = "tools/list order is not deterministic"
@@ -55,8 +59,14 @@ class NondeterministicToolOrder(Rule):
         "Sort the tools you return. Stable ordering lets clients cache and it lifts "
         "LLM prompt-cache hit rates for everyone downstream."
     )
+    languages = ("python", "typescript")
 
     def check(self, project: Project) -> list[Finding]:
+        if project.language == "typescript":
+            return self._check_ts(project)
+        return self._check_python(project)
+
+    def _check_python(self, project: Project) -> list[Finding]:
         out: list[Finding] = []
         for f in project.files:
             # A decorator (`@server.list_tools()`) and the `def` line right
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged — and thanks for coming back for a second one after R003.

I want to flag something concrete, because it is worth more to you than "looks good". The look-ahead window:

```python
end_line = min(len(f.lines), line_no + 40)
```

The Python side of this same rule deliberately does *not* do that, and the comment eight lines above your code says why:

> Scope the look-ahead to the enclosing block by indentation instead of a fixed number of lines: a fixed window either cuts a long handler short (false positive) or bleeds into the next function and picks up an unrelated `.sort()` (false negative).

Both halves of that reproduce against your merged code. I ran them:

```
unsorted handler + unrelated sort() 10 lines later -> 0 findings  (expected 1)
long handler that DOES sort, past line 40          -> 1 finding   (expected 0)
```

The first is the one that actually costs us: an MCP server with an unsorted `tools/list` and any `.sort()` anywhere in the following 40 lines gets a clean bill of health. That is a silent miss, which is the failure mode this project cares most about.

This is **not a request to fix it in this PR** — it is merged and it is a net improvement over having no TypeScript R004 at all. I have opened a follow-up and would rather you take it than anyone else, since you now have the most context on it: the fix is a brace-depth scan (`{`/`}` outside strings and comments) to find the end of the arrow function, which is the TypeScript equivalent of what `_body_bounds` does with indentation in Python. `_ts_spans` already tells you which braces are inside strings or comments, so most of the hard part exists.

Also worth knowing: `TS_HANDLER_RX` is `re.IGNORECASE`, so it matches `listtools` and `LISTTOOLS` as well as `listTools`. Harmless today, but TypeScript is case-sensitive and the SDK names are fixed, so the flag is buying nothing and could surprise someone later.

dheerajjha:
For the record, the follow-up I mentioned above is **#65** — I wrote "I have opened a follow-up" without the number, which was careless. You found it anyway and claimed it; it is assigned to you and I have left implementation notes there.

PR #62, opened 2026-08-02
feat(rules): port R015 (resultType) to TypeScript

### LINKED ISSUE #44: TypeScript: port R015 (Results are returned without the required resultType field)
Port **R015** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Results are returned without the required resultType field (`advisory`)
**Python implementation to mirror:** [`r015_result_type_required.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r015_result_type_required.py)
**Spec:** SEP-2322 https://modelcontextprotocol.io/specification/2026-07-28/changelog

Only fires for servers that build their own JSON-RPC envelopes -- if the project imports the SDK, the SDK owns `resultType` and the rule must stay silent. Read the Python version's early return first; that check is the whole point of the rule.

### How to do it

### Done when

`mcp-migrate check` reports R015 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

syf2211:
## Summary
Port `RequiredResultTypeMissing` (R015) to the TypeScript language backend so `mcp-migrate check` can flag hand-rolled JSON-RPC servers that omit the required `resultType` field.

## Motivation
Issue #44 asks for a TypeScript port of R015. The Python rule already exists; TypeScript servers are the majority of MCP codebases, so this rule needs the same SDK exemption and hand-rolled envelope detection on the TS side.

## Changes
- Add `languages = ("python", "typescript")` and split `check()` into `_check_python` / `_check_ts`
- Exempt projects that import `@modelcontextprotocol/sdk` (SDK stamps `resultType` on serialized results)
- Detect hand-rolled envelopes via `search_wire` on `jsonrpc` keys (quoted and unquoted) and MCP method string literals
- Keep the permissive raw-text `resultType` presence check (comment mention silences the rule, same as Python)
- Add TypeScript tests: legacy hand-rolled server, migrated server, comment-only mention, quoted `jsonrpc` key, SDK exemption
- Update partial-coverage denominator to `5 of 21`

## Tests
```bash
python3 -m pytest tests/test_typescript.py -q
python3 -m pytest -q
```
All 194 tests pass.

## Notes
- Uses `search_wire` (not `search_code`) for wire-protocol string literals, consistent with R001/R006
- Project-wide SDK exemption matches the Python rule's behavior

Fixes #44

### DIFF
--- a/src/mcp_migrate/rules/r015_result_type_required.py
+++ b/src/mcp_migrate/rules/r015_result_type_required.py
@@ -36,12 +36,19 @@ RESULT_TYPE_MENTION_RX = re.compile(r"resultType")
 # JSON-RPC responses and therefore owns the envelope itself.
 SDK_SERVER_MODULES = ("mcp.server", "mcp.types", "mcp.shared", "fastmcp")
 
+# TypeScript. The official SDK stamps resultType on every result it
+# serializes, same as the Python runner -- a file that imports
+# @modelcontextprotocol/sdk has nothing to add and nothing to fix.
+TS_SDK_IMPORT_RX = re.compile(r"@modelcontextprotocol/sdk")
+
 # What a server that owns its own envelope looks like: it writes the JSON-RPC
 # framing and the MCP method names itself, as string literals. Both halves are
 # required -- "jsonrpc" alone is any JSON-RPC service on earth, and a method
 # name alone shows up in tests and docs -- and the project must not import the
 # SDK, which is checked first.
 JSONRPC_ENVELOPE_RX = re.compile(r"[\"']jsonrpc[\"']")
+# TypeScript object literals usually use an unquoted `jsonrpc:` key.
+TS_JSONRPC_ENVELOPE_RX = re.compile(r"\bjsonrpc\s*:")
 MCP_METHOD_NAME_RX = re.compile(
     r"[\"'](?:tools/list|tools/call|resources/list|resources/read|"
     r"resources/templates/list|prompts/list|prompts/get)[\"']"
@@ -53,6 +60,10 @@ def _sdk_owns_serialization(project) -> bool:
         if mod == "mcp" or mod.startswith(SDK_SERVER_MODULES):
             return True
     return False
+
+
+def _ts_sdk_owns_serialization(project: Project) -> bool:
+    return any(TS_SDK_IMPORT_RX.search(f.text) for f in project.files)
 
 
 # Downgraded from "breaking" after auditing real servers: resultType is a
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged. Three PRs in one day — thank you.

Good judgement in here: hoisting the message to `self.MESSAGE` so the Python and TypeScript branches cannot drift apart, keeping the project-wide SDK exemption rather than making it per-file (matching the Python semantics exactly), and covering both the bare `jsonrpc:` key and the quoted `"jsonrpc":` form. The quoted-key test is the kind of thing that only occurs to someone who has actually read hand-rolled TS servers.

One real issue, which I would like you to take as a follow-up. In `_check_ts`:

```python
for f in project.files:
    ...
    has_jsonrpc = any(
        ff.path == f.path
        for ff, _line, _text in project.search_wire(TS_JSONRPC_ENVELOPE_RX.pattern)
    ) or any(...)
```

`project.search_wire(...)` scans **every file in the project**, and it is called from inside a loop over every file. So the rule is O(n²) in file count. Measured on synthetic projects:

```
  25 files ->    1.4 ms
  50 files ->    4.8 ms
 100 files ->   16.6 ms
 200 files ->   63.0 ms
```

Doubling the file count roughly quadruples the time. At 2,000 files — not unusual for a TS monorepo — that is several seconds for this rule alone, and R016 in #63 has the same shape.

The fix is to hoist the scans out of the loop and bucket by path once:

```python
jsonrpc_paths = {ff.path for ff, _, _ in project.search_wire(...)}
```

then the per-file loop is pure dict lookups. Interestingly **you already wrote it this way in #63** — `handler_hits.setdefault(...)` is exactly the right pattern. This one just predates it.

Nothing is broken and no result changes; it is purely the constant factor. If you want it, say so on the follow-up issue and it is yours.

dheerajjha:
Following up on my review above, with an apology attached.

The follow-up I mentioned was **#67**, and I wrote "@syf2211 has first refusal" in it. Then I merged a fix from someone else (#70) about forty minutes later without checking whether you had started. **#67 is closed.** If you had begun work on it, I wasted your time, and that is entirely on me — a first-refusal note is worthless if I do not honour it before merging.

I have changed how I will handle this: nothing offered to a specific person gets merged from a third party until either they have claimed something else or I have asked them directly.

For the record, here is what landed, since it is your code that was being changed:

```
          before      after
 100 files    16.6 ms     1.5 ms
 200 files    63.0 ms     3.1 ms
```

Quadratic to linear. Findings unchanged, all 223 tests green.

**Still open and genuinely unclaimed** if you want one:

- **#66** — the R016 false negative you found and documented in #63. Nobody has taken it. It is yours by right if you want it, and I will not merge anyone else on it without asking you first.
- **#36 (R007)** — the port with a real-world false positive that took a hand-audit to catch.
- Cookbook stubs 15 and 16, both covering rules you have already worked in.

You are at four merged PRs, more than anyone else here. Thank you — and sorry for the collision.

PR #63, opened 2026-08-02
feat(rules): port R016 (cacheable result metadata) to TypeScript

### LINKED ISSUE #45: TypeScript: port R016 (List/read results are returned without ttlMs / cacheScope)
Port **R016** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** List/read results are returned without ttlMs / cacheScope (`advisory`)
**Python implementation to mirror:** [`r016_cacheable_result_required.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r016_cacheable_result_required.py)
**Spec:** SEP-2549 https://modelcontextprotocol.io/specification/2026-07-28/changelog

`ttlMs`/`cacheScope` on list/read results, and whether cache hints are configured at construction instead.

### How to do it

### Done when

`mcp-migrate check` reports R016 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

syf2211:
## Summary

Port rule R016 () to the TypeScript language backend so `mcp-migrate check` can flag list/read handlers that omit `ttlMs`/`cacheScope` when no server-level cache hints are configured.

## Motivation

Issue #45 asks for a TypeScript port of R016. List/read MCP results now require cache metadata per SEP-2549; TypeScript servers using low-level `setRequestHandler` need the same advisory detection Python servers already get.

## Changes

- Add `languages = ("python", "typescript")` and split `check()` into `_check_python` / `_check_ts`
- TypeScript handler detection via `setRequestHandler` + list/read request schemas or wire method strings (`tools/list`, etc.)
- Project-wide exemption when `cacheHints` or per-resource `cacheHint:` appears anywhere in the project
- Per-file generous presence check for `ttlMs` / `cacheScope` (mirrors Python)
- Five new tests in `tests/test_typescript.py`; update partial-coverage denominator to 5 of 21

## Tests

```
python3 -m pytest tests/ -q
195 passed
```

## Notes

- Follows the same `search_code` / `search_wire` split as other TS rule ports
- `McpServer` + `registerTool` only (no explicit `setRequestHandler`) is intentionally out of scope for per-file handler detection, same class of gap as Python FastMCP functional registration; server-level `cacheHints` exemption still applies

Fixes #45

### DIFF
--- a/src/mcp_migrate/rules/r016_cacheable_result_required.py
+++ b/src/mcp_migrate/rules/r016_cacheable_result_required.py
@@ -38,6 +38,33 @@ CACHE_META_MENTION_RX = re.compile(r"ttlMs|cacheScope")
 # cache metadata at all.
 CACHE_HINT_CONFIG_RX = re.compile(r"\bcache_hints\s*=|\bCacheHint\s*\(")
 
+# --- TypeScript -----------------------------------------------------------
+#
+# Low-level servers register list/read handlers with request schemas or wire
+# method names. search_code for the schema identifiers; search_wire for the
+# method string literals in setRequestHandler calls.
+TS_CACHEABLE_HANDLER_RX = (
+    r"setRequestHandler\s*\(\s*(?:ListTools|ListPrompts|ListResources|"
+    r"ReadResource|ListResourceTemplates)RequestSchema\b"
+)
+TS_CACHEABLE_METHOD_RX = (
+    r"setRequestHandler\s*\(\s*"
+    r"[\"'](?:tools/list|prompts/list|resources/list|resources/read|"
+    r"resources/templates/list)[\"']"
+)
+
+# Server-level hints are configured once on construction:
+#   new McpServer({ ... }, { cacheHints: { 'tools/list': { ttlMs, cacheScope }}})
+# Per-resource hints use cacheHint: on registerResource. Either satisfies
+# the rule project-wide -- scanning handler files for literal ttlMs/cacheScope
+# would miss a server that configured hints correctly in a different file.
+TS_CACHE_HINT_CONFIG_RX = re.compile(r"\bcacheHints\b|\bcacheHint\s*:")
+
+MESSAGE = (
+    "This file implements a list/read handler but neither `ttlMs` nor "
+    "`cacheScope` appears in it."
+)
+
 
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged.

The `handler_hits.setdefault(...)` bucketing here is the right shape — one pass per pattern, then dict lookups — and it is worth knowing that this is a genuine improvement over how you wrote the equivalent loop in #62 an hour earlier. That one is quadratic; this one is not. I left the details there.

The thing I most want to acknowledge is this comment:

```python
# Comment mentions ttlMs/cacheScope but handler still lacks them -- the
# rule uses a generous presence check, so a comment counts as handled.
assert CacheableResultMetadataMissing().check(project) == []
```

You found a false **negative** — a comment saying `ttlMs` silences a real finding — and instead of quietly writing the assertion to match, you wrote the assertion to match *and said what was wrong with it*. That is the single most useful thing a contributor can do in a static-analysis codebase, and it is rarer than it should be. It is inherited from the Python side (`CACHE_META_MENTION_RX` searches `f.text` raw, comments included), so you did not introduce it — you documented it.

I have filed it as a follow-up. The fix is to run that mention check through the content spans the way `search_code` does, rather than raw `f.text`, which will need doing on both language branches at once.

R020 in #64 is merged too — that is four from you. If you want a fifth, **R007 (#36)** is the interesting one left: it had a false positive in the wild that took a hand-audit to catch, so the test-writing matters more than the pattern.

dheerajjha:
Small follow-up: the issue I filed for the false negative you documented here is **#66** — above I said "I have filed it as a follow-up" without giving you the number, which was not much use.

It is still unclaimed. Given you are the one who found it, it is yours if you want it, and I will not merge anyone else on it without checking with you first. (I failed to do exactly that on #67 and have apologised on #62.)

The shape of the fix: `CACHE_META_MENTION_RX` is searched against raw `f.text`, so it matches inside comments. Routing it through content spans — the way `search_code` already does — makes a mention in code count and a mention in a comment not count. It affects both language branches. The same raw-`f.text` pattern appears in R005 and R015, but those deserve their own issues rather than one sprawling PR.

PR #64, opened 2026-08-03
feat(rules): port R020 (Dynamic Client Registration) to TypeScript

### LINKED ISSUE #49: TypeScript: port R020 (Uses Dynamic Client Registration (RFC 7591), now deprecated)
Port **R020** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Uses Dynamic Client Registration (RFC 7591), now deprecated (`deprecated`)
**Python implementation to mirror:** [`r020_dynamic_client_registration_deprecated.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r020_dynamic_client_registration_deprecated.py)

RFC 7591 dynamic client registration -- `/register` endpoints and `client_id_issued_at`.

### How to do it

### Done when

`mcp-migrate check` reports R020 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

syf2211:
## What kind of PR is this?

- [x] Rule -- adds/changes a rule in `src/mcp_migrate/rules/`
- [ ] Something else

## Summary

Port R020 (`DynamicClientRegistrationDeprecated`) to the TypeScript language backend. The rule now detects RFC 7591 Dynamic Client Registration usage in `.ts` files via `search_code`, matching `RegisterClientRequest`, `DynamicClientRegistration`, and the camelCase SDK hook `registerClient`.

## Motivation

Fixes #49. Most MCP servers are TypeScript; without this port, R020 only runs on Python files and misses the majority of real DCR usage.

## Changes

- Add `languages = ("python", "typescript")` to `DynamicClientRegistrationDeprecated`
- Split Python (`register_client`) and TypeScript (`registerClient`) identifier patterns
- Add five TypeScript tests: legacy `registerClient`, `RegisterClientRequest` type usage, CIMD-migrated auth, comment-only mentions, unrelated register methods
- Update partial-coverage denominator assertion to `5 of 21`

## Rule checklist

- [ ] Fixtures added under `tests/fixtures/<rule-id>/` (inline tmp_path tests used, same as R001/R003/R005/R006)

## Tests

```bash
python3 -m pytest tests/test_typescript.py -q   # 27 passed
python3 -m pytest -q                           # 195 passed
```

## Notes

- Uses `search_code` (not `search_wire`) because all signals are code identifiers, mirroring the Python implementation.
- `/register` endpoint URLs in string literals are intentionally not matched (same as Python).
- Reviewed by composer-2.5 subagent before submission.

Fixes #49

### DIFF
--- a/src/mcp_migrate/rules/r020_dynamic_client_registration_deprecated.py
+++ b/src/mcp_migrate/rules/r020_dynamic_client_registration_deprecated.py
@@ -1,5 +1,3 @@
-import re
-
 from .base import Finding, Project, Rule
 
 # All three are specific enough to MCP/OAuth client-registration code to
@@ -10,8 +8,15 @@ from .base import Finding, Project, Rule
 # A generic "register a client" method in an unrelated CRM/billing app
 # wouldn't spell it exactly `register_client`, so this is precise enough
 # for `deprecated` (the lower-stakes severity this item calls for).
-DCR_CODE_RX = re.compile(
+PY_DCR_CODE_RX = (
     r"\bRegisterClientRequest\b|\bDynamicClientRegistration\b|\bregister_client\b"
+)
+
+# TypeScript uses camelCase for the SDK auth hook (`registerClient` in
+# @modelcontextprotocol/sdk/client/auth.ts) where Python spells it
+# `register_client`. The type names are the same across languages.
+TS_DCR_CODE_RX = (
+    r"\bRegisterClientRequest\b|\bDynamicClientRegistration\b|\bregisterClient\b"
 )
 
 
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged — your fourth, and the cleanest of the four.

Collapsing the two languages onto one `search_code` call with a swapped pattern is exactly right for this rule:

```python
pattern = TS_DCR_CODE_RX if project.language == "typescript" else PY_DCR_CODE_RX
```

R020 has no context gate and no second signal, so a `_check_ts` method would have been ceremony around a one-line difference. Knowing when *not* to mirror the `_check_python`/`_check_ts` split is a judgement call and you got it right — several of the other ports would have been better off this way too.

Renaming the existing constant to `PY_DCR_CODE_RX` rather than leaving it bare is a small thing that will keep paying off; so is the comment recording that the SDK spells the hook `registerClient` in TS and `register_client` in Python, which is precisely the sort of detail the next person would otherwise have to re-derive from the SDK source.

TypeScript coverage is **9 of 21** now and four of those nine are yours.

Issue #65, opened 2026-08-03
R004 (TypeScript): scope the sort look-ahead by brace depth, not a fixed 40 lines


dheerajjha:
The TypeScript branch of R004 uses a fixed look-ahead window:

```python
end_line = min(len(f.lines), line_no + 40)
```

The Python branch deliberately does not, and the comment right above it says why:

> Scope the look-ahead to the enclosing block by indentation instead of a fixed number of lines: a fixed window either cuts a long handler short (false positive) or bleeds into the next function and picks up an unrelated `.sort()` (false negative).

Both halves reproduce on `main` today:

```
unsorted handler + unrelated sort() 10 lines later -> 0 findings  (expected 1)
long handler that DOES sort, past line 40          -> 1 finding   (expected 0)
```

The false negative is the one that matters: an MCP server with an unsorted `tools/list` gets a clean bill of health if any `.sort()` appears in the following 40 lines.

## What to do

Replace the fixed window with a brace-depth scan to find the end of the handler body — the TypeScript equivalent of what `_body_bounds` does with indentation in Python. Count `{` and `}` from the match position and stop when depth returns to zero.

The hard part is already done: `_ts_spans(text, prose_only=False)` in `rules/base.py` tells you which offsets are inside strings or comments, so you can skip braces that are not real. A brace in `"} not code {"` must not count.

## Tests to write

Both cases above, plus a handler containing a template literal with a `}` in it.

## Context

Introduced in #61. @atiqur-rahman-pro has first refusal — say the word and it is yours. The helper this produces will likely be reusable by other TypeScript ports that need block scoping.

atiqur-rahman-pro:
Hi @dheerajjha,

Thank you so much for merging PR #61 and your detailed, thoughtful feedback!

I'd be happy to take on Issue #65. I will complete the brace-depth scan (`{`/`}`) using `_ts_spans` to properly scope the TypeScript handler block and add the test cases.

dheerajjha:
Yours — assigned. And your plan is exactly right: brace depth via `_ts_spans`, plus the tests.

One thing to save you time. `_ts_spans(text, prose_only=False)` returns spans covering comments **and** string literals, which is what you want here — a `}` inside `"} not code {"` or inside a template literal must not count toward depth. The helper takes `((start_line, start_col), (end_line, end_col))` tuples, and `_in_content_span(line, col, spans)` is already in `base.py` for testing membership.

[...]

PR #70, opened 2026-08-03
perf: hoist R015 TypeScript scans

### LINKED ISSUE #67: R015 (TypeScript): hoist the search_wire scans out of the per-file loop (it is O(n^2))
`_check_ts` in `r015_result_type_required.py` calls `project.search_wire(...)` from inside a loop over `project.files`. `search_wire` scans every file in the project, so the rule is quadratic in file count.

Measured on synthetic projects:

```
  25 files ->    1.4 ms
  50 files ->    4.8 ms
 100 files ->   16.6 ms
 200 files ->   63.0 ms
```

Doubling the file count roughly quadruples the time. A 2,000-file TypeScript monorepo would spend several seconds in this one rule.

## What to do

Scan once, bucket by path, then let the per-file loop do dict lookups:

```python
jsonrpc_paths = {ff.path for ff, _, _ in project.search_wire(TS_JSONRPC_ENVELOPE_RX.pattern)}
jsonrpc_paths |= {ff.path for ff, _, _ in project.search_wire(JSONRPC_ENVELOPE_RX.pattern)}

```

**No finding should change** — this is purely the constant factor. The existing R015 tests are the regression net; if any of them flips, the rewrite is wrong.

The pattern to copy is already in this repo: `_check_ts` in `r016_cacheable_result_required.py` does exactly this with `handler_hits.setdefault(...)`.

## Context

Introduced in #62, by the same author who then wrote the correct shape in #63 an hour later. @syf2211 has first refusal.

li2631026381-alt:
## What changed

Hoist the R015 TypeScript `search_wire` scans out of the per-file loop and bucket matches by path. This avoids rescanning the entire project for every file.

## Why

The previous implementation called `project.search_wire(...)` inside the loop over `project.files`, making the rule quadratic in project file count. The new implementation scans each pattern once while preserving the first method hit and existing file-order behavior.

## Validation

- `uv run pytest tests/test_rules.py tests/test_regressions.py tests/test_typescript.py -q` — 150 passed
- `uv run pytest -q` — 223 passed
- `git diff --check` — passed

Closes #67,

### DIFF
--- a/src/mcp_migrate/rules/r015_result_type_required.py
+++ b/src/mcp_migrate/rules/r015_result_type_required.py
@@ -137,33 +137,29 @@ class RequiredResultTypeMissing(Rule):
     def _check_ts(self, project: Project) -> list[Finding]:
         if _ts_sdk_owns_serialization(project):
             return []
+        # search_wire keeps string literals ("jsonrpc", "tools/list")
+        # and drops comments -- a README-style comment naming the
+        # protocol is not a hand-rolled envelope. Scan the project once
+        # per pattern, rather than once per file in the loop below.
+        jsonrpc_paths = {
+            f.path for f, _line, _text in project.search_wire(TS_JSONRPC_ENVELOPE_RX.pattern)
+        }
+        jsonrpc_paths.update(
+            f.path for f, _line, _text in project.search_wire(JSONRPC_ENVELOPE_RX.pattern)
+        )
+        method_hits: dict[object, tuple[int, str]] = {}
+        for f, line, text in project.search_wire(MCP_METHOD_NAME_RX.pattern):
+            method_hits.setdefault(f.path, (line, text))
+
         out: list[Finding] = []
-        seen_files = set()
         for f in project.files:
             if RESULT_TYPE_MENTION_RX.search(f.text):
                 continue
-            if f.path in seen_files:
+            if f.path not in jsonrpc_paths:
                 continue
-            # search_wire keeps string literals ("jsonrpc", "tools/list")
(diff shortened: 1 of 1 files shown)


dheerajjha:
Merged — this is exactly the fix, and you are the first PR to land under the new CI policy (your checks ran automatically instead of waiting on me, which is a change made about twenty minutes before you opened this).

Measured before and after on synthetic projects:

```
          before      after
  25 files     1.4 ms     0.4 ms
  50 files     4.8 ms     0.8 ms
 100 files    16.6 ms     1.5 ms
 200 files    63.0 ms     3.1 ms
 400 files       --       6.2 ms
```

Quadratic to linear. **20x at 200 files**, and the gap widens from there — a 2,000-file monorepo goes from roughly six seconds to about thirty milliseconds in this one rule.

Two things you did that I want to name, because they are the difference between a patch that works and one I can merge without worrying:

**You dropped `seen_files` rather than leaving it in.** Once the loop iterates `project.files` directly, each path is visited exactly once and the guard is dead weight that implies a duplicate risk which no longer exists. Removing it is the kind of thing that gets left behind in a hurry.

**You moved the `search_wire` comment rather than deleting it.** It explains why string literals are kept and comments dropped, and it is now attached to the hoisted scan where it still applies. Comments that explain a *decision* are the expensive kind to lose.

Findings are unchanged and all 223 tests pass, which is the property that mattered — this was meant to be a pure constant-factor change and it is.

If you want another: **#65** is the interesting one left. R004 uses a fixed 40-line look-ahead where it needs brace-depth scoping, and it fails in both directions — issue has runnable reproductions for both.

PR #71, opened 2026-08-03
docs(cookbook): fill R003 routing headers recipe

### LINKED ISSUE #9: Cookbook recipe: required Mcp-Method / Mcp-Name routing headers (R003)
`cookbook/14-routing-headers.md` exists as a stub.

**Context:** R003 ([`src/mcp_migrate/rules/r003_routing_headers.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r003_routing_headers.py))
flags hand-rolled HTTP clients speaking MCP's wire protocol that don't set
the required `Mcp-Method` (and, on `tools/call`/`resources/read`/
`prompts/get`, `Mcp-Name`) headers. Read the rule source's comments first --
this rule was downgraded from `breaking` to `advisory` after a real
false-positive incident (19 false hits on mcp-atlassian) and the recipe
should explain why.

**What "done" looks like:** fill in Before/After/Gotchas in
`cookbook/14-routing-headers.md` per `cookbook/_TEMPLATE.md`, with a
concrete example each of a file that *does* and *doesn't* trip the rule's
`_imports_mcp`/`MCP_METHOD_RX` gating, so a reader understands the
distinction between "hand-rolling MCP transport" and "wrapping an unrelated
backend REST API."

See CONTRIBUTING.md#add-a-cookbook-recipe-5-minutes.

syf2211:
## Summary

Complete the stub cookbook recipe for R003 (required `Mcp-Method` / `Mcp-Name` routing headers on Streamable HTTP).

## Motivation

Fixes #9. The stub had rule metadata and a one-line "what broke" but was missing the worked before/after examples and gotchas that make a cookbook recipe useful.

## Changes

- Fill `cookbook/14-routing-headers.md` with Python and TypeScript before/after examples
- Document gotchas: advisory severity, mcp-atlassian false-positive gating (`_imports_mcp` / `MCP_METHOD_RX`), `Mcp-Name` scope, per-file header checks, FastMCP exclusion
- Move recipe 14 from the stubs table to "Filed so far" in `cookbook/README.md`

## Tests

Cookbook-only change (no Python code). Verified R003 regression fixtures still pass:

```
pytest tests/test_regressions.py::test_plain_rest_post_with_no_mcp_surface_does_not_fire_r003 \
       tests/test_regressions.py::test_r003_still_fires_when_file_hand_rolls_jsonrpc_payload \
       tests/test_regressions.py::test_fastmcp_import_alone_does_not_satisfy_the_mcp_surface_gate
# 3 passed
```

## Notes

Reviewed against `r003_routing_headers.py` and the regression fixtures (`webhook.py`, `jira_client.py`, `fastmcp_wrapper_no_mcp_surface/server.py`).

Fixes #9

### DIFF
--- a/cookbook/14-routing-headers.md
+++ b/cookbook/14-routing-headers.md
@@ -1,6 +1,3 @@
-<!-- STUB: rule/spec filled in, before/after/gotchas still needed. See
-     .github/GOOD_FIRST_ISSUES.md for the ready-to-file issue for this one. -->
-
 # Required `Mcp-Method` / `Mcp-Name` routing headers
 
 - **Rule:** [R003](../src/mcp_migrate/rules/r003_routing_headers.py)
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged — your fourth, and the first non-rule contribution anyone has made.

I verified both examples against R003 rather than taking them on trust:

```
Before (should flag)       -> 1 finding
After  (should be clean)   -> 0 findings
```

They behave exactly as written, which is not a given for a doc PR — a "before" that does not actually trip the rule is worse than no example.

The Gotchas section is the part I did not expect. You went and found the mcp-atlassian incident:

> that combination was mcp-atlassian's worst false positive (19 hits before the rule was tightened), which is why severity is `advisory`, not `breaking`

That is not in the rule source. It is in the commit history and the audit notes, and you connected it to the severity choice on your own. Same with the `Mcp-Name` gotcha — `tools/list` and `initialize` need `Mcp-Method` but not `Mcp-Name`, and flagging its absence *was* an early R003 bug. Documenting the rule's own past mistakes is how a reader learns to trust the tool where it is confident and discount it where it is guessing.

Cookbook index updated correctly in the same PR, so 14 moved out of Stubs. **13 stubs left** if you want more of these — 15 (deterministic ordering, R004) and 16 (extensions map, R005) both cover rules you have now worked in.

Issue #32, opened 2026-08-01
TypeScript: port R002 (Keeps per-connection state in a module-level dict)


dheerajjha:
Port **R002** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Keeps per-connection state in a module-level dict (`breaking`)
**Python implementation to mirror:** [`r002_connection_state.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r002_connection_state.py)
**Spec:** SEP-2567 https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567

### What makes this one specific

Module-level mutable state in TS is a top-level `const sessions = new Map()` or a module-scoped object. The Python rule walks the AST; TS has no AST here, so this one is textual -- look for top-level `new Map(`/`new Set(`/`= {}` that a request handler then mutates.

### How to do it

1. Add `languages = ("python", "typescript")` to the rule class.
2. Split `check()` on `project.language`, exactly like the two reference ports:
   - [`r001_session_id_removed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r001_session_id_removed.py) — identifier match *and* a string-literal match anchored to a header access
   - [`r006_sse_transport_deprecated.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r006_sse_transport_deprecated.py) — the simpler shape
3. Pick your search mode deliberately. This is the single easiest thing to get wrong:
   - `search_code()` — skips comments **and** strings. For code identifiers.
   - `search_wire()` — skips comments and docstrings, **keeps** string literals. For removed JSON-RPC method names, which only ever appear inside strings.
   - `search()` — raw, matches prose too. Almost never right.
4. Add tests to `tests/test_typescript.py`: one that fires on a legacy server, one that stays silent on a migrated one, and one that stays silent when the thing is only *named in a comment*.

### Done when

`mcp-migrate check` reports R002 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

Ask in the thread if anything is unclear — happy to talk it through before you write code, and PRs get reviewed the same day.

IronLad123:
[...]

Issue #29, opened 2026-07-31
Fixer: older JSON Schema dialect pinned (R021)


dheerajjha:
No fixer exists yet for R021 ([`src/mcp_migrate/rules/r021_json_schema_2020_12_required.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r021_json_schema_2020_12_required.py)),
which flags an explicit older JSON Schema dialect (`draft-07`, `2019-09`,
...) pinned via a `$schema` URL string.

**Spec:** SEP-2106 -- https://modelcontextprotocol.io/specification/2026-07-28/changelog

**A reasonable scope:** this is one of the more mechanical ones in this
list, similar in spirit to the shipped R017 fixer -- rewriting an old
`$schema` dialect URL string to the 2020-12 equivalent
(`http://json-schema.org/draft/2020-12/schema`) is a close-to-exact string
substitution once you're confident the match is a real `$schema` value and
not, say, a comment mentioning an old draft. Confidence `safe` is plausible
here if you scope it tightly (e.g. only rewrite when the match is inside a
`"$schema":` key's value).

**What "done" looks like:** a `Fixer` subclass in `src/mcp_migrate/fixers/`,
with fixtures and tests per `tests/test_fixers.py`, including a fixture that
proves the fixer backs off on an ambiguous shape (the dialect string
appearing somewhere that isn't clearly a `$schema` value).

See CONTRIBUTING.md#add-a-fixer-45-minutes.

IronLad123:
Hi @dheerajjha! I'd love to take this on. I've been working with the mcp-migrate codebase (recently had PR #76 merged) and the R021 fixer looks like a clean, well-scoped addition. I'll implement the `Fixer` subclass in `src/mcp_migrate/fixers/`, add fixtures and tests following `tests/test_fixers.py`, and make sure the `safe` confidence scoping is tight — only rewriting when the dialect string is clearly inside a `"$schema":` key value. Will report back shortly!

dheerajjha:
Assigned to @IronLad123, who has #97 up for it.

Retroactive bookkeeping — the claim was made here at 06:10 and the PR was already open, but the issue itself said nothing, which is the gap that caused the mess on #32. Assigning on claim from now on rather than after the fact.

#97 is one constant away from merging: `_TARGET_DIALECT` is `http://` and 2020-12 is published under `https`. Details on the PR.

Issue #28, opened 2026-07-31
Fixer: Dynamic Client Registration deprecated (R020)


dheerajjha:
No fixer exists yet for R020 ([`src/mcp_migrate/rules/r020_dynamic_client_registration_deprecated.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r020_dynamic_client_registration_deprecated.py)),
which flags RFC 7591 Dynamic Client Registration usage (`register_client`,
`RegisterClientRequest`, `DynamicClientRegistration`).

**Spec:** https://modelcontextprotocol.io/specification/2026-07-28/changelog

**A reasonable scope:** since `deprecated` (not `breaking`) and migrating
off DCR requires standing up a real Client ID Metadata Document, a
mechanical fixer probably can't do the migration itself -- but it could
annotate the flagged code with a `# TODO(mcp-migrate): ...` pointing at
`cookbook/12-dynamic-client-registration-deprecated.md`, matching the
TODO-annotation pattern used elsewhere in this project.

**What "done" looks like:** a `Fixer` subclass in `src/mcp_migrate/fixers/`
using that pattern, `review` confidence, with fixtures and tests per
`tests/test_fixers.py`.

See CONTRIBUTING.md#add-a-fixer-45-minutes.

IronLad123:
Hi @dheerajjha! I'd like to implement the R020 fixer for Dynamic Client Registration. Following the pattern established in the codebase (and building on PR #97), I will implement the `Fixer` subclass in `src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py` with `review` confidence to annotate deprecated DCR usage (`register_client`, `RegisterClientRequest`, `DynamicClientRegistration`) with a `# TODO(mcp-migrate): ...` annotation pointing at `cookbook/12-dynamic-client-registration-deprecated.md`. Adding fixtures and unit tests to `tests/test_fixers.py`. Will submit a PR shortly!

dheerajjha:
Assigned to @IronLad123 — #103 implements it, and it's the first fixer in the repo that handles TypeScript as well as Python. Review is on the PR: two small fixes and it goes in.

Same retroactive-assignment note as #29 — the claim was visible here, the assignment wasn't.

PR #73, opened 2026-08-03
fix: ignore comment-only R016 metadata mentions

### LINKED ISSUE #66: R016: a comment mentioning ttlMs/cacheScope silences a real finding (both languages)

li2631026381-alt:
## What changed

Fix R016 so comments and docstrings mentioning `ttlMs` or `cacheScope` do not silence a real missing-metadata finding.

The rule now uses `search_wire` to collect metadata mentions that occur in code or string literals while excluding prose. This is applied to both Python and TypeScript branches.

Updated regression coverage verifies:

- Python comments still produce an R016 finding
- TypeScript comments still produce an R016 finding
- metadata in a string-built response is accepted
- the clean fixture configures real cache hints instead of relying on a comment

## Validation

- `uv run pytest -q` — 225 passed
- `git diff --check` — passed

Closes #66,

### DIFF
--- a/src/mcp_migrate/rules/r016_cacheable_result_required.py
+++ b/src/mcp_migrate/rules/r016_cacheable_result_required.py
@@ -11,11 +11,9 @@ CACHEABLE_HANDLER_RX = re.compile(
     r"list_resource_templates)\s*\("
 )
 
-# Presence check, deliberately raw text (same permissive direction as
-# r005_extensions.py / r015_result_type_required.py): a mention anywhere in
-# the file is enough to count the metadata as handled, because a wrong
-# "still missing" (breaking) claim costs far more than a generous read of
-# "present" does.
+# Presence check keeps string literals but ignores comments and docstrings:
+# a metadata field in a response or a string-built response counts, while a
+# prose mention must not silence a real finding.
 CACHE_META_MENTION_RX = re.compile(r"ttlMs|cacheScope")
 
 # The other, and more likely, way these fields get onto the wire.
@@ -100,12 +98,15 @@ class CacheableResultMetadataMissing(Rule):
         # project-wide question, not a per-file one.
         if any(project.search_code(CACHE_HINT_CONFIG_RX.pattern)):
             return []
+        cache_meta_paths = {
+            f.path for f, _line, _text in project.search_wire(CACHE_META_MENTION_RX.pattern)
+        }
         out: list[Finding] = []
         seen_files = set()
         for f, line, text in project.search_code(CACHEABLE_HANDLER_RX.pattern):
             if f.path in seen_files:
                 continue
-            if CACHE_META_MENTION_RX.search(f.text):
+            if f.path in cache_meta_paths:
                 continue
             seen_files.add(f.path)
             out.append(self.finding(MESSAGE, f, line, text))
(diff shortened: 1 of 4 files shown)


dheerajjha:
Thank you for this, and I owe you a straight answer rather than leaving it open.

**I merged #76 for this issue instead**, which does the same thing — routes the `ttlMs`/`cacheScope` mention check through content spans so a comment can't silence a real finding. @IronLad123 claimed #66 on the issue itself at 10:41 and had a PR up two minutes later; yours was a draft, so I read it as not-yet-ready and went with the one that was. That's a defensible call on the facts but a bad outcome for you, and the part I got wrong is that I never replied here to say the issue had been claimed. You had no way to know.

Your approach was sound. If you'd marked it ready it would have been a coin flip, and I'd rather tell you that plainly than pretend the other PR was better.

Concretely, so this doesn't happen again: I'll reply within the day when someone claims an issue, and I'll say on the issue when it's taken.

Three TypeScript ports are open and genuinely unclaimed right now — **R002 (#32), R008 (#37), R010 (#39)**. R010 is the interesting one: it's the rule with a known gap where three registration idioms real servers use go undetected, so it's more than a mechanical port. Say the word on any of them and I'll assign it to you within the hour so nobody duplicates you again.

dheerajjha:
@li2631026381-alt — following up on my own comment above, because it aged badly within six hours and you should hear it from me rather than discover it.

I offered you three things: R002 (#32), R008 (#37), R010 (#39). As of now:

- **#37 and #39 are closed** — both were already shipped by others when I named them. I offered you work that wasn't available. That's straightforwardly my error; I listed them from an issue view without checking their state.
- **#32 has been claimed and implemented** by @IronLad123, who commented on the issue at 09:03 and opened #104. They followed the process exactly. The gap is that I offered you the issue here and never annotated the issue itself, so there was no way for them to know, and no way for you to hold it.

That is twice now that a slot I pointed you at went to someone else, and both times the mechanism was the same: I said something in one thread and failed to record it where it was visible. It isn't bad luck and I'm not going to describe it as such.

What I've changed, concretely, not as an intention: claims now get **assigned on GitHub** at the moment they're made. #32, #28, #29 are assigned to IronLad123 as of this comment.

And one of them is assigned to you:

**#86 is yours** — *"Nothing stops the #67 quadratic scan from coming back."* You filed it, nobody else is on it, and it's assigned to you now with no reply needed and no window to miss. It's also the most characteristic piece of work on the board for you specifically: #70 was the hoist that fixed the quadratic scan, and #86 is the regression test that stops it returning. Nobody else in the repo has the context for it.

If you'd rather have something else instead or as well, these are open and I will assign on your word rather than the other way around:

- **#105** — fixers insert comments inside string literals and silently change their value. R017 and R019 do it on `main` today; I measured it this morning. It's the same class of problem as your #73: the code has no idea whether a line is code or the inside of a string. *(Offered to IronLad123 on #103 first, so this one is genuinely contested — say the word and I'll sort it out with them rather than letting it repeat.)*
- **#87 / #88** — the unbounded `\w*` and the missing wire-pattern end boundary. Small, verified, and both are false positives currently shipping in published grades.
- **#98** — every TypeScript project exits 2, so no TS user can fail a build on this tool.

Separately: #73 and #74 are both still open as drafts and both now conflict with `main`. I'm not going to close them out from under you. Tell me if you want either revived and I'll rebase-help, or say the word and I'll close them so they stop showing up in your list.

dheerajjha:
Closing as already fixed — see my note on #74 for the verification and for #123, which your #74 is responsible for surfacing.

PR #74, opened 2026-08-03
feat: port R017 resource-not-found detection to TypeScript

### LINKED ISSUE #46: TypeScript: port R017 (Uses the old -32002 resource-not-found error code)

li2631026381-alt:
## Summary

- port R017 to the TypeScript backend
- detect the legacy `-32002` resource-not-found code in TypeScript code and string literals
- keep migrated `-32602` code and comment-only mentions silent

## Validation

- `uv run pytest tests/test_typescript.py tests/test_regressions.py -q` (102 passed)
- `uv run pytest -q` (226 passed)
- `git diff --check`

Closes #46

### DIFF
--- a/src/mcp_migrate/rules/r017_resource_not_found_code_changed.py
+++ b/src/mcp_migrate/rules/r017_resource_not_found_code_changed.py
@@ -12,6 +12,13 @@ RESOURCE_NOT_FOUND_RX = re.compile(
     r"(?=.*-32002\b)(?=.*(?:resource|not[_ ]?found|notfound))", re.IGNORECASE
 )
 
+# TypeScript commonly spells the context as a camelCase identifier or keeps
+# it inside an error message. The same line-level positive evidence keeps the
+# generic numeric code from matching unrelated sentinels.
+TS_RESOURCE_NOT_FOUND_RX = (
+    r"(?=.*-32002\b)(?=.*(?:resource|not[_ -]?found))"
+)
+
 
 class ResourceNotFoundCodeChanged(Rule):
     id = "R017"
@@ -22,8 +29,20 @@ class ResourceNotFoundCodeChanged(Rule):
         "The resource-not-found error code changed from -32002 to -32602 (Invalid "
         "params). Update whatever raises or checks for -32002 in this context."
     )
+    languages = ("python", "typescript")
 
     def check(self, project: Project) -> list[Finding]:
+        if project.language == "typescript":
+            return [
+                self.finding(
+                    "-32002 for resource-not-found is the old code; 2026-07-28 uses -32602.",
+                    f, line, text,
+                )
+                for f, line, text in project.search_wire(
+                    TS_RESOURCE_NOT_FOUND_RX, flags=re.IGNORECASE
+                )
+            ]
+
         out: list[Finding] = []
         # Raw text, not search_code: the qualifying context is as likely
         # to be a trailing comment ("# resource not found") as it is to be
--- a/tests/test_typescript.py
+++ b/tests/test_typescript.py
@@ -30,6 +30,9 @@ from mcp_migrate.rules.r015_result_type_required import RequiredResultTypeMissin
 from mcp_migrate.rules.r016_cacheable_result_required import (
     CacheableResultMetadataMissing,
 )
+from mcp_migrate.rules.r017_resource_not_found_code_changed import (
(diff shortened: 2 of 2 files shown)


dheerajjha:
Thanks for this, and sorry it took me a day to get to it.

Two things you should know, both my fault for not saying sooner.

**This overlaps #54.** @PuvaanRaaj opened an R017 port before this one. I should have flagged the collision when yours appeared instead of letting you both carry it — that's a maintainer failure, not yours, and I'm sorry for the wasted time.

**Both ports have the same false positive**, which is genuinely interesting. I reproduced it on this branch:

```ts
  // old -32002 resource not found, migrated to -32602
  return { code: -32602 };
}
```

Reports `breaking`. The file is correct — it's already on `-32602` and the old code appears only in prose. Real `-32002` usage is still detected, so the port works; it just also fires on comments that aren't at column 0.

The cause is worth knowing because you'd hit it again: the rule's regex is built **entirely of lookaheads**, so it consumes nothing and `match.start()` is always `0`. `search_wire` then checks whether *column 0* is inside a comment — and column 0 is the indentation, which is code. A `//` starting at column 0 works by accident; an indented one leaks through. Both of you inherited it from the same place, so this is a sharp edge in the pattern rather than a mistake either of you made.

Fix is to anchor on something that actually consumes the token (`re.compile(r"-32002\b")`, with the surrounding conditions as separate lookaheads on the line), or to report at the column of the real signal instead of the match start.

Whoever gets there first gets the merge — I'm not going to hand it to either of you over the other. Your R015 perf work (#70) is still the biggest single speedup anyone has contributed here (63.0ms → 3.1ms at 200 files), so if you'd rather take something uncontested instead, R002 (#32), R008 (#37) and R010 (#39) are all open and nobody is near them.

dheerajjha:
@li2631026381-alt — closing this, and I want to be precise about why, because your version was not the worst of the three and in one respect it was the best.

**#46 had three open PRs against it**: yours, @PuvaanRaaj's #54, and @waterlemonnn's #122. I merged #122 — it was `MERGEABLE`, green, and carried three tests. Yours and #54 have been `CONFLICTING` since the stack under them moved.

**But your approach caught something the other two didn't.** You passed the flag explicitly:

```python
project.search_wire(TS_RESOURCE_NOT_FOUND_RX, flags=re.IGNORECASE)
```

The other two enable TypeScript by adding `languages = ("python", "typescript")` and reusing the existing pass — which is correct, and which quietly inherits a bug. R017 compiles its pattern with `re.IGNORECASE` and then hands `search_wire` only `RX.pattern`, the pattern *string*. The flags stay behind on the compiled object. Verified on `main`:

```
"resource not found"  -> R017 hits: 1
"Resource Not Found"  -> R017 hits: 0
```

The flag has been decorative since the rule was written, on the **Python** path, which is graded. Your PR is what made that visible — I only went looking because your diff passed `flags=` and the merged one didn't, and I wanted to know which was right.

I've filed it as **#123** and credited you in the issue. It's one line plus a test. **It's yours if you want it** — say the word and I'll assign it. If you'd rather not, it's up-for-grabs and no hard feelings.

**#73 I'm also closing**, for a different reason: it's already fixed. A comment mentioning `ttlMs`/`cacheScope` no longer suppresses R016 on `main` — I checked rather than assumed:

```
comment mentioning ttlMs/cacheScope -> R016 hits: 1   (not suppressed)
```

That landed via a different PR while yours was open. Same fix, same reasoning as your commit message — you were right about the diagnosis.

You have one merged PR here (the R015 quadratic-scan hoist, which #120 has now built a permanent regression guard around — that guard exists because of your fix). I'd rather not have your other two close as a footnote, so: #123 is a clean, small, verified bug with your name on the finding, and R013/R021 are still unported if you want something bigger.

PR #81, opened 2026-08-04
feat(rules): port R009 (initialize handshake) to TypeScript

### LINKED ISSUE #38: TypeScript: port R009 (Still implements the initialize / notifications/initialized handshake)
Port **R009** to TypeScript. One rule, self-contained — the scaffold already exists, so this is mostly writing a regex and a test.

**The rule:** Still implements the initialize / notifications/initialized handshake (`breaking`)
**Python implementation to mirror:** [`r009_initialize_handshake_removed.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r009_initialize_handshake_removed.py)

`InitializeRequestSchema`, `InitializedNotificationSchema`, and the literal `notifications/initialized`. The literal needs `search_wire`, not `search_code` -- it only ever appears inside a string.

### How to do it

### Done when

`mcp-migrate check` reports R009 on a TypeScript server that has the problem, stays silent on one that doesn't, and the tests cover both.

resuaico:
Fixes #38.

Ports **R009** (`Still implements the initialize / notifications/initialized handshake`, `breaking`) to TypeScript. `languages = ("python", "typescript")`, `check()` splits on `project.language`, and the Python path moves into `_check_python` unchanged — same layout as the R001/R006/R011 ports.

## The two signals

| Signal | Search mode | Why |
| --- | --- | --- |
| `InitializeRequest\w*` / `InitializeResult\w*` / `InitializedNotification\w*` | `search_code` | SDK names are code. A JSDoc block explaining that the handshake was removed is not a handler for it. |
| `notifications/initialized` | `search_wire` | Only ever valid as a JSON-RPC method-name string, so it always starts inside a string token and `search_code` would never find it. |

**The `\w*` suffix is the load-bearing difference from Python.** TypeScript ships a Zod schema alongside the inferred type, and the schema is the name a server actually writes:

```ts
server.setRequestHandler(InitializeRequestSchema, async () => { ... });
```

`\bInitializeRequest\b` cannot match inside `InitializeRequestSchema` — there is no word boundary before `Schema` — so without the suffix this port would find nothing on the one shape it most needs to catch. `r011_ping_removed.py` already does exactly this with `PingRequest\w*`.

## One line, one finding

A dispatcher can carry both signals on the same line:

```ts
case "notifications/initialized": return this.onInitializedNotification();
```

That is a single handshake implementation, so it is reported once. Findings are grade penalties, and double-charging one line for one problem overstates it — same dedupe R001 does for `const mcpSessionId = req.headers["mcp-session-id"]`.

## False positives

`initialize` on its own is one of the most overloaded words in software, so nothing here matches it — only the SDK's compound names. There is a test asserting a plain `db.initialize()` stays silent.

## Tests

Six new tests in `tests/test_typescript.py`: the SDK-name half, the wire half, the dual-signal line, a migrated server (`server/discover`), a file naming the handshake only in comments and JSDoc, and the unrelated `initialize()` case.

Full suite: **229 passed** (223 on `main` + 6). Verified end to end — `mcp-migrate check` on a TypeScript server now reports R009, and the coverage line moves to `10 of 21`.

---

*Supersedes #78, which was opened from the wrong account.*

### DIFF
--- a/src/mcp_migrate/rules/r009_initialize_handshake_removed.py
+++ b/src/mcp_migrate/rules/r009_initialize_handshake_removed.py
@@ -12,6 +12,32 @@ HANDSHAKE_CODE_RX = re.compile(
     r"\bInitializeRequest\b|\bInitializeResult\b|\bInitializedNotification\b"
 )
 
+# --- TypeScript -----------------------------------------------------------
+#
+# The same three SDK names, with one difference that matters: the TS SDK
+# exports a Zod schema alongside the inferred type, and the schema is the
+# name a server actually writes --
+# `server.setRequestHandler(InitializeRequestSchema, ...)`.
+# `\bInitializeRequest\b` cannot match inside `InitializeRequestSchema`
+# (there is no word boundary before `Schema`), so a trailing-suffix match
+# is required or this port would find nothing on the one shape it most
+# needs to catch. Same treatment `r011_ping_removed.py` gives
+# `PingRequest\w*`, for the same reason.
+TS_HANDSHAKE_CODE_RX = (
+    r"\bInitializeRequest\w*|\bInitializeResult\w*|\bInitializedNotification\w*"
+)
+
+# The wire name is spelled the same in both languages, and needs
+# `search_wire` in both for the same reason -- see the note in
+# `_check_python`.
+WIRE_RX = r"notifications/initialized"
+
+MESSAGE_CODE = (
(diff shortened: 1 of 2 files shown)


resuaico:
Closing — R009's TypeScript port already landed in main via #78, so this is the same change arriving twice and now conflicts. Opening a fresh port from this account instead.

Issue #90, opened 2026-08-04
R001 runs its identifier pattern through search_wire, so a log string reports breaking


dheerajjha:
```ts
console.info("Ignoring obsolete mcpSessionId metadata");
```

That line reports **R001 `breaking`**. It is a log message in a server that has already removed session handling — arguably the most migrated code possible — and we tell its author their transport is broken.

### Cause

R001 runs its *identifier* pattern through `search_wire()`. That mode deliberately keeps ordinary string literals, because it exists for JSON-RPC method names, which only ever live inside strings. An identifier is the opposite case: `mcpSessionId` appearing inside a string is somebody *talking about* the thing, not referencing it.

The three modes and what they're for are documented in [`rules/base.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/base.py):

- `search_code()` — skips comments **and** strings. **For code identifiers.**
- `search_wire()` — skips comments, keeps strings. For wire method names.
- `search()` — raw. Almost never right.

R001 legitimately needs *both*: the identifier half (`mcpSessionId`, `MCP_SESSION_ID`) through `search_code()`, and the header-string half (`"mcp-session-id"` in a `headers[...]` access) through `search_wire()`. Right now the identifier half is going through the wrong one.

### The fix

Split the two halves so each uses its own mode. The existing header-literal detection must keep working — that's the half that catches `req.headers["mcp-session-id"]`, which is the most common real occurrence.

### Done when

- A log/comment/error-message string mentioning `mcpSessionId` produces nothing.
- `const sid = req.headers["mcp-session-id"]` still reports.
- `const mcpSessionId = ...` as actual code still reports.
- Tests cover all three.

R001 is the oldest rule here and one of the most frequently triggered, so this is worth getting exactly right. Small change, high blast radius — ask in the thread if the split isn't obvious from the code.

Issue #88, opened 2026-08-04
Wire patterns lack an end boundary: `"roots/listeners"` reports as `roots/list`


dheerajjha:
`search_wire()` deliberately keeps ordinary string literals — that's what makes it the right mode for JSON-RPC method names. The consequence is that **every** string in the project is in scope, and a method-name pattern with no end boundary matches longer, unrelated strings.

### Reproduced, all five on `main`

| Rule | Unrelated string that reports a finding |
| --- | --- |
| R007 | `const route = "roots/listeners";` |
| R009 | `const topic = "notifications/initializedElsewhere";` |
| R012 | `const metric = "logging/setLevelLatency";` |
| R018 | `const endpoint = "roots/listeners";` |
| R019 | `const endpoint = "tasks/listeners";` |

`/roots/listeners` and `/tasks/listeners` are entirely plausible HTTP routes in a server that has correctly migrated. We would report `breaking` on it.

This is the same shape already found in #58's R013 port (`"/resources/subscribers"` matching `resources/subscribe`), so it's a pattern worth fixing everywhere at once rather than one rule at a time.

### The fix

Require the method name to end where it says it ends:

```python
# was:  r"tasks/list|tasks/result"
TASKS_WIRE_RX = r"\btasks/(?:list|result)(?![\w/-])"
```

`(?![\w/-])` is the important part: it rejects `tasks/listeners` (word char) and `tasks/list/extra` (slash) while still allowing `"tasks/list"` at the end of a string, followed by a quote, or inside `{"method": "tasks/list"}`.

### Worth thinking about in the PR

A shared helper would stop this recurring — something like `wire_method(*names)` in `rules/base.py` that builds the bounded pattern, so the next port cannot get it wrong. That's arguably the better fix, and it's what I'd lean toward, but it touches shared code so say what you're planning before you write it.

Each rule needs a test with the longer-string form asserting silence, plus the existing positive case still passing.

dheerajjha:
**Correction: this is not TypeScript-only.** Reproduced against the **Python** paths on `main` as well:

```python
ENDPOINT = "tasks/listeners"          # R019 fires
METRIC   = "logging/setLevelLatency"  # R012 fires
```

Both are plain strings in Python source with no MCP involvement. A fix that only touches the TypeScript branch leaves this live in the language all 14 published registry grades were computed in — please do both paths in one PR, with a test each side.

[...]

PR #93, opened 2026-08-04
fix(scan): skip *.examples.ts documentation modules by default

### LINKED ISSUE #83: Scan skips test files but not `*.examples.ts`, which is a docs convention
`*.examples.ts` is a documentation convention, and we scan it as production code.

Found by pointing the freshly ported TypeScript rules at
[`modelcontextprotocol/typescript-sdk`](https://github.com/modelcontextprotocol/typescript-sdk).
It has **68** of these files, and 7 findings came from them.

### Why they are documentation

Straight from the top of `packages/server/src/server/mcp.examples.ts`:

```ts
/**
 * Type-checked examples for `mcp.ts`.
 *
 * These examples are synced into JSDoc comments via the sync-snippets script.
 * Each function's region markers define the code snippet that appears in the docs.
 *
 * @module
 */
```

They are prose that happens to compile. A deprecated call inside one is a *docs* problem
at worst, and usually not even that — an example showing the legacy shape next to the new
one is exactly what good migration docs look like.

This is the same bug class as scanning `__tests__/`, which we already fixed in
[`scan.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/scan.py):
a file whose *purpose* is to demonstrate the protocol surface will always look like a
server implementing it.

### The findings it currently produces

| Rule | File |
| --- | --- |
| R003 | `packages/middleware/node/src/streamableHttp.examples.ts:52` |
| R003 | `packages/server/src/server/mcp.examples.ts:138` |
| R003 | `packages/server/src/server/middleware/bearerAuth.examples.ts:26` |
| R004 | `packages/client/src/client/client.examples.ts:144` |
| R007 | ×2 |
| R018 | ×1 |

### How to do it

`TEST_FILE_PATTERNS` in `scan.py` already does exactly this job for `*.test.ts` /
`*.spec.ts`. Add the examples spelling:

```python
"*.examples.ts", "*.examples.tsx", "*.examples.mts", "*.examples.cts",
```

Worth a thought in the PR: the constant is named `TEST_FILE_PATTERNS` and these are not
tests. Either rename it to something like `NON_PRODUCTION_FILE_PATTERNS`, or add a
separate constant and a comment saying why both exist. Your call — say which you picked
and why.

Also consider `*.example.ts` (singular). I have not seen it in the wild yet, so only add
it if you can point at a real repo using it — speculative patterns are how a skip-list
starts hiding real findings.

### Done when

A `*.examples.ts` file containing a legacy pattern produces no findings, `--include-tests`
still scans it, and there is a test in `tests/test_typescript.py` covering both directions.

Ask in the thread if anything is unclear — PRs get reviewed the same day.

KhyFee:
﻿## Summary
- Skip `*.examples.ts` / `.tsx` / `.mts` / `.cts` by default (closes #83)
- Same skip path as other non-production patterns; `--include-tests` still scans them
- Kept name `TEST_FILE_PATTERNS` with a comment — one skip list, one `--include-tests` toggle (avoided a second parallel constant)

## Test plan
- [x] `pytest tests/test_typescript.py -k examples -q`

### DIFF
--- a/src/mcp_migrate/scan.py
+++ b/src/mcp_migrate/scan.py
@@ -35,10 +35,14 @@ TEST_DIR_SEGMENTS = {
 # The `*.test.*` / `*.spec.*` pair is close to universal in JS/TS and is
 # the dominant convention for tests that sit *beside* the module they
 # cover, where no directory segment can give them away.
+# Name is historical (tests first). Also covers type-checked docs
+# examples (*.examples.ts) used by e.g. modelcontextprotocol/typescript-sdk
+# for JSDoc snippet sync — prose that compiles, not production servers.
 TEST_FILE_PATTERNS = (
     "test_*.py", "*_test.py", "conftest.py",
     "*.test.ts", "*.test.tsx", "*.test.mts", "*.test.cts",
     "*.spec.ts", "*.spec.tsx", "*.spec.mts", "*.spec.cts",
+    "*.examples.ts", "*.examples.tsx", "*.examples.mts", "*.examples.cts",
 )
 
 
--- a/tests/test_typescript.py
+++ b/tests/test_typescript.py
@@ -1130,6 +1130,8 @@ def test_ts_scanner_handles_escapes_and_unterminated_strings():
     "src/server.test.ts",
     "src/server.spec.ts",
     "src/server.test.tsx",
+    "src/server.examples.ts",
+    "src/server.examples.tsx",
 ])
 def test_typescript_test_code_is_skipped_by_default(tmp_path, rel):
     path = tmp_path / rel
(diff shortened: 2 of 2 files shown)


dheerajjha:
Merged, and thank you — twenty-five minutes from issue to correct fix.

What made it easy to merge: **you handled both edge cases I flagged rather than only the one I asked for.** I verified against a temp project before merging:

| File | Scanned? | |
| --- | --- | --- |
| `src/mcp.examples.ts` | no | the target |
| `src/client.examples.tsx` | no | tsx variant |
| `src/server.ts` | yes | ordinary source, untouched |
| `src/examples.ts` | **yes** | bare name is not the convention |
| `src/my.example.ts` | **yes** | singular `.example.` left alone |

The last two are the ones that mattered. A skip-list that quietly swallows `examples.ts` is how a scanner starts hiding real findings, and you didn't add the speculative singular pattern I explicitly warned against. `--include-tests` still picks them up, which I also checked.

And you answered the naming question in a comment instead of leaving it implicit:

```python
# Name is historical (tests first). Also covers type-checked docs
# examples (*.examples.ts) used by e.g. modelcontextprotocol/typescript-sdk
# for JSDoc snippet sync — prose that compiles, not production servers.
```

That's the right call. Renaming the constant would have been a bigger diff for no behavioural gain, and the comment stops the next reader wondering.

You're the eighth person to land a change here. If you want another, the two closest in shape are **#88** (wire patterns match `"roots/listeners"` as `roots/list` — needs an end boundary, five rules, same edit each time) and **#90** (R001 runs its identifier pattern through the wrong search mode, so a log string reports `breaking`). Both are bugs I reproduced this morning and both have the failing snippet in the issue.

Either way — thanks. Reviews here are same-day.

KhyFee:
﻿Thanks @dheerajjha — glad the examples skip matched both orientations. Appreciate the quick merge and the thorough verification matrix.

PR #94, opened 2026-08-04
feat(rules): port R008 (OpenTelemetry trace context) to TypeScript

### LINKED ISSUE #37: TypeScript: port R008 (Does not propagate OpenTelemetry trace context from _meta)

djubx:
Fixes #37.

Ports **R008** (`Does not propagate OpenTelemetry trace context from _meta`, `advisory`) to TypeScript. `languages = ("python", "typescript")`, `check()` splits on `project.language`, Python path unchanged.

Another evidence/absence rule, so it's a gate plus an evidence check — and each half needed a TypeScript answer.

## The gate: is OpenTelemetry actually in play?

Python uses `project.imports()`, which walks the AST — and the AST is Python-only (`SourceFile.tree` stays `None` for every other language), so this needed a textual equivalent.

A quoted `@opentelemetry/...` specifier is that equivalent. The string only ever appears in an `import` or `require`, so it needs no anchor to the keyword and can't be confused with prose about tracing.

`search_wire` keeps the string literal while dropping comments, so this does **not** open the gate:

```ts
// TODO: add "@opentelemetry/api" and start tracing tool calls.
```

That matters more here than in most rules — this gate opening is the only thing that lets R008 fire at all, so a comment opening it would produce a finding against a project that does no tracing whatsoever.

## The evidence: is the header read?

`traceparent` has two real spellings in TypeScript and **both** have to count:

```ts
const tp = request.params._meta?.["traceparent"];   // string key
const tp = meta.traceparent;                        // property access
```

`search_wire` is the only mode that sees both — `search_code` discards string tokens and would miss the first. Comments stay excluded either way, so *intending* to read the header is not reading it, and there's a test pinning that the rule still fires for:

```ts
// TODO: pull traceparent off _meta and seed the span context with it.
```

## One deliberate difference from the Python rule

The idiomatic OTel API consumes the header through the propagator rather than by name:

```ts
const parent = propagation.extract(context.active(), request.params._meta ?? {});
return context.with(parent, () => handle(request));
```

A server written this way propagates trace context **correctly** and may never spell `traceparent` anywhere. Flagging it would be a false positive against code that already does the right thing, so `propagation.extract(` counts as evidence.

That makes the TypeScript port strictly quieter than the Python one. Flagging it explicitly in case you'd rather they stayed identical — but it errs in the direction this project errs in, and the same gap arguably exists on the Python side (`propagate.extract` there) if it's worth a follow-up issue.

## Tests

Seven, in `tests/test_typescript.py`:

- fires on OTel imported with no `traceparent`
- silent for the string-key read
- silent for the property read
- silent for `propagation.extract`
- silent with no OpenTelemetry at all
- silent when OTel appears only in a comment (gate stays shut)
- **still fires** when `traceparent` appears only in a comment

Full suite: **268 passed**. Verified end to end — `mcp-migrate check` reports R008 on a TypeScript server that traces without reading `traceparent`.

### DIFF
--- a/src/mcp_migrate/rules/r008_trace_context.py
+++ b/src/mcp_migrate/rules/r008_trace_context.py
@@ -1,5 +1,47 @@
 from .base import Finding, Project, Rule
 
+# --- TypeScript -----------------------------------------------------------
+#
+# The Python gate is `project.imports()`, which walks the AST -- and the
+# AST is Python-only (`SourceFile.tree` stays None for every other
+# language), so the TypeScript side needs a textual equivalent.
+#
+# A quoted `@opentelemetry/...` specifier is that equivalent. The scope is
+# narrow on purpose: the string only appears in an `import`/`require`, so
+# matching it needs no anchor to the import keyword, and it cannot be
+# confused with ordinary prose about tracing. `search_wire` keeps string
+# literals (where the specifier lives) while dropping comments, so a
+# `// TODO: add @opentelemetry/api` does not read as "OpenTelemetry is in
+# use" -- which matters more here than in most rules, because this gate
+# opening is what lets the rule fire at all.
+TS_OTEL_SPECIFIER_RX = r"[\"'`]@opentelemetry/[^\"'`]*[\"'`]"
+
+# Evidence the header is actually read. Both spellings are real and both
+# have to count:
+#
+#   const tp = request.params._meta?.["traceparent"];   // string key
+#   const tp = meta.traceparent;                        // property access
+#
+# `search_wire` is the mode that sees both -- `search_code` discards string
+# tokens and would miss the first, while comments stay excluded either way,
+# so "we should read traceparent" still does not count as reading it.
+TS_TRACEPARENT_RX = r"\btraceparent\b"
+
+# The other correct way to do this, and the reason the TypeScript port is
+# quieter than the Python one: the idiomatic OTel API consumes the header
+# through the propagator rather than by name.
+#
+#   propagation.extract(context.active(), request.params._meta ?? {})
+#
+# A server written that way propagates trace context correctly and may
+# never spell `traceparent` anywhere. Flagging it would be a false positive
+# on code that already does the right thing, so the extract call counts as
+# evidence. `search_code` here: this one is a real call expression, not a
+# string.
+TS_PROPAGATION_EXTRACT_RX = r"\bpropagation\s*\.\s*extract\s*\("
+
+MESSAGE = "OpenTelemetry is in use but `traceparent` is never read from `_meta`."
+
 
 class NoTraceContextPropagation(Rule):
     id = "R008"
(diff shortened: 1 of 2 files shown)


dheerajjha:
Thanks for this — R008 was one of only three TypeScript ports nobody had touched, so it's a genuinely useful one to pick up.

**Heads up: it needs a rebase.** `main` moved under you while you were writing it — I merged five PRs this morning, and `tests/test_typescript.py` is the file every port touches, so GitHub now shows this as conflicting.

```
git fetch origin
git rebase origin/main
# resolve tests/test_typescript.py -- see below
git push --force-with-lease
```

**On resolving it:** the conflict will be additive on both sides — your R008 import and test block against everyone else's. Keep **both**. The one that catches people out is the import block, where both sides end on an unclosed `(` and appear to share a single `)`:

```python
from mcp_migrate.rules.r007_deprecated_features import (
    DeprecatedCoreFeatures,
<<<<<<< HEAD
)
from mcp_migrate.rules.r008_trace_context import (
    TraceContextNotPropagated,
)
```

Give each import its own closing paren. Naively concatenating the two halves produces `SyntaxError: '(' was never closed`, which I did to myself yesterday and lost twenty minutes to.

I have not reviewed the substance yet — I will do that once it rebases cleanly, same day. One thing I will be checking, so you may as well know now: R008 is about trace context propagated from `_meta`, and `traceparent` / `tracestate` are W3C names that appear in plenty of non-MCP HTTP code. I audited the rules this morning and found eighteen false positives across the suite, most of them from exactly this shape — a real name matched without enough context to know it's MCP. If your port has a snippet like

```ts
const headers = { traceparent: span.toTraceparent() };  // ordinary OTel, no MCP
```

staying silent, say so in the PR and it'll go in faster.

Ping me when it's rebased.

dheerajjha:
Merged, and the rebase came back clean — thanks for turning that round quickly.

I'll be honest about my own review: I probed this expecting to find the `traceparent` false positive I warned you about, and **my probe was wrong, not your port.** I fed it a bare MCP handler with no OpenTelemetry anywhere and expected a finding; R008 gates on an OTel import, so silence was correct. Your gate is right and mine was a bad test.

What I did verify:

- `const headers = { traceparent: span.toTraceparent() }` in ordinary OTel HTTP code with no MCP — **silent** ✓
- a commented-out `@opentelemetry/api` import doesn't open the gate ✓

The part I want to call out is this, from your comments:

> A server written that way propagates trace context correctly and may never spell `traceparent` anywhere. Flagging it would be a false positive on code that already does the right thing, so the extract call counts as evidence.

You found a false positive *in your own port* and closed it before submitting, with a test (`test_r008_stays_silent_when_the_propagator_extracts_the_context`). That is the single most useful thing a contributor can do here — I audited the rule suite this morning and found eighteen false positives, every one of which existed because nobody asked that question at write time.

Seven tests, both signal spellings, the propagator case, the comment case. **TypeScript is now 16 of 21.** Five rules left: R002, R013, R014, R017, R021.

Issue #87, opened 2026-08-04
Unbounded \w* suffix makes 4 rules fire on unrelated identifiers like `PingRequester`


dheerajjha:
Four rules use an unbounded `\w*` suffix on an SDK type name, and it matches more than it means to.

The suffix exists for a good reason. TypeScript servers reference the Zod schema, not the bare type:

```ts
server.setRequestHandler(PingRequestSchema, async () => ({}));
```

`\bPingRequest\b` cannot match inside `PingRequestSchema` — there's no word boundary before `Schema` — so the port would find nothing on the one shape it most needs to catch. `\w*` fixes that, and then keeps going.

### Reproduced, all four on `main`

| Rule | Unrelated code that reports `breaking` |
| --- | --- |
| R007 | `class CreateMessageRequester { send() {} }` |
| R009 | `class InitializeRequester { run() {} }` |
| R011 | `class PingRequester { send() {} }` |
| R012 | `class SetLevelRequester { send() {} }` |

None of those is MCP. `…Requester` is ordinary English and an extremely common class-name shape.

### The fix

Bound the suffix to the names that actually exist in the SDK rather than accepting anything:

```python
# was:  r"\bPingRequest\w*"
PING_NAMES_RX = re.compile(r"\bPingRequest(?:Params|Schema)?\b")
```

Check the real SDK surface before settling the alternation — `Request`, `RequestParams`, `RequestSchema`, `Result`, `ResultSchema` covers what I've seen, but confirm against [`@modelcontextprotocol/sdk`](https://github.com/modelcontextprotocol/typescript-sdk) rather than trusting this list.

### Please do all four in one PR

They are the same edit four times and reviewing them together is easier than four separate threads. Each needs a test asserting the `…Requester` form stays silent **and** that the `…Schema` form still fires — the second half is what stops someone "fixing" this by deleting the suffix entirely.

Found by running a false-positive audit over the TypeScript rule paths; a false positive is the failure mode this project treats as worse than a miss, which is why this is a bug and not an enhancement.

dheerajjha:
**Correction: this is not TypeScript-only.** I labelled it `typescript` because that is where the audit ran, but I have since reproduced the same class against the **Python** paths on `main`:

```python
class PingRequester:          # R011 -> "References the removed PingRequest handler."
    def send(self): ...
```

[...]

PR #95, opened 2026-08-04
fix(r001): use search_code for identifier patterns in TypeScript

### LINKED ISSUE #90: R001 runs its identifier pattern through search_wire, so a log string reports breaking

lesbass:
R001's `_check_ts` was running both the identifier pattern (`mcpSessionId`, `MCP_SESSION_ID`) and the header-literal pattern (`"mcp-session-id"`) through `search_wire`. That mode keeps string literals, so a log string like:

```ts
console.info("Ignoring obsolete mcpSessionId metadata");
```

would report as `breaking`.

## Change

Split the two halves so each uses the right search mode:

- **identifier pattern** (`mcpSessionId`, `MCP_SESSION_ID`) → `search_code` — skips string tokens, so a log/error-message string naming the identifier is not a false positive.
- **header-literal pattern** (`"mcp-session-id"`) → `search_wire` — keeps string literals (where the header name lives) and drops comments/prose.

## Tests

Three new tests cover the exact scenarios from the issue:

- `test_r001_ignores_the_identifier_in_a_log_string` — `console.info("... mcpSessionId ...")` produces nothing.
- `test_r001_finds_header_access_in_typescript` — `req.headers["mcp-session-id"]` still fires.
- `test_r001_finds_identifier_as_code_in_typescript` — `const mcpSessionId = ...` in code still fires.

All 272 existing tests continue to pass.

Closes #90.

### DIFF
--- a/src/mcp_migrate/rules/r001_session_id_removed.py
+++ b/src/mcp_migrate/rules/r001_session_id_removed.py
@@ -53,19 +53,26 @@ class SessionIdRemoved(Rule):
         ]
 
     def _check_ts(self, project: Project) -> list[Finding]:
-        # search_wire, not search_code: in TypeScript the header name lives
-        # in a string literal, which search_code discards wholesale -- it
-        # would find nothing at all. search_wire keeps string literals and
-        # drops comments, which is exactly the split needed here.
         seen: set[tuple[str, int]] = set()
         out: list[Finding] = []
-        for pattern in (TS_IDENT_RX, TS_HEADER_RX):
-            for f, line, text in project.search_wire(pattern, flags=re.IGNORECASE):
-                # The two patterns can both hit the same line
-                # (`const mcpSessionId = req.headers["mcp-session-id"]`);
-                # that is one problem, not two.
-                if (str(f.path), line) in seen:
-                    continue
-                seen.add((str(f.path), line))
-                out.append(self.finding(self.MESSAGE, f, line, text))
+
+        # Identifiers (mcpSessionId, MCP_SESSION_ID) use search_code:
+        # a log/comment/error-message string that merely mentions the
+        # identifier isn't a real reference to it.
+        for f, line, text in project.search_code(TS_IDENT_RX, flags=re.IGNORECASE):
+            seen.add((str(f.path), line))
+            out.append(self.finding(self.MESSAGE, f, line, text))
+
+        # Header-string literals ("mcp-session-id") use search_wire:
+        # the header name lives in a string literal, which search_code
+        # discards wholesale -- it would find nothing at all. search_wire
+        # keeps string literals and drops comments.
+        for f, line, text in project.search_wire(TS_HEADER_RX, flags=re.IGNORECASE):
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged. Verified all five cases before it went in:

| Case | Result |
| --- | --- |
| `console.info("Ignoring obsolete mcpSessionId metadata")` | silent ✓ |
| `req.headers["mcp-session-id"]` | fires ✓ |
| `const mcpSessionId = crypto.randomUUID()` | fires ✓ |
| `MCP_SESSION_ID` used as code | fires ✓ |
| the name in a comment | silent ✓ |

The half that matters is the second row. The obvious way to "fix" this issue is to move everything to `search_code`, which would have silently killed header-literal detection — the most common real occurrence of the bug — and the tests would still have passed. You split the two patterns instead of moving both, which is the whole point.

R001 is the oldest rule here and the most frequently triggered, so this had the widest blast radius of anything in the false-positive backlog.

`dev/fp-audit.py`, the standing corpus of every known false positive, now reads **17 reproduce · 1 refuted**. Yours is the 1.

If you want another: **#88** (five rules match `"roots/listeners"` as `roots/list` — needs an end boundary) is the same shape and touches both language backends. **#87** is four rules with an unbounded `\w*`. Both have the failing snippet in the issue.

Thanks — that was a clean fix on a subtle bug.

Issue #92, opened 2026-08-04
Re-verify the 14 published grades: 5 false-positive classes reproduce in the Python paths


dheerajjha:
We publish 14 grades in [`registry/servers/`](https://github.com/dheerajjha/mcp-migrate/tree/main/registry/servers). All 14 are Python, all computed with `mcp-migrate 0.1.1`.

An audit today found **18 false-positive classes** across the rules (#87, #88, #89, #90, #91). I initially assumed they were TypeScript-only. They are not — five reproduce against the **Python** paths, which is the language every one of those 14 grades was computed in:

```python
class ServerCapabilities:              # R005 fires
    max_connections: int = 4

class ConnectionPool:
    def register_client(self, c): ...  # R020 fires

ENDPOINT = "tasks/listeners"           # R019 fires
METRIC   = "logging/setLevelLatency"   # R012 fires

class PingRequester:                   # R011 fires
    def send(self): ...
```

Every one of those inflates a score downward, and score is what the grade is derived from.

### Why this is urgent rather than tidy-up

Two of the published grades are low, and they are exactly where a false positive does damage:

| Entry | Grade | Score |
| --- | --- | --- |
| `mcp-atlassian` | **C** | 60 |
| `mcp-server-git` | **D** | 54 |

`mcp-atlassian` is the repo an early iteration of this tool already graded **F/0 entirely on false positives** — Jira REST calls tripping the routing-header rule, `Mcp-Session-Id` inside a CLI help string. That incident is the origin of this project's first working principle. If its C is also inflated by false positives we have made the same mistake twice on the same repo, while a PR of ours sits open on it.

The README states plainly: *"No grade is published that hasn't been reproduced"* and *"Every entry in `registry/servers/` comes from a scan someone ran and read."* That was true when they were written. It needs to be true again after the fixes land.

### What to do

1. **After #87/#88/#89/#90/#91 land**, re-scan all 14 registry repos at the pinned commit each entry names.
2. Diff old score against new. Anything that moves gets its `grade`, `score` and `checked_with` updated.
3. **Hand-read every finding that survives** on the C and D entries specifically, against real source, before republishing. Not a spot check — all of them.
4. If a grade rises, say so in the commit message. Publishing a correction loudly is the only thing that makes the original number trustworthy.

### Doing it before the fixes land is also useful

[...]

PR #96, opened 2026-08-04
fix(r019): add TODO-annotation fixer for removed tasks methods

### LINKED ISSUE #27: Fixer: tasks/list and blocking tasks/result removed (R019)
No fixer exists yet for R019 ([`src/mcp_migrate/rules/r019_tasks_polling_replaces_blocking_result.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r019_tasks_polling_replaces_blocking_result.py)),
which flags removed `tasks/list` and the removed blocking `tasks/result`.

**Spec:** SEP-2663 -- https://modelcontextprotocol.io/specification/2026-07-28/changelog

**Why this is hard:** converting a blocking wait into a `tasks/get` +
`tasks/update` polling loop is a real control-flow rewrite (see
`cookbook/11-tasks-polling.md`), not a mechanical edit. A `review`-confidence
TODO-annotation on the old handler (same pattern as the R009/R011/R012/R013
fixer issues) is the most defensible starting scope.

with whatever scope you can defend as genuinely mechanical, with fixtures
and tests per `tests/test_fixers.py`.

syf2211:
## Summary

Adds a `review`-confidence fixer for R019 that comments out lines referencing removed `tasks/list` and blocking `tasks/result` usage (SDK types and wire string literals) and leaves a TODO pointing at the spec and cookbook.

## Motivation

Issue #27 — blocking `tasks/result` cannot be mechanically rewritten into a polling loop, but flagged call sites benefit from the same TODO-annotation pattern used by R014.

## Changes

- New `TasksPollingFixer` in `src/mcp_migrate/fixers/r019_tasks_polling.py`
- Wire matching scoped to quoted `"tasks/list"` / `"tasks/result"` literals (not docstring prose)
- Tests: before/after transformation, block-opener skip, docstring prose guard, idempotency

## Tests

`python3 -m pytest tests/test_fixers.py -v` — 47/47 pass

## Notes

Handler signatures that end with `:` are intentionally left for human rewrite, matching R014.

Fixes #27

### DIFF
--- a/src/mcp_migrate/fixers/r019_tasks_polling.py
+++ b/src/mcp_migrate/fixers/r019_tasks_polling.py
@@ -1,0 +1,91 @@
+"""Fixer for R019 -- removed tasks/list and blocking tasks/result.
+
+There is no mechanical replacement for a blocking wait: the 2026-07-28
+spec replaces it with a tasks/get + tasks/update polling loop, and tasks
+moved into the io.modelcontextprotocol/tasks extension. So this fixer does
+the one thing that is safe without understanding the surrounding control
+flow -- flag each flagged call site with a loud TODO and comment out the
+line when that cannot break a multiline expression. Confidence "review":
+every flagged site still needs a human pass.
+"""
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from .base import Fixer, FixResult
+
+SPEC_URL = "https://modelcontextprotocol.io/specification/2026-07-28/changelog"
+COOKBOOK = "cookbook/11-tasks-polling.md"
+TODO = (
+    "# TODO(mcp-migrate): tasks/list and blocking tasks/result are removed; "
+    f"poll with tasks/get + tasks/update and declare io.modelcontextprotocol/tasks "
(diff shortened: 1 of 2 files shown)


dheerajjha:
Merged — that's your seventh, and the R019 fixer means every removed-method rule now has one.

Reviews are same-day, so if you want the next thing: **#92** is the one I'd most like your eyes on. All 14 published registry grades are Python, and I confirmed this morning that five false-positive classes reproduce in the Python paths — including R019's, where `"tasks/listeners"` reports as `tasks/list`. `mcp-atlassian` is published at **C/60** and `mcp-server-git` at **D/54**, and those grades were computed with rules that have known false positives in them.

You have better instincts for this than most: **#66 exists because you documented a false negative in #63 instead of quietly working around it.** That is exactly the disposition #92 needs — re-scanning and hand-reading which findings survive, rather than assuming the new number is the right one.

No pressure if you'd rather keep porting. #87 and #88 are both mechanical and both touch rules you've already worked in.

PR #97, opened 2026-08-04
feat: Add R021 fixer – rewrite older JSON Schema dialect pins to 2020-12 (#29)

### LINKED ISSUE #29: Fixer: older JSON Schema dialect pinned (R021)

IronLad123:
Closes #29

## Summary

Adds a `safe`-confidence `Fixer` subclass (`OldJSONSchemaDialectFixer`) for **R021** — the rule that flags explicit older JSON Schema dialect pins (`draft-07`, `2019-09`, etc.).

## What the fixer does

Scans each line for a `"$schema"` key assignment containing an older dialect URL or short token, and rewrites it to the canonical 2020-12 URL:

```
- "$schema": "http://json-schema.org/draft-07/schema#"
+ "$schema": "http://json-schema.org/draft/2020-12/schema"
```

The rewrite is anchored to a `"$schema":` key on the same line — so comments, log messages, or values under a different key that merely _contain_ the old URL string are deliberately left untouched.

## Confidence: `safe`

- Transformation is an exact string substitution; no AST round-tripping
- Match requires the `"$schema"` key to be visible on the same line → no false-positive risk from ambient mentions
- Already-2020-12 lines are skipped (idempotency guard)

## Files changed

| File | Purpose |
|---|---|
| `src/mcp_migrate/fixers/r021_json_schema_2020_12_required.py` | New `OldJSONSchemaDialectFixer` class (auto-discovered) |
| `tests/test_fixers.py` | 8 new test cases: draft-07 URL, draft-04 URL, 2019-09 URL, short form, comment left alone, wrong key left alone, idempotency, already-2020-12 unchanged |
| `tests/fixtures/fixer_roundtrip/server.py` | Added a `$schema draft-07` pin so the fixer round-trip test exercises R021 |

## Tests

All 233 tests pass (`python3 -m pytest -q`). The fixer file passes `ruff check` clean.

### DIFF
--- a/src/mcp_migrate/fixers/r021_json_schema_2020_12_required.py
+++ b/src/mcp_migrate/fixers/r021_json_schema_2020_12_required.py
@@ -1,0 +1,116 @@
+"""Fixer for R021 -- older JSON Schema dialect pin.
+
+SEP-2106 (2026-07-28) requires MCP implementations to support at least
+JSON Schema 2020-12 for inputSchema/outputSchema. Explicitly pinning an
+older draft (draft-07, 2019-09, etc.) via a ``$schema`` URL key is a
+mechanical, unambiguous rewrite: the old dialect URL string becomes the
+canonical 2020-12 URL.
+
+Confidence ``safe`` because:
+- The match is anchored to a ``"$schema"`` key on the same line, so the
+  regex cannot hit a comment, a log message, or an unrelated dialect
+  mention that just happens to contain the old URL string.
+- The transformation is an exact string substitution of the dialect
+  segment of the URL; no structural changes to the surrounding code.
+- When the pattern is ambiguous (e.g. the old dialect string appears on a
+  line that does *not* look like a ``"$schema"`` assignment), the fixer
+  backs off and leaves the line untouched rather than guess.
+"""
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from .base import Fixer, FixResult
+
+# The authoritative 2020-12 $schema URL (no trailing #)
+_TARGET_DIALECT = "http://json-schema.org/draft/2020-12/schema"
+
+# Matches the old-dialect string only when it appears on the same line as
+# a "$schema": assignment (JSON, Python dict literal, YAML mapping, etc.).
+# The look-behind ensures there is a "$schema" key somewhere to the left;
+# the pattern then captures the old URL so we can replace just the URL
+# portion and leave surrounding quotes/whitespace intact.
+#
+# Matches:
+#   "$schema": "http://json-schema.org/draft-07/schema#"
+#   "$schema": "http://json-schema.org/draft-04/schema"
+#   "$schema": "https://json-schema.org/draft/2019-09/schema"
+#   "$schema" = "http://json-schema.org/draft-07/schema#"   (Python)
+#
+# Does NOT match:
+#   # old schema was draft-07         (no $schema key on the same line)
+#   some_other_key = "draft-07"       (no $schema key on the same line)
+_SCHEMA_KEY_RX = re.compile(
+    r"""
+    \"\$schema\"          # the "$schema" key (JSON / Python dict key)
+    \s*[:=]\s*            # colon (JSON/YAML) or equals (Python)
+    \"                    # opening quote of the value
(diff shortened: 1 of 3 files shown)


dheerajjha:
Good fixer — the `$schema`-key anchoring is exactly the right call, and the docstring explaining *why* it's `safe` is the kind of thing I wish every fixer had. Idempotency guard works, and I confirmed the checker accepts the fixer's output, so there's no fix-then-still-fails loop.

**One thing blocks it.** It emits `http://`:

```python
# The authoritative 2020-12 $schema URL (no trailing #)
_TARGET_DIALECT = "http://json-schema.org/draft/2020-12/schema"
```

Measured:

```
in : {"$schema": "https://json-schema.org/draft-07/schema#"}
out: {"$schema": "http://json-schema.org/draft/2020-12/schema"}
```

Two problems in one line.

**1. That isn't the 2020-12 identifier.** JSON Schema moved to `https` for 2019-09 and 2020-12; the canonical meta-schema URI is `https://json-schema.org/draft/2020-12/schema`. Only draft-07 and earlier use `http://…/draft-07/schema#`. And `$schema` is matched as an *opaque identifier string*, not fetched as a URL — so a validator comparing it exactly will not recognise the `http` spelling as 2020-12. The fixer would leave a server declaring a dialect nothing implements.

Your own example comment two lines below already has this right:

```python
#   "$schema": "https://json-schema.org/draft/2019-09/schema"
```

**2. It silently downgrades the scheme.** A user who had `https` gets `http` back. For a fixer marked `safe`, rewriting someone's URL scheme is not something that should happen as a side effect of a dialect bump — quite apart from 2020-12, that's a change nobody asked for.

The fix is the constant:

```python
_TARGET_DIALECT = "https://json-schema.org/draft/2020-12/schema"
```

Please add a test asserting the output is `https` — this is precisely the sort of thing that gets silently reintroduced, and it's invisible in review because both strings look almost identical.

While you're in there, one to consider (your call, not a blocker): should the fixer *preserve* the input's scheme rather than always emitting `https`? I don't think so — 2020-12's identifier is `https` regardless of what the old pin used, so normalising is correct. But if you disagree, say why and I'll listen.

**One correction to something I nearly told you:** I first thought draft-04 wasn't handled. That was my test, not your code — I'd written it as a bare assignment rather than a `$schema` key, so the anchoring correctly declined it. Re-tested properly and draft-04 works fine.

Fix the constant and add the test and I'll merge same day.

IronLad123:
Thank you for the thoughtful review @dheerajjha! You're completely right — JSON Schema 2020-12 uses `https` as its canonical identifier string, so emitting `http` was both non-canonical and an unintended scheme downgrade.

I have updated `_TARGET_DIALECT` in `src/mcp_migrate/fixers/r021_json_schema_2020_12_required.py` to `"https://json-schema.org/draft/2020-12/schema"` and updated `test_r021_rewrites_draft07_url` in `tests/test_fixers.py` to assert against the explicit literal `"https"` URL string. All R021 tests pass cleanly. Appreciate your guidance!

Issue #84, opened 2026-08-04
The TypeScript SDK gets a meaningless grade -- 670 of 713 findings come from it


dheerajjha:
Companion to #31, which says the same thing about the Python SDK. Now measured for TypeScript.

I ran the 14 TypeScript-enabled rules against
[`modelcontextprotocol/typescript-sdk`](https://github.com/modelcontextprotocol/typescript-sdk)
and the official servers repo. Result:

| Project | Findings |
| --- | --- |
| **typescript-sdk** | **670** |
| servers/src/everything | 41 |
| servers/src/filesystem | 2 |

**670 of 713 findings — 94% — come from the SDK.** And almost none of them are wrong in
the narrow sense. `packages/server/src/server/streamableHttp.ts` really does reference
`Mcp-Session-Id`; `LoggingMessageNotification` really is in there. Of course it is: an SDK
implements the protocol, including the parts being deprecated, and it has to keep
implementing them for every user who has not migrated yet.

The rules are behaving correctly. The **grade** is the category error.

### Why this matters more than it looks

The grade is the thing we publish about other people's projects, and an SDK is the single
most visible repo in any ecosystem. A `README` badge reading F on the reference
implementation is not a bug report, it's a false accusation with our name on it — and it
is the fastest possible way to lose the ecosystem's trust in every other grade we publish.

Related: our own working principle is that an unearned grade is as damaging as an
undeserved finding. This is the same failure with the sign flipped.

### The actual question

This needs a decision before it needs code, and I would rather discuss it in the thread
than have someone guess:

1. **Detect that the target is a protocol library, not a server**, and report findings
   without a grade. What is the signal? A `package.json` name under
   `@modelcontextprotocol/`? Exporting `Server`/`McpServer` rather than constructing one?
   Both are guessable and both have false positives.
2. **Let a project opt out of grading** with a config key, and use it ourselves in the
   registry. Honest and simple, but only works for projects that know we exist.
3. **Grade only what a server controls**, and treat transport/protocol plumbing as
   out of scope. Most principled, hardest to define.

I lean toward (1) with a *conservative* detector plus (2) as the escape hatch, but I hold
that loosely.

### Done when

`mcp-migrate check` on a protocol library reports its findings and explicitly declines to
grade, with a one-line reason the reader can act on — and there is a test proving a normal
server is still graded.

**This is a design discussion first.** Comment with which option you would take and why;
I will reply the same day. Do not open a PR until we have agreed the shape, so nobody
[...]

PR #102, opened 2026-08-04
feat: decline grade on protocol SDKs and list findings as migration checklist (fixes #31)

### LINKED ISSUE #31: mcp-migrate grades the official Python SDK F/0, which is meaningless
Running the scanner on `modelcontextprotocol/python-sdk` returns **Grade F (0/100) with 561 findings** across 321 files: 130× R018, 71× R013, 63× R009, 55× R004, 54× R007, 49× R001, 47× R012, 32× R019.

*(Measured with mcp-migrate 0.1.1 against python-sdk `a4f4ccd`, 2026-07-29. Reproduce with `mcp-migrate check path/to/python-sdk --json`.)*

Every one of those is technically a true match and every one is meaningless. An SDK *has* to implement `initialize`, `ping`, `logging/setLevel` and `resources/subscribe` — it implements **both** sides of the deprecation so its users can migrate gradually. The tool is applying server rules to a protocol implementation, which is a category error. Anything vendoring SDK internals hits the same problem.

What's needed is a way to tell "this repo *implements* MCP" from "this repo *uses* MCP", and then either skip grading or switch to a different rule set.

Open questions worth arguing about in the thread — none of these is decided, and the discussion is the useful part:

- Is it a package-name check (`mcp`, `fastmcp`, and then what about forks)?
- A heuristic on the ratio of protocol-*definition* code to protocol-*consumption* code?
- An opt-out marker file the SDK could add?
- Something else entirely?

Whatever it is, it has to **fail toward silence**. This project's standing rule is that a false positive is worse than a missed finding — we publish grades about other people's code, so a wrong "you're broken" costs more than a quiet miss. If the detector is unsure whether something is an SDK, the safe direction is to say less, not more.

**Done when:** scanning the Python SDK produces something defensible instead of F/0; scanning an ordinary server is unaffected; there is a regression test for both.

Good first issue in the sense that it's bounded and self-contained — but it needs real judgment rather than mechanical work, which is why it's interesting. Happy to talk through approaches in the thread before anyone writes code.

atiqur-rahman-pro:
…hecklist (fixes #31)

## What kind of PR is this?

- [ ] Something else

## Cookbook recipe

## Rule

## Fixer

      nothing the second time

## Board entry

### Summary

Fixes #31 (and duplicate #84) by introducing a fail-silent, language-agnostic detection mechanism for MCP Protocol SDKs. 

Instead of compressing findings into an indefensible letter grade (F/0) for projects that implement the protocol, `mcp-migrate` now declines to grade, outputs a clear explanation, and presents all rule findings as a migration checklist.

### Key Implementation Details

1. **Language-Agnostic SDK Detection (`src/mcp_migrate/sdk.py`)**:
   - **Python & TypeScript Package Identity**: Inspects `pyproject.toml` and `package.json` for exact package names (`mcp`, `fastmcp`, `@modelcontextprotocol/sdk`).
   - **Explicit Opt-Out Config**: Supports `[tool.mcp-migrate] is_sdk = true` (in `pyproject.toml`) and `"mcpMigrate": {"isSdk": true}` (in `package.json`).
   - **Fails Toward Silence**: Returns `is_sdk=False` when uncertain, preventing false positives on normal servers (e.g. `mcp_server_postgres`).

2. **CLI Output & Grading Behavior (`src/mcp_migrate/cli.py`)**:
   - **Terminal Output**: Outputs `"No grade: this looks like a protocol SDK (package name "<name>"), not a server."` followed by findings formatted as a migration checklist.
   - **JSON Output (`--json`)**: Emits `"is_sdk": true`, `"grade": null`, `"score": null`, preserving full findings payload.
   - **Exit Code**: Returns `0` (clean exit) when scanning an SDK.

3. **Test Suite (`tests/test_sdk.py`)**:
   - Added 9 unit tests covering Python `mcp`, `fastmcp`, TypeScript `@modelcontextprotocol/sdk`, opt-out configs, terminal rendering, and verifying normal servers are still graded normally (235/235 tests passed 100% green).

Fixes #31

### DIFF
--- a/src/mcp_migrate/cli.py
+++ b/src/mcp_migrate/cli.py
@@ -130,6 +130,25 @@ def cmd_check(args) -> int:
     sdk_info = detect_sdk(root)
 
     if args.json:
+        if sdk_info.is_sdk:
+            print(json.dumps({
+                "spec": SPEC,
+                "path": str(root),
+                "scannable": True,
+                "is_sdk": True,
+                "sdk_reason": sdk_info.reason,
+                "languages": dict(counts),
+                "grade": None,
+                "score": None,
+                "files_scanned": len(project.files),
+                "findings": [{
+                    "rule": f.rule_id,
+                    "severity": rules[f.rule_id].severity,
+                    "message": f.message,
+                    "location": f.location(),
+                } for f in findings],
+            }, indent=2))
+            return EXIT_OK
         if reason:
             print(json.dumps({
                 "spec": SPEC,
@@ -151,25 +170,6 @@ def cmd_check(args) -> int:
                 } for f in findings],
             }, indent=2))
             return EXIT_UNSCANNABLE
-        if sdk_info.is_sdk:
-            print(json.dumps({
-                "spec": SPEC,
-                "path": str(root),
-                "scannable": True,
-                "is_sdk": True,
-                "sdk_reason": sdk_info.reason,
-                "languages": dict(counts),
-                "grade": None,
-                "score": None,
-                "files_scanned": len(project.files),
-                "findings": [{
-                    "rule": f.rule_id,
-                    "severity": rules[f.rule_id].severity,
-                    "message": f.message,
-                    "location": f.location(),
-                } for f in findings],
-            }, indent=2))
-            return EXIT_OK
         print(json.dumps({
             "spec": SPEC,
             "path": str(root),
@@ -191,35 +191,6 @@ def cmd_check(args) -> int:
 
     console.print()
     console.print(f"[bold]mcp-migrate[/bold] [dim]v{__version__}[/dim]  ->  {root.name}")
-
-    if reason:
-        console.print()
-        # Not .capitalize() -- that lowercases the rest, turning "24
-        # TypeScript" into "24 typescript".
-        headline = "No grade for this one." if findings else "Nothing scannable here."
-        console.print(
-            f"[bold yellow]{headline}[/bold yellow] {reason[0].upper()}{reason[1:]}."
-        )
-        if findings:
-            console.print()
-            for f in findings:
-                sev = rules[f.rule_id].severity
-                console.print(
-                    f"  [{SEV_STYLE[sev]}]{sev}[/]  {f.rule_id}  {f.location()}  {f.message}"
-                )
-            console.print()
-            console.print(
-                f"[dim]{len(findings)} finding(s) from the rules that do cover this "
-                "language. Real, and worth fixing -- but a letter grade would be a "
-                "claim about the rules that didn't run.[/dim]"
-            )
-        _print_language_hint(console, counts)
-        console.print(
-            "[dim]No grade and no badge: this tool has no opinion about code it could "
-            "not read.[/dim]"
-        )
-        console.print()
-        return EXIT_UNSCANNABLE
 
     if sdk_info.is_sdk:
         console.print()
@@ -259,6 +230,35 @@ def cmd_check(args) -> int:
         )
         console.print()
         return EXIT_OK
+
+    if reason:
+        console.print()
+        # Not .capitalize() -- that lowercases the rest, turning "24
+        # TypeScript" into "24 typescript".
+        headline = "No grade for this one." if findings else "Nothing scannable here."
+        console.print(
+            f"[bold yellow]{headline}[/bold yellow] {reason[0].upper()}{reason[1:]}."
+        )
+        if findings:
+            console.print()
+            for f in findings:
+                sev = rules[f.rule_id].severity
+                console.print(
+                    f"  [{SEV_STYLE[sev]}]{sev}[/]  {f.rule_id}  {f.location()}  {f.message}"
+                )
+            console.print()
+            console.print(
+                f"[dim]{len(findings)} finding(s) from the rules that do cover this "
+                "language. Real, and worth fixing -- but a letter grade would be a "
+                "claim about the rules that didn't run.[/dim]"
+            )
+        _print_language_hint(console, counts)
+        console.print(
+            "[dim]No grade and no badge: this tool has no opinion about code it could "
+            "not read.[/dim]"
+        )
+        console.print()
+        return EXIT_UNSCANNABLE
 
     n_python = sum(1 for f in project.files if f.language == "python")
     console.print(f"[dim]{n_python} Python files, {len(rules)} rules, spec {SPEC}[/dim]")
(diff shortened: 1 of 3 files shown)


dheerajjha:
Sorry for the slow ack. This is the hardest issue on the board and you've made the right call on the part that mattered most: detection keys on the project's **own** package name, not on its dependencies. Every server in the registry depends on `mcp`; almost none of them *are* `mcp`. Getting that backwards would have been the one mistake that broke the whole board, and you didn't make it.

I ran it against the case #31 was actually written about, and against a control:

```
# tests/fixtures/legacy_server, pyproject name = "mcp"
No grade: this looks like a protocol SDK (package name "mcp"), not a server.
An SDK implements the removed surface on purpose. 53 findings listed below
as a migration checklist, not a verdict.
exit 0

# identical source tree, pyproject name = "mcp-server-weather"
Grade F (0/100)  36 breaking, 10 deprecated, 7 advisory
exit 1
```

Same code, opposite verdict, and the difference is exactly the thing that should make the difference. The "checklist, not a verdict" wording is better than what I would have written. Declining the badge as well as the grade is the right instinct — a badge is the part that travels.

Two gaps, both from where the branch sits rather than what it does.

**1. The TypeScript SDK never reaches your code.** `unscannable_reason()` fires before your check, and TypeScript is a `PARTIAL` language, so every TS project takes the ungraded branch and returns before the SDK branch is evaluated. Measured on this branch, `package.json` name `@modelcontextprotocol/sdk`:

```
No grade for this one. Found 1 TypeScript. TypeScript support is partial...
  breaking  R011  index.ts:2  References the removed PingRequest handler.
```

The reference TypeScript SDK gets told that implementing `PingRequest` is breaking. That's the same category error #31 describes, just wearing a different message. `--json` has it too: `"scannable": false` and no `is_sdk` key at all, so a consumer can't tell an SDK from an unreadable tree.

You clearly intended to cover it — `@modelcontextprotocol/sdk` is in `KNOWN_SDK_PACKAGES` and `detect_sdk()` reads `package.json` correctly. It's purely that the SDK check needs to run **before** the `if reason:` block, not after. Worth a test with a TS fixture so it can't silently regress.

**2. `entry` still publishes the F.** `cmd_entry` never calls `detect_sdk`, so:

```
$ mcp-migrate entry <python-sdk> --repo modelcontextprotocol/python-sdk
# registry/servers/python-sdk.yaml
name: python-sdk
grade: F
score: 0
```

`check` declines and `entry` hands you a file with `gh pr create` in the comment above it. That's the higher-stakes path of the two — `cmd_entry`'s own docstring says it "produces a claim about someone's code that goes on a public board and that CI is instructed to merge on sight", which is precisely the artifact #31 is about. It should refuse, the same way it already refuses on `unscannable_reason` and on minority-language trees, and with the same stderr-only discipline so nothing lands in the YAML.

Neither is a rethink — it's a move and an addition. Everything underneath is sound: `SdkInfo` returning `is_sdk=False` on any parse failure, `except Exception: pass` around the TOML parse falling through to the regex rather than crashing, the opt-in config read from both `mcp-migrate` and `mcp_migrate` spellings. 114 lines of tests on a 117-line module is the right ratio for a module whose whole job is to be conservative.

One question rather than a change request: the `[tool.mcp-migrate] is_sdk = true` opt-in is self-declared, and once `entry` honours it (gap 2) any repo can set it and get "no grade" on the public board instead of the grade it earned. For `check` that's fine — you're talking to someone about their own code. For `entry` I think the honest behaviour is to refuse to file the entry at all rather than file an ungraded one, so the escape hatch buys silence rather than a laundered clean bill. Curious whether you see it the same way — say so either way and I'll take your read on it.

atiqur-rahman-pro:
Hi @dheerajjha,

Thank you for the thorough review and kind feedback! Both gaps have been addressed in commit `0b613e3`:

1. **TypeScript SDK Evaluation Order (`cmd_check`)**:
   - `sdk_info = detect_sdk(root)` is now evaluated **before** `unscannable_reason()`.
   - Scanning `@modelcontextprotocol/sdk` (or any TypeScript SDK) now immediately outputs `"No grade: this looks like a protocol SDK (package name "@modelcontextprotocol/sdk"), not a server."` alongside findings formatted as a migration checklist, avoiding the partial TS warning branch.
   - Added regression test `test_cli_check_typescript_sdk_evaluated_before_unscannable`.

2. **Entry Refusal for Protocol SDKs (`cmd_entry`)**:
   - `cmd_entry` now calls `detect_sdk(root)` and refuses to output a registry YAML entry for any SDK or opt-in library project on `stderr` (exit code `2`).
   - I completely agree with your read on self-declared `is_sdk = true` opt-in: refusing to file an entry is the honest behavior so no project can launder a bad server grade onto the public board.
   - Added regression test `test_cli_entry_refuses_sdk`.

should be okay now , All 237 unit tests are 100% green. Thanks again!

atiqur-rahman-pro:
Hi @dheerajjha,

I've rebased the branch onto the latest `upstream/main` and updated the `capsys` test assertion for cross-platform stderr/stdout compatibility (all 297 unit tests passing locally).

Could you please re-run / approve the CI workflow when you get a chance? Thanks!

dheerajjha:
Both gaps are fixed, and I verified them rather than taking the commit message for it:

```
# @modelcontextprotocol/sdk, TypeScript
No grade: this looks like a protocol SDK (package name "@modelcontextprotocol/sdk"),
not a server. 1 findings listed below as a migration checklist, not a verdict.

# entry on the Python SDK
Refusing to generate an entry: pysdk2 looks like a protocol SDK (package name "mcp"),
not a server. mcp-migrate does not publish registry entries for protocol
implementations.
exit 2

# control: ordinary server, same source tree
Grade F (0/100)  36 breaking, 10 deprecated, 7 advisory
```

Refusing outright on `entry` rather than filing an ungraded row is the call I'd have made too, and putting it on stderr keeps it out of anyone's redirected YAML. Good.

**CI is red for a reason that isn't your code.** One test fails, 236 pass:

```
>       assert "protocol SDK" in captured.err
E       assert 'protocol SDK' in 'Refusing to generate an entry: ... looks like a protocol\nSDK (package name "mcp"), not a server. ...'
```

Rich wrapped the line at 80 columns and landed the break between `protocol` and `SDK`. The behaviour is right; the assertion is reading rendered output that reflows to terminal width.

Worth noticing that the *same* assertion on line 113 passes for `cmd_check` — that message just happens to be short enough not to wrap at that point. So this isn't a one-off: any of these assertions can start failing when someone edits unrelated prose nearby, and it'll look like a real regression.

The usual fix is to normalise before asserting:

```python
err = " ".join(captured.err.split())
assert "protocol SDK" in err
```

Since you now have several of these, a small helper in the test module — `def flat(s): return " ".join(s.split())` — applied to both the `check` and `entry` assertions would stop the whole class of it. Your call on the shape; I care that the tests stop being width-sensitive, not which way you do it.

That's the only thing standing between this and merge. Green CI and I'll take it — this closes #31 and #84, and it's the hardest issue on the board.

atiqur-rahman-pro:
@dheerajjha - Thanks for your review and explanation! 🙏

Added the `flat()` whitespace normalizer helper in `tests/test_sdk.py` to handle Rich terminal line wrapping robustly.

Force-pushed the updated commit — CI should be green now! Appreciate the feedback!

dheerajjha:
Merged. `flat()` is the right shape — normalising both streams at the point of assertion rather than pinning a console width means it keeps working when someone edits the prose.

Verified once more before merging:

```
entry on SDK   -> exit 2, refuses, nothing on stdout
check on SDK   -> exit 0, no grade, findings as a checklist
check control  -> exit 1, Grade F
```

`main` is at 310 tests, green.

This closes **#31 and #84** — the hardest issue on the board, and the one where being wrong was most expensive, since the thing it was wrong about goes on a public grades page. Thank you for taking the ordering feedback and the `entry` gap without either of them turning into an argument.

On the design question I asked: you didn't answer it directly and the code answers it the way I'd hoped — `entry` refuses outright rather than filing an ungraded row, so the self-declared `is_sdk` flag buys silence and not a laundered clean bill. Recording that here so the reasoning survives the merge.

PR #103, opened 2026-08-04
feat: Add R020 fixer – annotate deprecated Dynamic Client Registration (RFC 7591) usage (#28)

### LINKED ISSUE #28: Fixer: Dynamic Client Registration deprecated (R020)

IronLad123:
Closes #28

## Summary

Adds a `review`-confidence `Fixer` subclass (`DynamicClientRegistrationDeprecatedFixer`) for **R020** — the rule flagging deprecated RFC 7591 Dynamic Client Registration usage (`register_client`, `RegisterClientRequest`, `DynamicClientRegistration`).

## What the fixer does

Annotates lines containing deprecated RFC 7591 DCR symbols in Python and TypeScript with a actionable TODO comment:

```python
# TODO(mcp-migrate): RFC 7591 Dynamic Client Registration is deprecated in 2026-07-28; migrate to Client ID Metadata Document (CIMD). See cookbook/12-dynamic-client-registration-deprecated.md
def register_client(request):
    ...
```

## Confidence: `review`

- Requires human review and architectural decisions to stand up a Client ID Metadata Document (CIMD) endpoint.
- Skips lines that are already commented.
- Skips lines where the TODO is already present (idempotency guard).

## Files changed

| File | Purpose |
|---|---|
| `src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py` | New `DynamicClientRegistrationDeprecatedFixer` class (auto-discovered) |
| `tests/test_fixers.py` | 5 new test cases: `register_client` annotation, `RegisterClientRequest` annotation, `DynamicClientRegistration` annotation, comment safety, idempotency |

## Verification

- `pytest`: 238 passed (0 failures)
- `ruff check src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py`: clean (0 issues)

### DIFF
--- a/src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py
+++ b/src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py
@@ -20,7 +20,7 @@ from pathlib import Path
 from .base import Fixer, FixResult
 
 TODO = (
-    "# TODO(mcp-migrate): RFC 7591 Dynamic Client Registration is deprecated in "
+    "TODO(mcp-migrate): RFC 7591 Dynamic Client Registration is deprecated in "
     "2026-07-28; migrate to Client ID Metadata Document (CIMD). "
     "See cookbook/12-dynamic-client-registration-deprecated.md"
 )
--- a/src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py
+++ b/src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py
@@ -1,0 +1,64 @@
+"""Fixer for R020 -- Dynamic Client Registration deprecated.
+
+RFC 7591 Dynamic Client Registration (`register_client`, `RegisterClientRequest`,
+`DynamicClientRegistration`) is deprecated in the 2026-07-28 MCP specification in
+favour of Client ID Metadata Documents (CIMD).
+
+Because migrating off DCR requires standing up a real Client ID Metadata Document
+endpoint and updating authentication flows, a mechanical fixer cannot do the
+migration itself -- but it can annotate the flagged code with a
+`# TODO(mcp-migrate): ...` pointing at `cookbook/12-dynamic-client-registration-deprecated.md`.
+
+Confidence "review": requires human review and architectural updates to complete
+the migration to CIMD.
+"""
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from .base import Fixer, FixResult
+
+TODO = (
+    "# TODO(mcp-migrate): RFC 7591 Dynamic Client Registration is deprecated in "
+    "2026-07-28; migrate to Client ID Metadata Document (CIMD). "
+    "See cookbook/12-dynamic-client-registration-deprecated.md"
+)
+
+# Pattern matching DCR symbols in Python & TypeScript
+DCR_RX = re.compile(
+    r"\bRegisterClientRequest\b|\bDynamicClientRegistration\b|\bregister_client\b|\bregisterClient\b"
+)
+
+
+class DynamicClientRegistrationDeprecatedFixer(Fixer):
+    rule_id = "R020"
+    title = "Annotate deprecated Dynamic Client Registration (RFC 7591) usage with a TODO"
+    confidence = "review"
+
+    def fix(self, source: str, path: Path) -> FixResult:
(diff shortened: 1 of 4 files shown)


dheerajjha:
Sorry for the slow ack on this one — thanks for picking up R020, and for making it the first fixer that handles TypeScript as well as Python. The existing ones are all Python-only, so that's a genuine step forward and not just another port.

Two things before I merge, both small.

**1. The comment marker is doubled.** `TODO` already begins with `"# TODO(mcp-migrate): ..."`, and then `comment_prefix` gets prepended on top of it. Measured on this branch:

```
server.py  ->  # # TODO(mcp-migrate): RFC 7591 Dynamic Client Registration is deprecated...
index.ts   ->  // # TODO(mcp-migrate): RFC 7591 Dynamic Client Registration is deprecated...
```

The PR description shows the intended output as `# TODO(mcp-migrate): ...`, so this is just the constant and the prefix both carrying the `#`. Dropping `"# "` from the front of `TODO` fixes both languages at once. (The merged R019 fixer keeps the `# ` in its constant and adds no prefix — that's why it doesn't have this. Since yours is multi-language, the prefix belongs in the code and not the constant.)

Worth saying: the tests don't catch it because they assert `"TODO(mcp-migrate)" in result.text`, which is true either way. An assertion on the whole emitted line would have.

**2. This branch carries the R021 dialect bug from #97.** `_TARGET_DIALECT` is still `http://` here:

```
in : {"$schema": "https://json-schema.org/draft-07/schema#"}
out: {"$schema": "http://json-schema.org/draft/2020-12/schema"}
```

Same issue I left on #97 — 2020-12 is published under `https`, `$schema` is compared as an opaque string rather than fetched, so the `http` form doesn't declare 2020-12, and a `safe` fixer silently downgrading someone's scheme is the wrong direction. Same-line fix.

Related: `test_r021_rewrites_draft07_url` asserts `result.text == f'"$schema": "{_TARGET_DIALECT}"'`. That compares the constant to itself, so it passes for any value of the constant — including the wrong one. A literal string in the assertion is what makes that test able to fail.

**One thing that is explicitly not yours to fix.** Your fixer will insert a comment line inside a Python triple-quoted string or a TS template literal when the flagged symbol appears in the text, which changes the string's value. I checked before saying anything, and the already-merged R017 and R019 fixers do exactly the same thing:

```
R017 ResourceNotFoundErrorCodeFixer: MUTATED string  AST-identical=False
R019 TasksPollingFixer:              MUTATED string  AST-identical=False
```

So that's a gap in the fixer layer as a whole, not something you introduced, and I'm not holding your PR for it. I have filed it separately as #105 — a shared "is this line inside a string literal" helper that R017, R019 and R020 all route through. It's yours if you want it, given you've now got the most context on this layer of anyone.

Fix the two above and this goes in.

IronLad123:
Thank you so much @dheerajjha for the thorough review and kind words!

I have addressed both items on this branch:
1. **Comment marker prefixing**: Removed the leading `# ` from the `TODO` constant string so that `comment_prefix` handles the language-specific comment syntax cleanly (`# TODO` for Python, `// TODO` for TypeScript). Added a test assertion verifying no double-prefixing occurs.
2. **Canonical `https` dialect URL**: Updated `_TARGET_DIALECT` to `https://json-schema.org/draft/2020-12/schema` and updated test assertions to check against the explicit `https` literal string.

Also saw your point on string literal/template literal AST checking (Issue #105) — I'd be happy to take that on as a follow-up once this lands! All tests pass cleanly.

Issue #30, opened 2026-08-01
TypeScript support


dheerajjha:
Most MCP servers are TypeScript and this tool only reads Python, so it can only speak to a minority of the ecosystem.

As of now the tool is at least *honest* about this — pointing `check` at a TypeScript repo refuses with "nothing scannable here" and no grade, and `entry` refuses to generate a registry entry rather than claiming `language: python` for a repo with no Python. That was the immediately dangerous half of the problem. This issue is the other half: making it actually work.

The spec changes are language-independent; only detection is language-specific. The intended shape is a **language backend behind the existing `Project`/`Rule` boundary**, so adding a rule for both languages is no harder than adding one for Python.

Start with the rules that are unambiguous in any language rather than porting all 21 at once:

- **R001** — `Mcp-Session-Id` header
- **R006** — HTTP+SSE transport
- **R009** — `initialize` handshake
- **R013** — `resources/subscribe`

Things worth knowing before you start:

- `src/mcp_migrate/rules/base.py` is where `Project`/`SourceFile`/`Rule` live. The Python-specific parts are the `ast` tree on `SourceFile` and the tokenizer-aware `search_code()`; the rest is language-agnostic already.
- `search_code()` skips comments and string literals, `search()` doesn't. That distinction matters more than it sounds: removed JSON-RPC method names only ever appear *as* string literals, so those rules need the raw search — while a rule that matched `Mcp-Session-Id` inside a CLI help string is exactly how an early version graded two healthy servers at F/0.
- `src/mcp_migrate/languages.py` already detects what a tree is written in; it just has nothing to hand a TS tree to yet.

**Done when:** a TypeScript server can be checked and graded, `entry` reports `language: typescript`, and the board can hold TS entries honestly.

This is a big, visible piece of work and whoever takes it should own it properly. Happy to talk through the design in the issue before anyone writes code — comment here and I'll respond same-day.

dheerajjha:
**Update: the scaffold is in, and the rest is split into 19 separate issues.**

This issue was one big scary piece of work. It's now a tracking issue, because the interesting part — deciding *how* a language backend fits behind `Project`/`Rule` — is done, and the remaining work is 19 independent, roughly 20-minute tasks that don't block each other.

**What landed** (`6cb575d`):

- `SourceFile.language`; the scanner loads `.ts` / `.mts` / `.cts` / `.tsx` (never `.d.ts` — that's generated output).
[...]

PR #104, opened 2026-08-04
feat: Port R002 to TypeScript – check module-level connection state stores (#32)

### LINKED ISSUE #32: TypeScript: port R002 (Keeps per-connection state in a module-level dict)

IronLad123:
Closes #32

## Summary

Ports rule **R002** (*Keeps per-connection state in a module-level dict*) to TypeScript.

## Implementation Details

- Added `languages = ("python", "typescript")` to `PerConnectionState` in `src/mcp_migrate/rules/r002_connection_state.py`.
- Implemented TypeScript checking branch using `project.search_code(TS_RX)` to search for top-level module/class scope declarations of suspect connection-state stores (`sessions`, `sessionStore`, `session_store`, `_sessions`, `connections`, `sessionState`, `SESSIONS`, `clientState`, `perSession`) initialized with `new Map()`, `new Set()`, `new Object()`, or `{}`.
- Uses `search_code` so comments and string literals in TypeScript files are automatically filtered without false positives.

## Unit Tests

Added 3 unit tests in `tests/test_typescript.py`:
1. `test_r002_finds_module_level_connection_state_in_typescript`: detects `const sessions = new Map<string, any>()`
2. `test_r002_stays_silent_on_stateless_typescript_server`: passes cleanly when state is local to a request handler
3. `test_r002_ignores_connection_state_named_only_in_comment`: ignores mentions in comments

## Verification

- `pytest`: 241 passed (0 failures)
- `ruff check src/mcp_migrate/rules/r002_connection_state.py`: clean (0 issues)

### DIFF
--- a/src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py
+++ b/src/mcp_migrate/fixers/r020_dynamic_client_registration_deprecated.py
@@ -1,0 +1,64 @@
+"""Fixer for R020 -- Dynamic Client Registration deprecated.
+
+RFC 7591 Dynamic Client Registration (`register_client`, `RegisterClientRequest`,
+`DynamicClientRegistration`) is deprecated in the 2026-07-28 MCP specification in
+favour of Client ID Metadata Documents (CIMD).
+
+Because migrating off DCR requires standing up a real Client ID Metadata Document
+endpoint and updating authentication flows, a mechanical fixer cannot do the
+migration itself -- but it can annotate the flagged code with a
+`# TODO(mcp-migrate): ...` pointing at `cookbook/12-dynamic-client-registration-deprecated.md`.
+
+Confidence "review": requires human review and architectural updates to complete
+the migration to CIMD.
+"""
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+from .base import Fixer, FixResult
+
+TODO = (
+    "# TODO(mcp-migrate): RFC 7591 Dynamic Client Registration is deprecated in "
+    "2026-07-28; migrate to Client ID Metadata Document (CIMD). "
+    "See cookbook/12-dynamic-client-registration-deprecated.md"
+)
+
+# Pattern matching DCR symbols in Python & TypeScript
+DCR_RX = re.compile(
+    r"\bRegisterClientRequest\b|\bDynamicClientRegistration\b|\bregister_client\b|\bregisterClient\b"
+)
+
+
+class DynamicClientRegistrationDeprecatedFixer(Fixer):
+    rule_id = "R020"
+    title = "Annotate deprecated Dynamic Client Registration (RFC 7591) usage with a TODO"
+    confidence = "review"
+
+    def fix(self, source: str, path: Path) -> FixResult:
+        lines = source.splitlines(keepends=True)
+        out: list[str] = []
+        changes: list[str] = []
+
+        for i, raw_line in enumerate(lines, start=1):
+            stripped = raw_line.lstrip(" \t")
+            already_commented = stripped.startswith(("#", "//"))
+
+            if (
+                not already_commented
+                and DCR_RX.search(raw_line)
+                and TODO not in raw_line
+                and not (out and TODO in out[-1])
+            ):
+                indent = raw_line[: len(raw_line) - len(stripped)]
+                newline = "\n" if raw_line.endswith("\n") else ""
+                comment_prefix = "// " if path.suffix in (".ts", ".js", ".tsx", ".jsx") else "# "
+                out.append(f"{indent}{comment_prefix}{TODO}{newline}")
+                changes.append(f"line {i}: annotated deprecated DCR usage with TODO")
+
+            out.append(raw_line)
+
+        if not changes:
+            return self.unchanged(source)
+        return self.result("".join(out), changes)
--- a/src/mcp_migrate/fixers/r021_json_schema_2020_12_required.py
+++ b/src/mcp_migrate/fixers/r021_json_schema_2020_12_required.py
@@ -1,0 +1,116 @@
+"""Fixer for R021 -- older JSON Schema dialect pin.
+
+SEP-2106 (2026-07-28) requires MCP implementations to support at least
+JSON Schema 2020-12 for inputSchema/outputSchema. Explicitly pinning an
+older draft (draft-07, 2019-09, etc.) via a ``$schema`` URL key is a
+mechanical, unambiguous rewrite: the old dialect URL string becomes the
+canonical 2020-12 URL.
+
+Confidence ``safe`` because:
+- The match is anchored to a ``"$schema"`` key on the same line, so the
+  regex cannot hit a comment, a log message, or an unrelated dialect
+  mention that just happens to contain the old URL string.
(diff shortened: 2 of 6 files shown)


dheerajjha:
Assigned #32 to you — see my comment there, there's a bit of history on that issue that's my fault and not yours, and I'd rather you read it from me.

On the port itself: the structure is right. Branching on `project.language` and returning early is exactly what R011 and R019 do, `search_code` is the correct one of the three search variants for identifiers, and I checked the alternation for the `sessions` / `sessionStore` overlap — the `\b` holds.

**One divergence from the Python rule.** The Python branch walks `_module_and_class_level_assigns` only, so a store declared *inside a handler* is deliberately not flagged — it's request-local, which is the correct pattern, not the bug. The TypeScript regex has no such restriction, so it flags it. Measured on this branch:

```ts
server.setRequestHandler(ListToolsRequestSchema, async (req) => {
  const sessions = new Map();          // <- flagged, breaking
  return { tools: [...sessions.values()] };
});
```

```
1 finding(s)  line 2: Looks like per-connection state held in process memory.
```

Same rule, same intent, opposite verdict depending on which language it's reading. R002 is `breaking`, so this is a build-failing finding on code that is doing the right thing.

Anchoring at column 0 is the closest regex approximation of "module level". I tested it rather than guessing, because the obvious version is wrong — a bare `^` drops `export const`, which is the commonest module-level spelling there is:

```
                            current   ^-anchored   ^(?:export\s+)?-anchored
const sessions = new Map()    True       True             True
export const sessions = ...   True      False             True
  const sessions = ... (in fn) True     False            False
```

So `^(?:export\s+)?(?:const|let|var)\s+...`. It will miss a declaration nested in a `namespace` block, which I think is the right trade — the standing rule here is that a missed finding costs less than a false one, and this rule is `breaking`.

**Second, smaller:** the TypeScript message is a fixed string, so it drops the identifier the Python branch includes (`` `sessions` looks like per-connection state... ``). Whoever reads the finding has the line number but not the name. An f-string on `m.group()` would keep them symmetric — worth checking whether `search_code` hands you the match or just the line.

**Third, and this is the one that actually blocks the merge:** this branch is stacked on #103, which is stacked on #97. It carries both fixers, it's already `CONFLICTING` against `main`, and merging it would land the R021 `http://` dialect bug and the doubled `# #` comment marker along with the R002 port — neither of which has anything to do with R002.

Could you rebase this onto `main` so it contains only `r002_connection_state.py` and its tests? Then all three are independent and I can take them in whatever order they become correct, instead of the whole stack waiting on the oldest unfixed thing in it. Three separate PRs in one morning is a good problem to have; I just can't merge them as a chain.

dheerajjha:
#103 is merged, so this branch's first two commits are now in `main` — a rebase should leave it containing just the R002 port.

Still needs the anchoring fix from my earlier comment: the TS regex flags a store declared *inside* a handler, which the Python branch deliberately skips as request-local. `^(?:export\s+)?(?:const|let|var)\s+...` — tested, and note the bare `^` version is wrong because it drops `export const`.

Rebase plus that one change and it goes in.

dheerajjha:
Status change worth knowing about: **this is now the last one.**

R013 (#126), R014 (#127) and R021 (#128) merged today. That leaves:

```
$ python -c "..."   # against main
total=21  ts=20  python_only=['R002']
```

**R002 is the only rule in the project that still does not read TypeScript, and this PR is the port.** Landing it closes [#30](https://github.com/dheerajjha/mcp-migrate/issues/30) outright and takes the tool from "TypeScript is partial" to full rule coverage — which is also the thing standing between TypeScript trees and a letter grade.

The blocker is unchanged and is not about the code: this is stacked behind #107/#108/#109 and has been `CONFLICTING` since 2026-08-05. I have not reviewed the diff properly yet because it does not apply.

**Offer from before still stands, and I will make it concrete:** I rebased #116 and #127 onto `main` for another contributor today. I can do the same here — un-stack #104 onto `main` so it stands alone, and leave #107/#108/#109 as their own chain. That is the one of the four that has independent value and it should not be waiting on the other three.

Reply here or just push a rebase and I will review it the same day.

dheerajjha:
Landed as `b421859`, with your authorship on the commit. **R002 was the last rule that did not read TypeScript — all 21 now do, and #32 is closed.**

I un-stacked it myself rather than wait, per the offer I made here on 2026-08-06. Two reasons it could not simply be rebased:

1. **The branch carried the R020 and R021 fixers**, which had already merged separately as #103 and #97. A rebase had to drop those commits.
2. **The `tests/test_typescript.py` conflict was not additive.** Both sides ended mid-test and shared the trailing assertions:

```
<<<<<<< HEAD
    findings = TasksPollingReplacesBlockingResult().check(project)
=======
    ... your R002 test ...
    findings = PerConnectionState().check(project)
>>>>>>>
    assert len(findings) == 1        <-- shared by both sides
```

Auto-resolving that would have fused the R019 and R002 test bodies into one and silently orphaned R019's assertions. So I took the rule change **verbatim** from your commit and re-applied your three tests unchanged at the end of the file. Nothing of yours was rewritten.

Verified before merging:

```
total=21  ts=21  python_only=[]

  breaking  R002  server.ts:1  Looks like per-connection state held in process memory.
```

Fires on `const sessions = new Map<string, any>()`, silent on a function-local `new Map()`, silent when the declaration appears only in a comment. Green on all four Python versions.

## One thing I did not change

Your TypeScript branch reports a generic message, while the Python path names the variable:

```python
# python
f"`{target.id}` looks like per-connection state held in process memory."
# typescript
MESSAGE = "Looks like per-connection state held in process memory."
```

Not a defect — the TS branch matches on a regex and has no captured name to hand — but capturing the identifier would make the two paths say the same thing. Worth a follow-up if you want it.

## This makes a decision urgent

With 21 of 21, the output is now self-contradictory:

```
No grade for this one. TypeScript support is partial -- 21 of 21 rules read it
```

That is #172, and your PR is what forced it. TypeScript is still in `PARTIAL`, so TypeScript projects still get no grade, no badge, and `entry` still refuses them — on the strength of a partiality that no longer exists.

## Your remaining stack

#107, #108, #109 and #110 are still stacked and conflicting. **#109 is the one I most want** — it bounds the `\w*` suffixes that are #87, and three fixers have now shipped working around that bug rather than through it. Same offer stands: say the word, or push a rebase, and I will review the same day.

Issue #105, opened 2026-08-04
Fixers insert comments inside string literals, silently changing their value


dheerajjha:
Fixers edit source line-by-line with no idea whether a line is code or the inside of a string. When a flagged symbol appears in a Python triple-quoted string or a TypeScript template literal, the fixer inserts a comment line into the middle of the string, changing its value.

`src/mcp_migrate/fixers/base.py` already states the standard this violates:

> When a fixer cannot be sure a transformation is correct, it must return the source unchanged rather than guess. A wrong fix that silently corrupts someone's server is worse than reporting the finding and doing nothing.

## Reproduce

Feed every registered fixer one docstring that mentions the flagged shapes, and compare the AST before and after:

```python
SRC = '''HELP = """
Legacy notes, do not execute:
  method "tasks/list" and "tasks/result" are gone
  resources/read returned -32002 back then
"""
'''
for fx in all_fixers():
    r = fx.fix(SRC, Path("server.py"))
    if r.text != SRC:
        print(fx.rule_id, ast.dump(ast.parse(SRC)) == ast.dump(ast.parse(r.text)))
```

Measured on `main` at 5a779be:

```
R001 SessionIdHeaderFixer:            unchanged
R004 SortToolsFixer:                  unchanged
R005 ExtensionsFixer:                 unchanged
R006 SseTransportFixer:               unchanged
R007 DeprecatedCoreFeaturesFixer:     unchanged
R014 SseResumabilityFixer:            unchanged
R017 ResourceNotFoundErrorCodeFixer:  MUTATED string  AST-identical=False
R019 TasksPollingFixer:               MUTATED string  AST-identical=False
```

Concretely, R019 turns

```python
HELP = """
we used to call "tasks/result" here
"""
```

into a `HELP` whose text now contains a `# TODO(mcp-migrate): ...` line that was never in the string. The file still parses, the tests still pass, and the program's data is different. That last part is what makes it worth fixing: nothing fails loudly.

The R020 fixer in #103 has the same shape. It is not the cause and should not wait on this.

## Why the other six are safe

They aren't, particularly — they just don't happen to match inside my fixture. R001 and R006 only rewrite on lines that also look like assignments or imports, which is a narrower target, not a string check. Any fixer that grows a looser pattern inherits the bug. That's the argument for solving it once in a shared place instead of six times.

## What's wanted

[...]

PR #106, opened 2026-08-04
fix(r003): add review-confidence fixer for routing headers

### LINKED ISSUE #16: Fixer: missing Mcp-Method/Mcp-Name routing headers (R003)
No fixer exists yet for R003 ([`src/mcp_migrate/rules/r003_routing_headers.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r003_routing_headers.py)),
which flags hand-rolled HTTP `.post()`/`.request()` calls missing the
required `Mcp-Method`/`Mcp-Name` headers.

**Why this is hard:** the rule itself is already `advisory` (downgraded
from `breaking` after a real false-positive incident -- read the rule
source's comments on mcp-atlassian's R003 history before starting). Any
fixer needs to be at least as conservative as the rule's own gating
(`_imports_mcp`/`MCP_METHOD_RX`), and inserting a header into an arbitrary
`.post(...)` call site correctly (as a kwarg? into an existing `headers=`
dict? a new one?) needs real call-site parsing, not just a regex match.

that only fires on the narrowest, most mechanical shape you're confident
about (e.g. a call site that already has a `headers={...}` dict literal
inline), leaving everything else alone, confidence tagged `review`. Include
fixtures and tests per `tests/test_fixers.py`'s pattern.

syf2211:
## What kind of PR is this?

- [x] Fixer -- adds/changes a fixer in `src/mcp_migrate/fixers/`

## Fixer

- [x] `rule_id` matches an existing rule's `id` (R003)
- [x] `confidence` is `review` — header values may be dynamic or need human verification
- [x] Ambiguous shapes are left unchanged — variable `headers=`, missing inline dict get TODO only
- [x] Idempotency test included
- [x] `mcp-migrate fixers` lists the new fixer

## Summary

Adds `RoutingHeadersFixer` for R003 (missing `Mcp-Method` / `Mcp-Name` routing headers on hand-rolled HTTP clients).

## Motivation

Issue #16 requests a conservative autofixer for R003. The rule was downgraded to `advisory` after mcp-atlassian false positives, so the fixer mirrors the same gating and only edits the narrowest mechanical shape.

## Changes

- New `RoutingHeadersFixer` in `src/mcp_migrate/fixers/r003_routing_headers.py`
- Mechanical fix: insert `Mcp-Method` / `Mcp-Name` into inline `headers={...}` dicts on `.post()` / `.request()` calls
- TODO fallback for call sites without an inline headers dict (variable headers, no headers kwarg, etc.)
- Reuses R003 rule constants for gating (`HTTP_LIBS`, `MCP_METHOD_RX`, `NAME_REQUIRED_RX`, `_imports_mcp`)
- Tests: inline + multiline headers, plain REST backoff, TODO-only path, idempotency
- Updated `cookbook/14-routing-headers.md` fixer reference

## Tests

```
pytest tests/test_fixers.py -k r003  # 6 passed
pytest  # 292 passed
```

## Notes

- Method/name values are inferred from nearby `json=` literals/identifiers when possible; otherwise placeholders like `"<set-mcp-method>"` are used (review confidence).
- Does not add a new `headers=` kwarg when one is absent — only TODO.

Fixes #16

### DIFF
--- a/cookbook/14-routing-headers.md
+++ b/cookbook/14-routing-headers.md
@@ -1,7 +1,7 @@
 # Required `Mcp-Method` / `Mcp-Name` routing headers
 
 - **Rule:** [R003](../src/mcp_migrate/rules/r003_routing_headers.py)
-- **Fixer:** none
+- **Fixer:** [`RoutingHeadersFixer`](../src/mcp_migrate/fixers/r003_routing_headers.py) (`review` confidence — adds headers to inline `headers={...}` dicts and leaves TODOs elsewhere)
 - **Severity:** advisory (downgraded from an earlier `breaking` after a real
   false-positive incident -- see the rule source's comment for the
   mcp-atlassian story before reaching for a fixer here)
--- a/src/mcp_migrate/fixers/r003_routing_headers.py
+++ b/src/mcp_migrate/fixers/r003_routing_headers.py
@@ -1,0 +1,259 @@
+"""Fixer for R003 -- Mcp-Method / Mcp-Name routing headers on hand-rolled HTTP.
+
+Hand-rolled MCP wire traffic must carry routing headers so proxies can
+dispatch without parsing the body. There is no single mechanical value for
+every call site (method/name may be dynamic), so this fixer handles only
+the narrowest safe shape -- an inline ``headers={...}`` dict on a
+``.post()``/``.request()`` call -- and leaves a TODO everywhere else.
+Confidence "review": header values still need a human pass.
+"""
+from __future__ import annotations
+
+import ast
+import re
+from pathlib import Path
+
+from mcp_migrate.rules.r003_routing_headers import (
+    HAND_ROLLED_CALL,
+    HTTP_LIBS,
+    MCP_METHOD_RX,
+    NAME_REQUIRED_RX,
+    _imports_mcp,
+)
+
+from ._textedit import find_matching_close, leading_ws
+from .base import Fixer, FixResult
+
+COOKBOOK = "cookbook/14-routing-headers.md"
+SPEC_URL = (
+    "https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http"
+)
+TODO = (
+    "# TODO(mcp-migrate): add Mcp-Method"
+    " (and Mcp-Name for tools/call, resources/read, or prompts/get)"
+    f" — see {SPEC_URL} and {COOKBOOK}"
+)
+
+POST_CALL_RX = re.compile(r"\.(?:post|request)\s*\(")
+HEADERS_KW_RX = re.compile(r"\bheaders\s*=\s*\{")
+METHOD_LITERAL_RX = re.compile(r"""["']method["']\s*:\s*["']([^"']+)["']""")
+METHOD_IDENT_RX = re.compile(r"""["']method["']\s*:\s*([A-Za-z_][A-Za-z0-9_]*)""")
+NAME_IDENT_RX = re.compile(r"""["']name["']\s*:\s*([A-Za-z_][A-Za-z0-9_]*)""")
+
+
+def _file_context(source: str):
+    """Return gating flags mirroring the R003 rule's per-file checks."""
+    tree = None
+    try:
+        tree = ast.parse(source)
+    except SyntaxError:
+        pass
+
+    class _F:
+        def __init__(self, text: str, tree):
+            self.text = text
+            self.tree = tree
+            self.lines = text.splitlines()
+
+    f = _F(source, tree)
+    has_http = any(lib in source for lib in HTTP_LIBS)
+    mcp_surface = bool(MCP_METHOD_RX.search(source)) or _imports_mcp(f)
+    requires_name = bool(NAME_REQUIRED_RX.search(source))
+    missing_method = "Mcp-Method" not in source
+    missing_name = requires_name and "Mcp-Name" not in source
+    return has_http, mcp_surface, requires_name, missing_method, missing_name
+
+
+def _method_value(call_span: str) -> str:
+    m = METHOD_LITERAL_RX.search(call_span)
+    if m:
+        return f'"{m.group(1)}"'
+    m = METHOD_IDENT_RX.search(call_span)
+    if m:
+        return m.group(1)
+    return '"<set-mcp-method>"'
+
+
+def _name_value(call_span: str) -> str:
+    m = NAME_IDENT_RX.search(call_span)
+    if m:
+        return m.group(1)
+    return '"<set-mcp-name>"'
+
+
+def _insert_headers_entries(
+    lines: list[str],
+    open_i: int,
+    open_col: int,
+    close_i: int,
+    close_col: int,
+    *,
+    add_method: bool,
+    add_name: bool,
+    method_val: str,
+    name_val: str,
+) -> None:
+    """Insert Mcp-Method / Mcp-Name entries before the closing ``}``."""
+    entries: list[str] = []
+    if add_method:
+        entries.append(f'"Mcp-Method": {method_val}')
+    if add_name:
+        entries.append(f'"Mcp-Name": {name_val}')
+    if not entries:
+        return
+
+    if open_i == close_i:
+        # Single-line headers={...} — splice comma-separated entries before ``}``.
+        line = lines[close_i]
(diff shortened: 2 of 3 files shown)


dheerajjha:
Sorry for the slow ack. This is the most careful fixer anyone has written for this repo — `find_matching_close` for brace balancing instead of regex-guessing, a separate mechanical path and TODO path, reverse-sorted insertions so earlier edits don't invalidate later offsets, and reusing the rule's own gating constants so the fixer and the rule can't drift on what counts as MCP surface. That last one I initially wrote down as a concern and then changed my mind: duplicating `HTTP_LIBS` and `MCP_METHOD_RX` into the fixer would be worse. (`_imports_mcp` is underscore-private, so if this stays, that one probably wants a public name.)

Two bugs, both reproduced on this branch.

**1. A fixed call also gets a "you need to fix this" TODO.** `fixed_call_lines` records line indices during pass 1, but `_insert_headers_entries` calls `lines.insert(...)`, so every mechanical fix shifts every index below it. Pass 2 then walks the *mutated* list using the *stale* indices. With two fixable calls in one file:

```python
14 def call_two(client):
15     # TODO(mcp-migrate): add Mcp-Method (and Mcp-Name for tools/call, ...)
16     return client.post(
17         "/mcp",
18         headers={
19             "Content-Type": "application/json",
20             "Mcp-Method": "tools/call",
21             "Mcp-Name": "<set-mcp-name>",
22         },
```

The TODO says to add `Mcp-Method` immediately above a call that now has `Mcp-Method`. One fixable call is fine; two is enough to break it, because the reverse-sort protects pass 1's own offsets but nothing carries the shift into pass 2.

Cheapest fix is to make pass 2 not depend on line numbers at all — do both passes on the same list, or record something stable (the matched call's identity, or a sentinel inserted during pass 1) instead of an index into a list you're about to mutate.

**2. `Mcp-Name` is decided per file, so it leaks onto calls that don't need it.** `requires_name` comes from `NAME_REQUIRED_RX.search(source)` — the whole file. Isolated:

```
tools/list ALONE in the file:                 Mcp-Name inserted?  False
same tools/list, tools/call also in the file: Mcp-Name inserted?  True
    headers={..., "Mcp-Method": "tools/list", "Mcp-Name": "<set-mcp-name>"}
```

Identical call, opposite treatment, decided by an unrelated function elsewhere in the file. The rule computes `requires_name` per file too, so you've mirrored it faithfully — but the rule only *reports*, and a coarse gate that's acceptable for reporting isn't acceptable for editing. You already compute `call_span` per call site; deriving `requires_name` from that instead of from `source` fixes it exactly where it's wrong.

**One judgement call, not a defect.** `"<set-mcp-method>"` and `"<set-mcp-name>"` are written into the source as valid string literals, so the code keeps running and starts sending a header whose value is `<set-mcp-name>`. Given `confidence = "review"` I can see the argument that a human is required to look anyway. My instinct is that a placeholder which *runs* is more dangerous than one that doesn't, because it survives a skim — where a bare TODO and no header at all leaves the call in its current, honestly-broken state. What made you pick the placeholder? If there's a reason I'm not seeing I'll defer to it.

Not blocking on the string-literal issue (a TODO inserted into a docstring changes its contents) — that's #105, it affects R017 and R019 on `main` already, and it isn't yours to carry.

Fix those two and I'll merge. This one's genuinely good work.

dheerajjha:
@syf2211 — no rush, just flagging this is still open on the two bugs from my review (the stale line indices in pass 2, and `requires_name` being computed per file rather than per call site).

Context that may matter for how you fix the second one: **#102 merged an hour ago**, so `main` now has `sdk.py` and a changed `cmd_entry`. Nothing that touches your fixer, but worth rebasing onto before you push so CI runs against current `main`.

Also — six of the twenty-six PRs in yesterday's v0.1.3 were yours, and the release notes say so.

dheerajjha:
Checking back — no rush, and this is not a ping to hurry you. **Bug 2 still reproduces on the current branch**, so here is a concrete case you can drop straight into a test rather than re-deriving it from my prose.

```python
import httpx

def a(client):
    return client.post(url, headers={"Content-Type": "application/json"}, json={"method": "tools/list"})

def b(client):
    return client.post(url, headers={"Content-Type": "application/json"}, json={"method": "tools/call", "name": "x"})
```

`fix --write` on this branch produces:

```python
    ... "Mcp-Method": "tools/list", "Mcp-Name": "<set-mcp-name>" ...   # <-- wrong
    ... "Mcp-Method": "tools/call", "Mcp-Name": "<set-mcp-name>" ...   # correct
```

Site `a` is `tools/list`, which does not take `Mcp-Name` —

```python
# src/mcp_migrate/rules/r003_routing_headers.py:27
NAME_REQUIRED_RX = re.compile(r"tools/call|resources/read|prompts/get")
```

The cause is line 141 of the fixer:

```python
requires_name = bool(NAME_REQUIRED_RX.search(f.text))
```

`f.text` is the whole file, so one `tools/call` anywhere makes **every** call site in that file get an `Mcp-Name`. Deciding it per call site — off the same `json={...}` literal you already parse for `Mcp-Method` — is the fix.

`Mcp-Method` itself is correct per site in both, so the per-site plumbing already exists; it is only `requires_name` that is hoisted.

I could not get bug 1 (stale line indices in pass 2) to reproduce on the two-call-site shape above — the output was clean. **It may be narrower than I implied or it may already be gone; treat my original description as unconfirmed rather than as a second thing you owe me.**

## Standing offer

You have been offline since 2026-08-04 and there is no deadline here. But if you would rather not come back to a three-day-old review: **say the word and I will push the `requires_name` fix to your branch and merge it under your name.** I have done that for two contributors this week. Six of the twenty-six PRs in 0.1.3 were yours — this is a one-line hoist, not a reason for the PR to go stale.

dheerajjha:
Landed as #208 — **your fixer and cookbook recipe verbatim, your six tests re-applied intact, your authorship on the commit.** Closing this because the work is in `main`, not because it was turned down. Sorry it sat for three days.

@waterlemonnn opened #206 for the same issue today. **Yours was three days earlier and does more** — theirs annotates only, yours actually writes the headers when it can recover them. That inference is the part worth having, and I verified it works:

```python
headers={"Content-Type": "application/json"},
json={"method": "tools/call", "params": {"name": name}},
```
becomes
```python
headers={"Content-Type": "application/json", "Mcp-Method": "tools/call", "Mcp-Name": name},
```

Correct on both values, from the call site alone.

## Two changes on top, and they were the same bug

**The placeholder fallback had to go.** When the method couldn't be recovered, the fixer wrote `"Mcp-Method": "<set-mcp-method>"`. That compiles, runs, and sends a literal `Mcp-Method: <set-mcp-method>` header on every request. I ran the round trip:

```
$ mcp-migrate check .          # R003 advisory, Grade A (97/100)
$ mcp-migrate fix . --write    # writes "<set-mcp-method>"
$ mcp-migrate check .          # Grade A. Nothing to fix.
```

The tool flagged a real problem, replaced it with a broken header, then declared the file clean. That is worse than the finding it replaced — the finding was actionable. The fixer already had a TODO path for call sites it couldn't handle; it now falls through to that instead of guessing.

**And R003 itself was silencing-able.** The rule decided "is this header set?" with `"Mcp-Method" not in f.text` — a raw substring test over the whole file, comments included. Your TODO text names the header. So did @waterlemonnn's in #206. Either fixer would have made `fix --write` stop the rule firing without setting anything:

```
before:                          ['R003']
after adding ONLY a TODO comment: []
```

Fixed with `search_wire`, which keeps ordinary string literals — where `headers={"Mcp-Method": ...}` lives — and skips comments, where a TODO lives. `search_code` is the wrong tool here: it skips string literals, so it would never see a real header either.

That is a **pre-existing rule bug**, not something you introduced. It just needed a fixer to exist before anyone could trip it.

## Also worth flagging

I initially rebuilt this by copying your branch's whole `rules/r003_routing_headers.py`, which silently reverted a `_check_ts` optimisation that landed on `main` after you branched. `tests/test_scan_complexity.py` caught it (`R003: 20 -> 200` scans). Your PR never touched the rule — that was mine, and the guard did its job.

**End state, verified:** the inferable call site gets real values and stops firing; the uninferable one keeps its source untouched, gets a TODO, and **still reports R003** — because it still isn't fixed. **532 tests pass.**

Issue #99, opened 2026-08-04
R018/R019 miss the SDK schema names, which is what real servers write


dheerajjha:
R018 and R019 detect the **wire** spelling of the removed methods but not the **SDK schema** spelling — which is the one real TypeScript servers actually use.

Verified on `main`:

```ts
import { ListTasksRequestSchema, GetTaskPayloadRequestSchema } from "@modelcontextprotocol/sdk/types.js";
server.setRequestHandler(ListTasksRequestSchema, async () => ({ tasks: [] }));
```

→ **R019 reports nothing.**

```ts
export async function onReply(params: { elicitationId: string }) {
  return store.resolve(params.elicitationId);
}
```

→ **R018 reports nothing**, even though `elicitationId` is named in R018's own fix text as removed.

### Why this is the shape that matters

A server using the low-level SDK never writes the wire string. It references the Zod schema:

```ts
server.setRequestHandler(ListTasksRequestSchema, handler);   // no "tasks/list" anywhere
```

`search_wire` only sees string literals, so a rule that searches only the wire form is blind to the single most common real implementation. The tool stays silent on exactly the servers most likely to be broken.

### R012 already does this correctly — copy it

This is not a design question, it's an inconsistency. R012 catches both halves and is the model to follow:

```ts
import { SetLevelRequestSchema } from "@modelcontextprotocol/sdk/types.js";
server.setRequestHandler(SetLevelRequestSchema, async () => ({}));
```

→ R012 **detects** this. R009 and R011 do the same thing with `InitializeRequest\w*` / `PingRequest\w*`.

So: add a `search_code` pass for the SDK identifiers alongside the existing `search_wire` pass.

### Do not reintroduce #87 while fixing this

#87 is open because `PingRequest\w*` also matches `PingRequester`. **Bound the suffix from the start** rather than copying the unbounded pattern:

```python
TASKS_CODE_RX = r"\b(?:ListTasks|GetTaskPayload)Request(?:Params|Schema)?\b"
```

Confirm the real names against [`@modelcontextprotocol/sdk`](https://github.com/modelcontextprotocol/typescript-sdk) rather than trusting mine.

### Also worth handling: one line, one finding

A dispatcher can carry both signals on the same line:

```ts
case "tasks/list": return this.handleListTasks(ListTasksRequestSchema);
```

That's one problem and should be reported once. R009 already dedupes exactly this way — mirror it.

### Done when

[...]

Issue #91, opened 2026-08-04
R015/R004/R016 report correct code that is spelled indirectly


dheerajjha:
Three rules require a *literal spelling* of the correct thing, and report a finding when the code is correct but spelled differently. Grouped because they share one theme: we can't see through one level of indirection.

All three reproduced on `main`.

### R015 — an outgoing request is not a result

```ts
const request = { jsonrpc: "2.0", method: "tools/list", id: 1 };
send(request);
```

Reports *"builds JSON-RPC responses for MCP methods without `resultType`"*. That object is a **request**. `resultType` is a result-only field; requiring it here is asking for a field that must not exist. Any client, or any server that also makes outbound calls, trips this.

The rule needs to tell a request from a result. `id` + `method` with no `result` key is a request; `result`/`error` alongside `id` is a response. Neither is subtle, but right now nothing distinguishes them.

**This is the highest-severity of the three** — it's `breaking`, and it fires on an extremely ordinary shape.

### R004 — a sorting helper is still sorting

```ts
const alphabetize = (xs: Tool[]) => xs.toSorted((a, b) => a.name.localeCompare(b.name));
server.setRequestHandler(ListToolsRequestSchema, () => {
  return { tools: alphabetize(registry) };
});
```

Reports *"Tools are returned without an explicit sort."* They are explicitly sorted; the sort lives one function away. The rule matches a fixed list of sorting spellings **inside the handler body**, so extracting a helper — which is better code — makes the finding appear.

Honest question for the thread: how far should this chase? Following one level of local indirection is tractable. Following arbitrary call graphs is not, and this is a regex-based tool by design. It may be that the right answer is to **downgrade the finding when the handler returns a call result** rather than to resolve the call — silence is an acceptable output here.

### R016 — computed property keys

```ts
const TTL = "ttl" + "Ms";
server.setRequestHandler("tools/list", async () => ({ tools: [], [TTL]: 1000 }));
```

Reports the metadata as absent. It isn't; it's constructed. **Lowest priority of the three** — this is a genuinely exotic way to write it and I'd accept a "won't fix" with a comment in the rule explaining the limit. Listed for completeness, not because I think it's urgent.

### Suggested order

R015 first and separately — it's `breaking` and the fix is well-defined. R004 wants a design conversation first. R016 may want nothing but a documented limitation.

Take any one of them; say which in the thread so nobody doubles up.

dheerajjha:
[...]

Issue #89, opened 2026-08-04
Five rules decide 'is this MCP?' too loosely and fire on unrelated code


dheerajjha:
Five rules decide "is this MCP?" from a signal that is too weak, and then report `breaking` or `deprecated` on code that has nothing to do with the protocol.

### Reproduced, all five on `main`

| Rule | Code that wrongly fires | What the gate actually is |
| --- | --- | --- |
| R003 | `import { Server } from "@modelcontextprotocol/sdk/server";`<br>`await fetch("https://billing.example/invoices");` | An SDK import **anywhere in the file** makes every `fetch()` in it "hand-written MCP traffic" |
| R005 | `interface ServerCapabilities { maxConnections: number; }` | The bare name `ServerCapabilities`, with no SDK context required |
| R006 | `const iconUrl = "/sse/icons/status.svg";` | Any string starting `/sse/` |
| R011 | `import { Client } from "@modelcontextprotocol/sdk/client";`<br>`if (method === "ping") runHealthCheck();` | An SDK import turns every `method === "ping"` into MCP dispatch |
| R020 | `class ConnectionPool { registerClient(client: Client) {} }` | The method name `registerClient`, ungated |

Every one of these is code a real MCP server could plausibly contain **while being fully migrated**. A server that talks to a billing API, serves an SSE-based status page, has a `/health` endpoint, or pools connections gets punished for it.

R003 is the worst of them, and it is not hypothetical: an early iteration of this tool graded `mcp-atlassian` at F/0 almost entirely on Jira REST calls tripping the routing-header rule. That incident is why "a false positive is worse than a missed finding" is the first working principle in the README. The gate got narrower then; it is still too wide.

### Why this one needs judgment, not a regex

There is no single edit that fixes all five, and the honest tension is:

- **Tighten too little** and we keep publishing wrong `breaking` verdicts about other people's repos.
- **Tighten too much** and the rule silently stops finding the thing it exists for — which is worse, because nobody notices a missing finding.

Some directions, none of which I'm committed to:

- Gate on **proximity** rather than file-membership: an SDK import 400 lines away from a `fetch()` is not evidence about that `fetch()`.
- Gate on the **call target**, not the file: `server.` / `client.` / `session.` receivers, which is what R007 already does with `TS_OWNER`.
- For R005/R020, require an **SDK import in the same file** at minimum — currently there is no context requirement at all.
- For R006, require the `/sse` string to be used as a *transport endpoint*, not merely to appear.

[...]

PR #107, opened 2026-08-05
fix(rules): Add wire_method helper with (?![\w/-]) end boundary to prevent false positives on longer strings (#88)

### LINKED ISSUE #88: Wire patterns lack an end boundary: `"roots/listeners"` reports as `roots/list`

IronLad123:
Closes #88

## Summary

Adds a shared `wire_method(*names: str) -> str` helper function in `src/mcp_migrate/rules/base.py` that builds JSON-RPC wire method regex patterns bounded by `(?![\w/-])`.

## Rationale & Impact

- `search_wire()` keeps ordinary string literals. Previously, method-name patterns with no end boundary matched longer, unrelated strings (e.g. `"roots/listeners"` reported as `roots/list`, or `"tasks/listeners"` reported as `tasks/list`).
- Using `(?![\w/-])` ensures that method names end where they say they end, rejecting trailing word characters or slashes while still matching `"roots/list"` inside JSON-RPC envelopes or assignments.
- Updated wire pattern searches across **R007**, **R009**, **R012**, **R018**, and **R019**.

## Verification

- Added `test_wire_method_bounds_prevent_false_positives_on_longer_strings` and `test_wire_method_bounds_still_catch_exact_wire_method_names` in `tests/test_rules.py`.
- `pytest`: **225/225 passed** (`1.53s`).
- `ruff check`: **0 errors** across all modified files.

### DIFF
--- a/src/mcp_migrate/rules/base.py
+++ b/src/mcp_migrate/rules/base.py
@@ -383,3 +383,16 @@ class Rule:
             line=line,
             snippet=snippet,
         )
+
+
+def wire_method(*names: str) -> str:
+    r"""Build a regex pattern for JSON-RPC wire method names bounded by (?![\w/-]).
+
+    Prevents false positives on longer, unrelated strings like 'roots/listeners'
+    or 'tasks/listeners'.
+    """
+    if len(names) == 1:
+        return rf"\b{re.escape(names[0])}(?![\w/-])"
+    joined = "|".join(re.escape(n) for n in names)
+    return rf"\b(?:{joined})(?![\w/-])"
+
--- a/src/mcp_migrate/rules/r007_deprecated_features.py
+++ b/src/mcp_migrate/rules/r007_deprecated_features.py
@@ -1,4 +1,4 @@
-from .base import Finding, Project, Rule
+from .base import Finding, Project, Rule, wire_method
 
 # Every pattern here has to be unambiguously MCP. A bare `create_message`
 # was not: it has no word boundary and no MCP context, so it matched
@@ -13,7 +13,7 @@ from .base import Finding, Project, Rule
 # API -- so the `session.` qualifier keeps every true positive while
 # dropping the collision.
 FEATURES = {
-    r"\broots/list\b|list_roots|RootsCapability": ("Roots", "Use resource URIs instead."),
+    rf"{wire_method('roots/list')}|list_roots|RootsCapability": ("Roots", "Use resource URIs instead."),
     r"sampling/createMessage|SamplingCapability|\bsession\.create_message\b"
     r"|\bCreateMessageRequest\b|\bCreateMessageResult\b": (
         "Sampling", "Sampling is deprecated; plan a migration."),
--- a/src/mcp_migrate/rules/r009_initialize_handshake_removed.py
+++ b/src/mcp_migrate/rules/r009_initialize_handshake_removed.py
@@ -1,6 +1,6 @@
 import re
 
-from .base import Finding, Project, Rule
+from .base import Finding, Project, Rule, wire_method
 
 # `InitializeRequest`, `InitializeResult` and `InitializedNotification` are
 # the MCP SDK's own pydantic model names for the initialize handshake --
@@ -40,7 +40,7 @@ class InitializeHandshakeStillImplemented(Rule):
         # so it always starts inside a STRING token and search_code would
         # silently never find it. Match the literal directly instead, the
         # same way r004_tool_ordering.py matches the literal `tools/list`.
-        for f, line, text in project.search_wire(r"notifications/initialized"):
+        for f, line, text in project.search_wire(wire_method("notifications/initialized")):
             out.append(self.finding(
                 "References the removed notifications/initialized handshake message.",
                 f, line, text,
--- a/src/mcp_migrate/rules/r012_logging_set_level_removed.py
+++ b/src/mcp_migrate/rules/r012_logging_set_level_removed.py
@@ -1,6 +1,6 @@
 import re
 
-from .base import Finding, Project, Rule
+from .base import Finding, Project, Rule, wire_method
 
 # `SetLevelRequest`/`SetLevelRequestParams` are the MCP SDK's own model
 # names for this request -- distinctive, no false-positive risk.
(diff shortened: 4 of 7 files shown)


dheerajjha:
**#97 and #103 are merged.** Thank you for turning both around quickly — I verified the fixes rather than assuming:

```
R021  "$schema": "https://json-schema.org/draft-07/schema#"  ->  "https://json-schema.org/draft/2020-12/schema"
R021  "$schema": "http://json-schema.org/draft-04/schema"    ->  "https://json-schema.org/draft/2020-12/schema"
R020  server.py  ->  # TODO(mcp-migrate): RFC 7591 Dynamic Client Registration...
R020  index.ts   ->  // TODO(mcp-migrate): RFC 7591 Dynamic Client Registration...
R020  idempotent: True     output parses: True
```

`main` is at 299 tests, green. R020 is the first fixer here that emits the right comment marker for TypeScript as well as Python.

**Now the thing that's costing you more than any individual review.** You have five PRs still open and every one of them is `CONFLICTING`, in two chains:

```
A:  #104                          (was on #103 — now merged, needs a rebase)
B:  #107 -> #108 -> #109 -> #110  (each contains all the commits before it)
```

#110 carries four commits and touches fifteen files, of which two are its own. The practical effect is that **#110 cannot merge until #107 is correct**, so one review comment anywhere in the chain freezes everything downstream of it. That's why nothing of yours moved for a day while #97 sat on a one-character fix.

Rebasing each branch onto current `main` so it contains only its own change would let me take them in any order, as they become ready, instead of in a queue. Four small PRs land faster than four stacked ones even when the total work is identical.

**And one ask, which is about the project rather than your code.** You currently hold #28, #29, #32, #87, #88, #99 and #105. Every one of those was labelled `good first issue` or `up-for-grabs`.

This morning I assigned #88 to @Shxnque, a first-time contributor, and had to retract it twenty minutes later because #107 was already open against it. That was my error — I checked the assignee and not the PR list — but the reason there was a collision to have is that the small, well-scoped, clearly-specified issues are the ones a newcomer can actually start on, and they're the ones being cleared fastest.

I'm not asking you to slow down, and I'm not going to ration work to someone who ships. What I'd ask is that you leave the ones tagged `good first issue` for a bit and take the ones nobody else can: **#91** (three rules reporting correct code that's spelled indirectly), **#92** (re-verifying fourteen published grades against the false positives we've now confirmed), or **#89** (five rules deciding "is this MCP?" too loosely). Those are `size: M`, genuinely hard, and you have more context on this rule engine than anyone except me at this point.

That's a request, not a rule. If you'd rather keep going as you are, say so and I'll stop assigning first-issue-labelled work to anyone else without checking with you first — which is the honest alternative and I'd rather do that than have you find out by collision.

dheerajjha:
Following up with an offer rather than a repeat of the ask.

These four have been frozen for a day, and between them they fix #87, #88, #99 and #105 — which are four of the false-positive classes currently shipping in v0.1.3 and named in its release notes. They matter more than anything else open right now, and the only thing wrong with them is the shape they're stacked in.

**If the un-stacking is the tedious part, I'll do it.** Say the word and I'll cherry-pick each commit onto current `main`, push each to its own branch on your fork (`maintainerCanModify` is on), and force-update these four PRs so each contains only its own change. Your commits, your authorship, your PRs — I'd just be doing the git plumbing.

If you'd rather do it yourself, that's completely fine and I'll wait. I'd just rather ask than let four FP fixes sit while the false positives they fix are live on PyPI.

Separately, since you claimed #105 and #87 on the issues yesterday: both are assigned to you, and #108 and #109 are the PRs for them, so nothing further is needed there.

Two more merges landed today — #102 (SDK grading, closes #31) and #112 — so `main` is at 310 tests and has moved under all four of these.

dheerajjha:
Merged. Sorry this sat conflicting for a few days — the conflict was mine, from a message-constant refactor in R009/R012/R019 that landed after you branched.

**The bug is real and still was.** Before merging I checked it still reproduces on current `main` rather than taking the issue at its word:

```
false positives on main: R009:5  R012:6  R018:4  R018:8  R019:7
```

on a file containing `roots/listeners`, `notifications/initializedAt`, `logging/setLevelPolicy`, `tasks/listeners`. Five wrong findings, exactly the class #88 describes. Zero now.

Two changes I made on top of what you submitted, both worth explaining:

**1. Bound at the constant, not at the call site.** `WIRE_RX` in R009 and R019, and the `logging/setLevel` literal in R012, each feed *two* call sites — the Python path and the TypeScript one. Your patch reached the Python sites, so the same file scanned as TypeScript still produced the false positives:

```
python      false positives: NONE
typescript  false positives: R007:2  R018:2
```

R007's TS literal and R018's `FEATURES_LITERAL` had no boundary at all. Moving `wire_method()` to the definition fixes both languages from one place, and is why the conflict resolved cleanly rather than needing a choice between your version and mine.

**2. Your own test caught a bug I introduced.** With the dict keys bounded at the definition, R018's Python path was still wrapping them again — `wire_method(wire_method(x))` escapes the inner pattern's metacharacters and matches nothing. Silent false *negative*, invisible to the false-positive test. `test_wire_method_bounds_still_catch_exact_wire_method_names` went red immediately.

That is exactly why the companion test belongs in the PR, and a lot of people wouldn't have written it. It's the discipline this repo needs more of, not less.

I added the TypeScript halves of both directions, so the boundary is now pinned in four tests across two languages. 597 pass.

I'll look at #108 and #110 next — both still conflicting, and I'd rather resolve them than ask you to rebase work that's been ready since Wednesday.

PR #108, opened 2026-08-05
fix(fixers): Prevent line-based fixers from mutating lines inside string literals (#105)

### LINKED ISSUE #105: Fixers insert comments inside string literals, silently changing their value

IronLad123:
Closes #105

## Summary

Prevents line-by-line fixers (such as R001, R006, R017) from inserting comment lines or replacing text inside multiline Python docstrings (`"""` and `'''`) or TypeScript template literals (` `...` `), preserving AST values and data integrity.

## Key Changes

1. **`string_lines(source, path_or_lang)` Helper**:
   - Added to `src/mcp_migrate/fixers/_textedit.py`.
   - Uses `tokenize` in Python to track triple-quoted string spans (`"""` / `'''`).
   - Uses a single-pass scanner in TypeScript to track multiline template literals (` `...` `).
   - Fail-safe design: if tokenization or scanning fails on malformed source, returns all 1..N line numbers so fixers decline to edit rather than risk silent corruption.

2. **Fixer Guard Checks**:
   - Updated `SessionIdHeaderFixer` (R001), `SortToolsFixer` (R004), `ExtensionsFixer` (R005), `SseTransportFixer` (R006), and `ResourceNotFoundErrorCodeFixer` (R017) to query `string_lines()` and skip lines residing inside multiline strings.

3. **Verification**:
   - Added unit test `test_fixers_decline_to_edit_inside_docstrings_issue_105` running the maintainer's reproduction snippet over all registered fixers, confirming `r.changed == False` and AST equality before and after.
   - Verified that symbols on actual code lines adjacent to docstrings are still accurately fixed.
   - Tested idempotency across all fixers.
   - **229 / 229 tests passed** (`1.53s`).

### DIFF
--- a/src/mcp_migrate/fixers/_textedit.py
+++ b/src/mcp_migrate/fixers/_textedit.py
@@ -9,6 +9,8 @@ splice, never an `ast.unparse` round-trip.
 """
 from __future__ import annotations
 
+from pathlib import Path
+
 OPEN = "([{"
 CLOSE = ")]}"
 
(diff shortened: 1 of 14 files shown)


dheerajjha:

dheerajjha:
Merged — full review on #110, which landed alongside it. Short version: the docstring bug reproduced exactly as you described on current `main`, including a `safe`-confidence fixer rewriting `-32002` → `-32602` inside prose about historical behaviour and leaving the file parseable, so nothing caught it.

One addition: the **R019 fixer needed the same guard**. You covered R001/R004/R005/R006/R017, but your own test asserts every fixer declines, and `TasksPollingFixer` is one — it was still writing `# TODO(...)` into the string literal. Guarding it is what makes that test true rather than aspirational, and it's the right test to have written.

PR #109, opened 2026-08-05
fix(rules): Replace unbounded \w* suffix in R007/R009/R011/R012 with bounded alternation (#87)

### LINKED ISSUE #87: Unbounded \w* suffix makes 4 rules fire on unrelated identifiers like `PingRequester`

IronLad123:
Closes #87

## Summary

Replaces the unbounded `\w*` regex suffix in rules **R007**, **R009**, **R011**, and **R012** with bounded alternation (`(?:Params|Schema|Result|ResultSchema)?`) to prevent false positives on common, unrelated class names like `PingRequester`, `CreateMessageRequester`, `InitializeRequester`, and `SetLevelRequester`.

## Key Changes

1. **R007 (`r007_deprecated_features.py`)**:
   - Updated `CreateMessageRequest` and `CreateMessageResult` regex patterns to use bounded suffix alternation `\bCreateMessageRequest(?:Params|Schema)?\b` and `\bCreateMessageResult(?:Schema)?\b`.

2. **R009 (`r009_initialize_handshake_removed.py`)**:
   - Updated `HANDSHAKE_CODE_RX` pattern to use bounded suffix alternation for `InitializeRequest`, `InitializeResult`, and `InitializedNotification`.

3. **R011 (`r011_ping_removed.py`)**:
   - Updated `PING_CODE_RX` and `TS_PING_CODE_RX` from `r"\bPingRequest\w*"` to `r"\bPingRequest(?:Params|Schema|Result|ResultSchema)?\b"`.

4. **R012 (`r012_logging_set_level_removed.py`)**:
   - Updated `SET_LEVEL_CODE_RX` to `r"\bSetLevelRequest(?:Params|Schema|Result|ResultSchema)?\b"`.

5. **Regression Tests (`tests/test_rules.py`)**:
   - Added `test_unbounded_suffix_bounded_pattern_issue_87` asserting that `...Requester` class shapes stay silent (no false positives) while `...Schema` and `...Params` shapes continue to fire.
   - **229 / 229 tests passed** (`1.69s`).

### DIFF
--- a/src/mcp_migrate/fixers/_textedit.py
+++ b/src/mcp_migrate/fixers/_textedit.py
@@ -9,6 +9,8 @@ splice, never an `ast.unparse` round-trip.
 """
 from __future__ import annotations
 
+from pathlib import Path
+
 OPEN = "([{"
 CLOSE = ")]}"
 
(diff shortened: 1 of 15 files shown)


dheerajjha:

dheerajjha:
Flagging a change in this PR's standing rather than chasing the rebase again: **#109 became the blocker for three other PRs today.**

@waterlemonnn opened fixers for R011 (#124), R012 (#125) and R009 (#141). Each one copies the unbounded suffix out of the rule it serves — `\bPingRequest\w*`, `\bSetLevelRequest\w*`, `\bInitializeRequest\w*` — which is precisely the set this PR bounds. I am holding all three, because in a *fixer* that pattern stops being a wrong finding and becomes a wrong edit:

```
$ mcp-migrate check b.py
Grade A. Nothing to fix.

$ mcp-migrate fix b.py --write     # with #141 applied
# TODO(mcp-migrate): the initialize/notifications/initialized handshake is removed; ...
# helper = InitializeRequesterHelper()
result = helper.run()
```

A clean file, graded A, and `--write` comments out the assignment and leaves `helper.run()` pointing at a name that no longer exists.

**Your `(?:Params|Schema|Result|ResultSchema)?` is the right shape and I would rather land it here, once, than have three fixers each invent their own spelling.** I have pointed all three PRs at this one.

The blocker is still the same as on 2026-08-05: this sits on top of #107 and #108 and the stack has been `CONFLICTING` for two days. **That offer to un-stack it for you is still open** — I rebased #116 and #127 onto `main` for @waterlemonnn today, so it is a thing I am doing anyway rather than a favour I am promising. Say the word, or push a rebase yourself, and this goes in.

Separately, #104 is now the single most valuable open PR in the repo: R013, R014 and R021 landed today, so **R002 is the only rule left that does not read TypeScript, and #104 is the port.** Commented there too.

dheerajjha:
Un-stacked this and landed it as #207 — **your rule changes verbatim, your test applied as written, your authorship on the commit.** Closing this one because the work is in `main`, not because it was rejected.

Same reason as #104: this branch carried #107's and #108's commits underneath it, so it could not merge without them, and those two need more discussion. #109 did not — it was correct and self-contained, and it had been waiting.

## One thing I extended

`main` grew a **separate TypeScript pattern** in each of these four rules after you branched — `TS_PING_CODE_RX`, `TS_SET_LEVEL_CODE_RX`, and the TS arms in R007 and R009. Those are a second set of `\w*` sites your branch predates.

Bounding only the Python half would have left the identical false positive live for every TypeScript project, which is most MCP servers and the reason #87 was filed. Both halves are bounded now, using your alternation.

## Verified

18 false-positive and 18 true-positive cases, each in both `.py` and `.ts`. Every FP silent, every real SDK name still caught, nothing lost to overshoot. Parametrized so both directions stay covered.

## What this actually fixed

R009, R011 and R012's **fixers** had already bounded their own patterns while the rules had not — so `check` reported lines that `fix` then refused to touch, and the CHANGELOG documented that gap as deliberate. Three fixers shipped working *around* this bug. They now agree with their rules instead.

**509 tests pass.** #87 closed.

**#107, #108 and #110 are still open** and still yours — this only removes #109 from underneath them. If you want to rebase #107 onto current `main` it should be much simpler now. Thanks for this one; it had more reach than its diff suggested.

PR #110, opened 2026-08-05
fix(rules): Add SDK schema and params names to R018/R019 code rules (#99)

### LINKED ISSUE #99: R018/R019 miss the SDK schema names, which is what real servers write

IronLad123:
### Summary of Changes

Fixes #99 where rules **R018** (`MultiRoundTripReplacesServerInitiated`) and **R019** (`TasksPollingReplacesBlockingResult`) failed to detect SDK schema/params names (`ListRootsRequestSchema`, `CreateMessageResultParams`, `ElicitResultSchema`, `ListTasksResultSchema`, `GetTaskPayloadResultParams`), causing servers defining or consuming SDK schema types to stay silent.

1. **R018 (`r018_multi_round_trip_replaces_server_initiated.py`)**: Added bounded alternation for `ListRootsRequest(?:Params|Schema)?`, `ListRootsResult(?:Schema)?`, `CreateMessageRequest(?:Params|Schema)?`, `CreateMessageResult(?:Schema)?`, `ElicitRequest(?:Params|Schema)?`, `ElicitResult(?:Schema)?`, and `ElicitCompleteNotificationParams(?:Schema)?`.
2. **R019 (`r019_tasks_polling_replaces_blocking_result.py`)**: Added bounded alternation for `ListTasksRequest(?:Params|Schema)?`, `ListTasksResult(?:Schema)?`, `GetTaskPayloadRequest(?:Params|Schema)?`, and `GetTaskPayloadResult(?:Schema)?`.
3. **Tests (`tests/test_rules.py`)**: Added regression test `test_r018_r019_sdk_schema_names_issue_99` verifying both rules fire on SDK schema/params model declarations. All 230 tests passing.

### Test Verification
```bash
pytest
# 230 passed in 2.49s
```

### DIFF
--- a/src/mcp_migrate/fixers/_textedit.py
+++ b/src/mcp_migrate/fixers/_textedit.py
@@ -9,6 +9,8 @@ splice, never an `ast.unparse` round-trip.
 """
 from __future__ import annotations
 
+from pathlib import Path
+
 OPEN = "([{"
 CLOSE = ")]}"
 
@@ -50,3 +52,81 @@ def find_matching_close(lines: list[str], open_idx: int, open_col: int) -> tuple
 
 def leading_ws(line: str) -> str:
     return line[: len(line) - len(line.lstrip(" \t"))]
+
+
+def string_lines(source: str, path_or_lang: str | Path = "python") -> set[int]:
+    """Return 1-indexed line numbers in ``source`` that are part of a string literal.
+
+    - Python: tracks triple-quoted (''' and ''") and single/double quoted strings via tokenize.
+    - TypeScript / JavaScript: tracks backtick template literals (`...`), single/double quotes.
+
+    If parsing fails or encounters unterminated multi-line string state at EOF,
+    returns all line numbers (1..N) as a fail-safe so line-based fixers decline to edit.
+    """
+    import io
+    import tokenize
+
+    lines_list = source.splitlines()
+    total_lines = len(lines_list)
+    all_lines = set(range(1, total_lines + 1))
+    if total_lines == 0:
+        return set()
+
+    if isinstance(path_or_lang, Path):
+        lang = path_or_lang.suffix.lower()
+    else:
+        lang = str(path_or_lang).lower()
+
+    if lang in (".ts", ".tsx", ".js", ".jsx", "ts", "typescript", "js", "javascript"):
+        return _ts_string_lines(source, total_lines, all_lines)
+    return _py_string_lines(source, total_lines, all_lines, io, tokenize)
+
+
+def _py_string_lines(
+    source: str, total_lines: int, all_lines: set[int], io_mod: any, tok_mod: any
+) -> set[int]:
+    lines: set[int] = set()
+    try:
+        g = tok_mod.generate_tokens(io_mod.StringIO(source).readline)
+        for tok in g:
+            if tok.type == tok_mod.STRING:
+                s = tok.string.lstrip("rRbBuUfF")
+                if s.startswith('"""') or s.startswith("'''"):
+                    s_line = tok.start[0]
+                    e_line = tok.end[0]
+                    lines.update(range(s_line, e_line + 1))
+    except (tok_mod.TokenError, SyntaxError, IndentationError, ValueError):
+        return all_lines
+    return lines
+
+
+def _ts_string_lines(source: str, total_lines: int, all_lines: set[int]) -> set[int]:
+    lines: set[int] = set()
+    row = 1
+    i = 0
+    n = len(source)
+    in_template: bool = False
+
+    while i < n:
+        ch = source[i]
+        if not in_template:
+            if ch == "`":
+                in_template = True
+                lines.add(row)
+            elif ch == "\n":
+                row += 1
+        else:
+            lines.add(row)
+            if ch == "\\":
+                i += 1
+                if i < n and source[i] == "\n":
+                    row += 1
+            elif ch == "`":
+                in_template = False
+            elif ch == "\n":
+                row += 1
+        i += 1
+
+    if in_template:
+        return all_lines
+    return lines
--- a/src/mcp_migrate/fixers/r001_session_id.py
+++ b/src/mcp_migrate/fixers/r001_session_id.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import re
 from pathlib import Path
 
+from ._textedit import string_lines
 from .base import Fixer, FixResult
 
 SPEC_URL = "https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567"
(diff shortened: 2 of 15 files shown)


dheerajjha:

dheerajjha:
@IronLad123 — an update on this one you won't love, plus the reason your work here is still needed.

**#114 landed this morning and covers part of #99.** @waterlemonnn opened it overnight against the same issue. Before merging I checked whether it duplicated yours, because you claimed #99 first and I wasn't going to merge over you on a guess. It doesn't — I ran both branches against both halves:

**#110 against #99's TypeScript repro** (the one in the issue body):
```
(nothing)
```
**#114 against the same repro:**
```
R018  elicit.ts:1  elicitationId ...      R019  tasks.ts:1  ListTasksRequest ...
R018  elicit.ts:2  elicitationId ...      R019  tasks.ts:2  ListTasksRequest ...
```

Your R018/R019 changes go into `FEATURES_CODE` and `TASKS_CODE_RX`, both of which are read only on the Python path — R018 and R019 each `return` early for TypeScript through `TS_FEATURES_CODE` / the TS branch. So the Zod schema names you added land where a Zod schema name can never appear.

**Now the other direction**, which is why I reopened #99:

```python
from mcp.types import ListTasksRequestParams, GetTaskPayloadResult
def h(p: ListTasksRequestParams) -> GetTaskPayloadResult: ...
```

**#114 on this: nothing. #110 on this: two findings.** `ListTasksRequestParams` and `GetTaskPayloadResult` are real `mcp.types` exports, and `main` is still blind to them. That's your half, it's still broken, and nothing else open fixes it.

So: not a duplicate, and #99 is reopened and waiting on you.

**The honest part.** This one was ready a day ago and couldn't land, because it sits fourth in a chain behind #107. A PR that arrived twelve hours later merged first because it was one commit against `main`. I'm not saying that to score a point — I'm saying it because it's the concrete cost of the stacking, and it just came due.

**The rebase is now smaller than it was.** #114 pulled R019's constants out to module level, so your Python change reduces to widening one line:

```python
TASKS_CODE_RX = re.compile(r"\bListTasksRequest\b|\bGetTaskPayloadRequest\b")
```
→
```python
TASKS_CODE_RX = re.compile(
    r"\bListTasksRequest(?:Params|Schema)?\b|\bListTasksResult(?:Schema)?\b|"
    r"\bGetTaskPayloadRequest(?:Params|Schema)?\b|\bGetTaskPayloadResult(?:Schema)?\b"
)
```

That part is independent of #107 and could land today on its own. Your `wire_method(...)` wrapping genuinely does depend on #107's helper, so that half belongs with #107 — worth splitting them.

**My offer from yesterday stands and I'll extend it:** say the word and I'll rebase all four onto current `main`, push each to its own branch on your fork, and force-update each PR so it holds only its own change — your commits, your authorship. Or if you'd rather just take R018/R019, reply here and I'll leave the rest alone.

If I don't hear back I'm not going to touch your branches. But I'd rather not watch #87, #88 and #105 miss another release the way this one just did.

dheerajjha:
Merged, and #108 before it — all three of your PRs are in.

**#108.** I reproduced the bug on current `main` before merging. Two fixers edited *inside* a docstring:

```
ResourceNotFoundErrorCodeFixer:  -32002 -> -32602, inside prose describing
                                 historical behaviour, at `safe` confidence
TasksPollingFixer:               inserted `# TODO(...)` into a string literal,
                                 where `#` opens no comment
```

The first is the one that bothers me most: a `safe` fixer silently rewriting documentation so it states something factually wrong, while the file still parses so nothing catches it. Good find.

One change on top: the **R019 fixer needed the same guard**. You guarded R001/R004/R005/R006/R017, but your own test asserts *every* fixer declines — and `TasksPollingFixer` is a fixer. Guarding it is what makes that assertion true rather than aspirational. Verified afterwards that no fixer now mutates a Python docstring or a TypeScript template literal, and that the live code paths still fix.

**#110.** Same discipline — the miss reproduced:

```
main:  a file importing ListRootsRequestSchema, CreateMessageResultSchema,
       ElicitRequestParams, ListTasksRequestSchema, GetTaskPayloadResultSchema
       fires R007 only. R018 and R019 see nothing.
```

Both catch them now, which is the whole point of #99: the names real SDK code uses are not the names the rules were matching.

Three conflicts, resolved on merit rather than by side. Two went to `main`. **One went to yours, and it's worth calling out** — R011's comment on `main` said the rule matches "any identifier starting with `PingRequest`". #87 stopped that being true; the live regex matches four specific names and `PingRequester` does not match. Your comment described the code that exists. I checked before choosing.

**One consequence I want you to know about rather than discover.** Widening R018 into the `CreateMessage*` family means R007 and R018 now both fire on the same symbol, with different severities and both feeding the score. Both claims are true, so it isn't a defect in your PR — but it's noise, and it's filed as #221 rather than left for a user to hit.

I also checked this couldn't quietly move a published board grade: the widening can only *add* findings, 15 of 16 board entries have zero R018/R019 today, and `modelcontextprotocol/servers` contains none of the newly matched names — `src/fetch`, `src/git` and `src/time` all still reproduce B/91, D/54, B/91 exactly.

Sorry these three sat conflicting since Wednesday; the conflicts were mine, from refactors that landed after you branched. That's on me to notice faster, not on you to rebase repeatedly. 601 tests pass.

PR #112, opened 2026-08-05
docs: document search_wire and fix CONTRIBUTING rule example

### LINKED ISSUE #111: CONTRIBUTING teaches the search variant that causes our false positives
`CONTRIBUTING.md` is the first thing a new contributor reads and the file they copy their first rule from. Its guidance on choosing a search variant is now a version behind the code, and its one complete worked example demonstrates the exact mistake that produced #87, #88 and #89.

## 1. `search_wire()` is not documented at all

The "What a rule gets" list covers `search_code` and `search`. It does not mention `search_wire`, which is the variant the codebase actually settled on:

```
search_code    17 rule files
search_wire    16 rule files
project.search  0 rule files      <- documented as the option for wire names
```

Zero rules use raw `project.search()`. Every one that needs a JSON-RPC method name uses `search_wire()`.

## 2. The doc actively recommends the wrong one for wire names

Current text, under `project.search`:

> Only use this when the rule genuinely wants string *content* -- e.g. `r003_routing_headers.py` checking whether the literal header name `Mcp-Method` appears anywhere in the file, or any JSON-RPC method name with a slash in it (`tools/list`, `resources/subscribe`, ...), **which can only ever appear inside a string token and so `search_code` would never find it**

The premise is right and the conclusion is out of date. `search_wire()` exists precisely for that case, and its own docstring explains why raw `search` is wrong for it:

> a module docstring reading "Runs the MCP handshake (initialize -> notifications/initialized)" is documentation, not an implementation of it. **That exact line produced a `breaking` finding on a real server.**

So the failure mode is documented in the source and contradicted in the file contributors are told to read.

## 3. The complete worked example uses `search()` on a code construct

`CONTRIBUTING.md:117`, inside the `R022` example that new contributors are invited to copy:

```python
for f, line, text in project.search(r"[\"']experimental[\"']\s*:"):
```

A capabilities dict key is a code construct, so this should be `search_code`. Running the example rule verbatim:

```
1 finding(s)   real code            <- correct
1 finding(s)   a COMMENT about it   <- false positive
1 finding(s)   a docstring          <- false positive
1 finding(s)   a log message        <- false positive
```

Three false positives out of four cases, in the template. The prose two sections earlier says **"Use this [`search_code`] by default"** — so the example also contradicts the page it's on.

This matters more than a normal doc bug because the false positives currently shipping in published registry grades are all this family. Someone copying the template inherits the bug before they've written a line of their own logic.

## What's wanted

- Add `project.search_wire(pattern, *, flags=0)` to the list, positioned as the middle setting: skips comments and triple-quoted strings, keeps ordinary string literals. Point at any of the 16 rules using it.
- Rewrite the `project.search` entry to say what's now true — it matches prose, no shipped rule uses it, reach for it only if you have a reason neither other variant fits.
- Change the `R022` example to `search_code`, and add the comment/docstring negative case to the example's fixture list so the template teaches the fixture habit too.
- A short decision line worth having near the top, because it's the single commonest bug in this repo:
  > identifier or class name -> `search_code` · JSON-RPC method name -> `search_wire` · prose -> almost certainly neither

## Done when

The three variants are documented with the distinction between them stated, the worked example uses `search_code`, and running the example rule against a comment mentioning `"experimental":` produces no finding.

Docs-only, no code change, and reading the three docstrings in `rules/base.py` gives you everything you need. Good first issue for someone who wants to understand the rule engine before writing a rule — writing this up *is* the way to learn it.

AshSgDe29071999:
## Summary
`CONTRIBUTING.md` taught the search variant that produces false positives (#87–#89 family) and omitted `search_wire` entirely.

## Changes
- Document `project.search_wire` as the middle setting (string-literal wire text / JSON-RPC methods).
- Rewrite `project.search` guidance: no shipped rule uses it; last resort only.
- Add a one-line decision guide: identifier → `search_code`, wire name → `search_wire`, prose → neither.
- Change the R022 worked example to `search_code` and note a comment/docstring negative fixture.

Docs-only. Closes #111

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,11 @@ class Rule:
     def finding(self, message, f=None, line=None, snippet=None) -> Finding: ...
 ```
 
-`Project` gives you the whole scanned tree:
+`Project` gives you the whole scanned tree.
+
+**Quick pick:** identifier or class name → `search_code` · JSON-RPC method
+name (or other string-literal wire text) → `search_wire` · prose / comments →
+almost certainly neither (and no shipped rule uses raw `search`).
 
 - `project.search_code(pattern, *, flags=0)` -- like `search` below, but
   ignores matches that start inside a comment or a string/docstring
@@ -69,14 +73,20 @@ class Rule:
   class name, or code construct -- a docstring or `--help` string that
   merely *mentions* `Mcp-Session-Id` isn't a real usage of it. See
   `r001_session_id.py`, `r006_sse_transport.py`, `r007_deprecated_features.py`.
+- `project.search_wire(pattern, *, flags=0)` -- skips comments and
+  triple-quoted strings, but **keeps ordinary string literals**. Use this for
+  JSON-RPC method names and other wire text that only appears inside strings
+  (e.g. `tools/list`, `resources/subscribe`). Raw `search` over-matches prose
+  such as module docstrings describing the handshake; that produced real
+  false-positive `breaking` findings. See any of the ~16 rules that call
+  `search_wire`, and the docstring on `Project.search_wire` in
+  `src/mcp_migrate/rules/base.py`.
 - `project.search(pattern, *, flags=0)` -- yields `(SourceFile, lineno, stripped_line)`
   for every line matching a regex, including matches inside comments and
-  strings. Only use this when the rule genuinely wants string *content* --
-  e.g. `r003_routing_headers.py` checking whether the literal header name
-  `Mcp-Method` appears anywhere in the file, or any JSON-RPC method name
-  with a slash in it (`tools/list`, `resources/subscribe`, ...), which can
-  only ever appear inside a string token and so `search_code` would never
-  find it -- see the note in `r009_initialize_handshake_removed.py`.
+  strings. **No shipped rule uses this.** Reach for it only if you have a
+  reason neither `search_code` nor `search_wire` fits (for example, matching
+  free-form prose you intentionally want). Prefer `search_wire` for method
+  names and other string-literal wire content.
 - `project.imports()` -- returns the flat `set[str]` of every module name
   imported anywhere in the project (from `ast.Import` / `ast.ImportFrom`).
   Use this to gate a rule on a library actually being in use, the way
@@ -114,7 +124,7 @@ class LegacyExperimentalCapability(Rule):
 
     def check(self, project: Project) -> list[Finding]:
         out: list[Finding] = []
-        for f, line, text in project.search(r"[\"']experimental[\"']\s*:"):
+        for f, line, text in project.search_code(r"[\"']experimental[\"']\s*:"):
             out.append(self.finding(
                 "`experimental` capabilities key found; 2026-07-28 clients "
                 "look for `extensions` instead.",
@@ -129,8 +139,9 @@ Put one file that should trigger the rule and, ideally, one that shouldn't
 under `tests/fixtures/<rule-id>/`, lowercase:
 
 ```
-tests/fixtures/r022/bad.py     # contains "experimental": {...}
-tests/fixtures/r022/good.py    # contains "extensions": {...}, no "experimental"
+tests/fixtures/r022/bad.py      # contains "experimental": {...} as real code
+tests/fixtures/r022/good.py     # contains "extensions": {...}, no "experimental"
+tests/fixtures/r022/comment.py  # mentions "experimental": only in a comment/docstring — must not match
 ```
 
 Then a test under `tests/` that loads that fixture directory as a project and
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,10 +83,13 @@ almost certainly neither (and no shipped rule uses raw `search`).
   `src/mcp_migrate/rules/base.py`.
 - `project.search(pattern, *, flags=0)` -- yields `(SourceFile, lineno, stripped_line)`
   for every line matching a regex, including matches inside comments and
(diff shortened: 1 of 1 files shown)


dheerajjha:
Welcome, and thank you — you turned this around in under an hour, and it's a faithful, well-judged edit. The quick-pick line at the top is better placed than I'd have managed; putting it *before* the bullet list means someone skimming gets the answer without reading three paragraphs. Adding `comment.py` to the fixture list is the part I care about most, because it teaches the negative-case habit at the moment someone is copying the template.

**One claim needs correcting, and the correction is more interesting than the original.**

> `project.search(...)` — **No shipped rule uses this.**

One does:

```
src/mcp_migrate/rules/r010_server_discover_missing.py:119
    if any(project.search(r"server/discover")):
```

Counting call sites rather than files: `search_code` 37, `search_wire` 25, `search` 1.

That single use is a **suppression check** — it's asking "does this project already have `server/discover`, so I should stay quiet?" — not a detection. And it's the reason I've just filed #113, because it's broken in an interesting way: a comment reading `# TODO: server/discover is not implemented yet` satisfies the check and silences R010 entirely.

```
R010 findings: 1   <- no mention of discover at all
R010 findings: 0   <- plus a comment saying it is NOT implemented
```

Meanwhile `_ts_has_discover`, thirty lines below in the same file, uses `search_wire` and carries a comment explaining that letting prose satisfy the check "would suppress the finding for the whole project -- the worst failure mode available here".

So the accurate line is something like: *one rule uses raw `search`, as a suppression check, and it's a known bug (#113) — the TypeScript equivalent uses `search_wire` and documents why.* That's more useful to a contributor than "nobody uses it", because "nobody uses it" invites someone to grep, find the one call site, and assume the doc is out of date rather than that they've found a bug.

Everything else I checked holds: `search_wire` is used across 17 rule modules, the `search_wire` description matches its docstring, and the `R022` example is correct as `search_code` — I ran the old version and it produced findings on a comment, a docstring and a log line, which is exactly what you've fixed.

Adjust that one bullet and I'll merge it. And if you want #113 itself, it's yours — you're already the person who's read this corner of the code most recently, and it's a good second issue: small, verified, with the correct answer already written down in a comment thirty lines away from the bug.

dheerajjha:
Correction to my own review, and it matters for how that reads: **you got "No shipped rule uses this" from me.** #111 says `project.search  0 rule files` and "Zero rules use raw `project.search()`", and the task list told you to write "no shipped rule uses it". You implemented the issue faithfully; the wrong number was mine.

I counted rule *files* with a grep that mis-escaped the paren, then re-counted call sites while reviewing your PR and got 1. I've corrected #111.

Nothing changes about what the line should say — one rule uses it, as a suppression check, and it's #113 — but you shouldn't have to carry the framing that you asserted something unverified. You didn't.

dheerajjha:
@AshSgDe29071999 — gentle nudge, this is still one line from merging and I don't want it to go stale.

The only change needed is the `project.search` bullet, which still reads **"No shipped rule uses this."** As I noted above, that wrong number came from my issue, not from you — but it's wrong either way:

```
call sites in rules/:  search_code 37   search_wire 25   search 1
```

Suggested replacement for that sentence:

> **One rule uses it** — `r010_server_discover_missing.py` — as a suppression check, and that use is itself a bug (#113). Reach for it only if you have a reason neither `search_code` nor `search_wire` fits.

Everything else in the PR is already right and I've verified it. Push that and it goes in.

If you'd rather not, say so and I'll make the edit myself and merge it with your authorship intact — I'd just rather it be your commit.

dheerajjha:
**Merged.** And to be transparent about how: I pushed one commit to your branch rather than waiting longer, having offered that above. It changes exactly one sentence — the `project.search` bullet — and nothing else. Everything else in the PR is your work, unchanged, and the merge credits you.

What that sentence now says:

> **Exactly one shipped rule uses it** — `r010_server_discover_missing.py`, as a suppression check — and that use is itself a bug (#113), because a comment saying `server/discover` is missing convinces the rule that it isn't.

The reason for naming R010 rather than saying "nobody uses it": the next person to read that line will grep, find the one call site, and conclude the doc is stale rather than that they've found a bug. Pointing at it turns a contradiction into a lead.

Worth repeating clearly, since the review thread reads worse than the facts: **the wrong claim was mine.** #111 told you to write "no shipped rule uses it" and gave you a count of zero that I'd measured with a bad grep. You implemented the issue accurately.

**You're the eleventh contributor to this project**, and this closes #111.

If you want a second one, #113 is the bug your PR just documented, it's unassigned, and you've now read the exact code it lives in:

> A comment saying `server/discover` is missing convinces R010 that it isn't.

Three lines to fix, and the correct answer is already written in a comment thirty lines below the bug — `_ts_has_discover` uses `search_wire` and explains why prose must not satisfy the check. Mostly it needs the fixtures. Say the word and it's assigned.

dheerajjha:
@AshSgDe29071999 — a heads-up, since I pointed you at #113 yesterday and shouldn't let you find this by opening a PR against it.

**#113 is fixed and closed.** @waterlemonnn opened #115 overnight and it landed this morning. You hadn't claimed it and I hadn't assigned it to you, so it was open work — but I'd named it to you specifically, and it would be poor form to let you discover that by wasting an evening on it.

If you do want something else here, **#89**, **#91** and **#92** are open and genuinely unclaimed — I checked the open PR bodies for `Closes #N`, not just the assignee field. #92 in particular is closer to the documentation work you did on #112: re-verifying the fourteen published grades on the registry board against the false positives we've since confirmed. Say which one and I'll assign it to you, so this doesn't happen twice.

PR #118, opened 2026-08-06
fix(fixers): derive the comment prefix from the file, not the fixer

### LINKED ISSUE #117: fix --write corrupts TypeScript: five fixers emit Python `#` comments
`mcp-migrate fix --write` writes Python `#` comments into TypeScript files, which is a syntax error. The file it "fixed" no longer parses.

This is the most damaging bug currently in the repo: every other known issue produces a wrong *report*, which a human reads and discards. This one edits the user's source and leaves it broken.

## Repro

`server.ts`:

```ts
export function register(server: Server) {
  const sessionId = req.headers["Mcp-Session-Id"];
  const m = "tasks/list";
  return { sessionId, m };
}
```

```
mcp-migrate fix . --write
```

Result on `main` (`83df1b1`):

```ts
export function register(server: Server) {
  # TODO(mcp-migrate): replaced by an explicit handle argument, see https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567
  # const sessionId = req.headers["Mcp-Session-Id"];
  # TODO(mcp-migrate): tasks/list and blocking tasks/result are removed; poll with tasks/get + tasks/update ...
  # const m = "tasks/list";
  return { sessionId, m };
}
```

Four `#` lines. `#` does not start a comment in TypeScript. **This file is now unparseable**, and the tool reported `No findings remain.` / `Changes written.` on its way out.

## Which fixers

Only `r020_dynamic_client_registration_deprecated.py` gets this right, because #103 taught it to:

```python
comment_prefix = "// " if path.suffix in (".ts", ".js", ".tsx", ".jsx") else "# "
```

Every other comment-emitting fixer hardcodes `#`:

| Fixer | Hardcodes `#` | Rule reads TS? | Status |
|---|---|---|---|
| `r001_session_id.py` | yes | yes | **confirmed corrupting** (repro above) |
| `r019_tasks_polling.py` | yes | yes | **confirmed corrupting** (repro above) |
| `r006_sse_transport.py` | yes | yes | latent — didn't fire on my fixture, but the same hardcoded prefix |
| `r007_deprecated_features.py` | yes | yes | latent, same |
| `r014_sse_resumability.py` | yes | not yet | latent, and **#60 is open to port R014 to TypeScript** — that PR turns this one live |

`r004`, `r005`, `r017`, `r021` don't emit comments and are unaffected.

## The fix

Lift R020's suffix check into one shared helper rather than copy-pasting it five times — something like `comment_prefix(path)` in `fixers/base.py` — and route every emitting fixer through it. A copy-pasted ternary in six files is how this drifts again the moment someone adds a seventh fixer.

## Two things not to get wrong

**1. Commenting out is not just about the marker.** In the repro, `const sessionId = ...` is commented out while `return { sessionId, m }` two lines down still references it. Even with `//` that file is broken — now at type-check time instead of parse time. `_safe_to_comment_out` in `r019_tasks_polling.py` also reasons in Python:

```python
if stripped.endswith(":"):   # Python block opener
    return False
```

A TypeScript line ending in `{` is the equivalent, and `:` in TS is a type annotation — a perfectly ordinary thing to end a line with. So the guard is both wrong and inverted for TS. **Scope this issue to the comment marker if you like**, but say so in the PR, and file the block-safety half separately rather than leaving it implied.

**2. Don't fix this by making fixers refuse TypeScript.** That would trade a loud bug for a silent gap, and `review`-confidence fixers exist precisely to leave a human a note. The marker is the bug.

## Good first issue?

Partly. The marker fix is mechanical and #103 is a worked example to copy. The `_safe_to_comment_out` half is genuinely harder and needs a view on TS syntax. Take either; just be explicit about which.

Verified on `main` at `83df1b1`, 317 tests green — this is not caught by any existing test, which is its own finding: **no test asserts that a fixer's output is still valid in the language it edited.**

Vicky-Jha:
Fixes #117.

`fix --write` wrote Python `#` comments into TypeScript, where `#` opens nothing at all. The "fixed" file stopped parsing, and the tool printed `Changes written.` on its way out.

That asymmetry is why this one seemed worth doing carefully rather than patching five string literals: every other known issue in this repo produces a wrong **report**, which a human reads and discards. This one edits their source and leaves it broken.

## The fix

The prefix is derived once, from the path, in `fixers/base.py`:

```python
comment_prefix(Path("server.ts"))  # -> "// "
comment_prefix(Path("server.py"))  # -> "# "
```

…rather than remembered correctly by each of six fixers independently.

| Fixer | Before | After |
|---|---|---|
| `r001_session_id` | hardcoded `# ` | `comment_prefix(path)` |
| `r006_sse_transport` | hardcoded `# ` | `comment_prefix(path)` |
| `r007_deprecated_features` | hardcoded `# ` | `comment_prefix(path)` |
| `r014_sse_resumability` | hardcoded `# ` | `comment_prefix(path)` |
| `r019_tasks_polling` | hardcoded `# ` | `comment_prefix(path)` |
| `r020_dynamic_client_registration` | correct, own inline list | shared helper |

R020 already got this right after #103, but with its own inline suffix list. I moved it onto the shared helper as well — one place to be wrong beats two places that have to stay in agreement.

## Two bugs that fell out of centralising

**Missing module suffixes.** R020's inline list was `(".ts", ".js", ".tsx", ".jsx")`. The scanner also reads `.mts`, `.cts`, `.mjs` and `.cjs` (`languages.EXTENSIONS`), so those were corrupting silently even in the one fixer thought to be correct.

**Idempotency on TypeScript.** `already_commented` only recognised `#`. On a `.ts` file, the `//` written by the first run didn't read as commented, so a second run commented the comment. There's a test pinning that now.

An unknown suffix still falls back to `# `, so forgetting to add one degrades to today's behaviour rather than to something new.

## Verification

The issue's exact repro, on this branch:

```ts
export function register(server: Server) {
  // TODO(mcp-migrate): replaced by an explicit handle argument, see https://github.com/...
  // TODO(mcp-migrate): tasks/list and blocking tasks/result are removed; poll with tasks/get...
  // const m = "tasks/list";
  return { sessionId, m };
}
```

Valid TypeScript.

## Tests

The one worth looking at is `test_no_fixer_writes_a_python_comment_into_typescript`: it walks **every** fixer from `all_fixers()`, runs it over a TypeScript trigger, and asserts nothing emits a `#` line. A newly-contributed fixer that hardcodes `# ` fails there without anyone remembering to add it to a list — which matters, since this bug reached five fixers precisely because each was written in isolation.

Plus: per-language prefix assertions, all eight C-family suffixes parametrised, the unknown-suffix fallback, TypeScript idempotency, and the repro end to end.

Full suite: **331 passed**.

Also drops the `README` warning added by #117, since it no longer applies. Happy to split that into its own commit if you'd rather it landed separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,15 +166,6 @@ explicitly; it's implied when you pass neither flag) -- it prints the exact
 unified diff for every file that would change and writes nothing. Pass
 `--write` to apply it. `--dry-run` and `--write` are mutually exclusive.
 
-> [!WARNING]
-> **Do not run `fix --write` on TypeScript yet.**
-> ([#117](https://github.com/dheerajjha/mcp-migrate/issues/117)) Five fixers
-> emit Python `#` comments regardless of the file they are editing, so on a
-> `.ts` file they write lines that are a syntax error. The file will no
-> longer parse, and the tool reports success on its way out. This affects
-> v0.1.3. `check` is unaffected -- it only reads. On TypeScript, use `fix`
-> without `--write` and apply the diff yourself.
-
 Every change is tagged `safe` or `review` in the diff output:
 
 - **`safe`** -- the fixer is certain the transformation can't change
--- a/src/mcp_migrate/fixers/base.py
+++ b/src/mcp_migrate/fixers/base.py
@@ -23,6 +23,48 @@ from pathlib import Path
 
 CONFIDENCES = ("safe", "review")
 
+# Suffixes whose line comment is `//` rather than `#`.
+#
+# Kept in sync with the C-family entries of `languages.EXTENSIONS` -- and
+# deliberately including the `.mts`/`.cts`/`.mjs`/`.cjs` module variants,
+# which the scanner already reads and which an earlier hardcoded list here
+# missed. A suffix absent from this set falls back to `#`, so the failure
+# mode of forgetting one is a `#` in a C-family file: exactly the bug this
+# module exists to prevent. Add the suffix rather than assuming nobody
+# writes it.
+SLASH_COMMENT_SUFFIXES = frozenset({
+    ".ts", ".tsx", ".mts", ".cts",
+    ".js", ".jsx", ".mjs", ".cjs",
+})
+
+# Every prefix a line could already be commented with, in any language a
+# fixer might touch. Used to decide "have we (or a human) already commented
+# this out?" -- and it has to accept *both* families regardless of the file
+# in hand, because a `#` in a TypeScript file is precisely the corruption
+# being repaired here, and re-commenting it would just deepen the damage.
+COMMENT_PREFIXES = ("#", "//")
+
+
+def comment_prefix(path: Path) -> str:
+    """The line-comment prefix to write into `path`, including its space.
+
+    A fixer that hardcodes `# ` writes a Python comment into TypeScript,
+    where `#` starts nothing at all -- the "fixed" file stops parsing, and
+    the tool reports success on its way out. That is strictly worse than
+    the finding it was repairing: every other failure mode in this project
+    produces a wrong *report*, which a human reads and discards. This one
+    edits their source and leaves it broken.
+
+    So the prefix is derived from the file, once, here -- rather than
+    remembered correctly by each of six fixers independently.
+    """
+    return "// " if path.suffix.lower() in SLASH_COMMENT_SUFFIXES else "# "
+
+
+def is_commented(line: str) -> bool:
+    """True if `line`'s first non-whitespace token opens a line comment."""
+    return line.lstrip(" \t").startswith(COMMENT_PREFIXES)
+
 
 @dataclass
 class FixResult:
--- a/src/mcp_migrate/fixers/r001_session_id.py
+++ b/src/mcp_migrate/fixers/r001_session_id.py
@@ -14,10 +14,10 @@ from __future__ import annotations
 import re
 from pathlib import Path
 
-from .base import Fixer, FixResult
+from .base import Fixer, FixResult, comment_prefix, is_commented
 
 SPEC_URL = "https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567"
-TODO = "# TODO(mcp-migrate): replaced by an explicit handle argument, see " + SPEC_URL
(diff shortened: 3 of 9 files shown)


dheerajjha:
Merged, and this is the fix I'd have written if I'd written it well. `main` is at 377 tests.

Verified rather than trusted — before, on `main`:

```ts
  # TODO(mcp-migrate): replaced by an explicit handle argument, see ...
  # const sessionId = req.headers["Mcp-Session-Id"];
```

after:

```ts
```

Zero `#` lines in either a `.ts` or a `.mts` file.

**Three things you did that the issue didn't ask for, all of them right.**

**You found a bug in the thing I called the worked example.** I pointed at `r020`'s ternary as the pattern to copy. Its suffix list is `(".ts", ".js", ".tsx", ".jsx")` — and `languages.EXTENSIONS` reads `.mts`, `.cts`, `.mjs` and `.cjs` too. So the one fixer I described as correct was also wrong, for four suffixes, and copying it five times would have propagated that. I checked `EXTENSIONS` after reading your comment and you're right.

**`is_commented` accepts both prefixes regardless of the file.** That is subtle and it matters: a `#` sitting in a TypeScript file is precisely the damage being repaired, and a fixer that only recognised `//` there would re-comment it and deepen it. Nothing in the issue mentioned re-running the fixer over already-corrupted files.

**You scoped it.** I asked for the marker half and said to leave `_safe_to_comment_out` alone and file the block-safety half separately. You left it alone — `ends_as_block_opener` and the `endswith(":")` guard are untouched. That's still an open problem (a TS line ending in `:` is a type annotation, not a block opener) and it's still worth filing, but it isn't tangled into this.

The docstring on `comment_prefix` is doing real work too — it explains *why* this class of bug outranks the others, which is the thing that stops someone helpfully reverting it later.

You also removed the README warning I added this morning, which was correct — it was true for about forty minutes.

Welcome. If you want another: **#123** is a one-liner I filed twenty minutes ago (R017 compiles with `re.IGNORECASE` and throws the flag away), and the `_safe_to_comment_out` block-safety problem above is unfiled and yours if you want to write it up.

dheerajjha:

```

1
```

@Vicky-Jha, you're named in the release notes and in `CHANGELOG.md`. **Every fix in this release came from outside the repo**, which is worth saying plainly.

Issue #85, opened 2026-08-04
No machine-readable output: add `check --format json`


dheerajjha:
`mcp-migrate check` prints a table for a human. There is no way to get the findings out as data.

That blocks the two things people actually want to do with a linter:

- **Fail a build on a specific rule** rather than on the process exit code alone.
- **Diff two runs** — "did this PR introduce a breaking finding?" — which is the whole
  value proposition for a migration tool. You care about the delta, not the absolute.

Right now the only integration point is exit code 0/1/2, which cannot answer "which rule,
which file, which line".

### Shape

```console
$ mcp-migrate check ./server --format json
```

```json
{
  "tool": "mcp-migrate",
  "version": "0.1.2",
  "spec": "2026-07-28",
  "root": "server",
  "grade": "C",
  "counts": { "breaking": 2, "deprecated": 1, "advisory": 0 },
  "findings": [
    {
      "rule": "R001",
      "severity": "breaking",
      "path": "src/transport.ts",
      "line": 64,
      "message": "Mcp-Session-Id was removed from the Streamable HTTP transport.",
      "fix": "Drop the session header; the 2026-07-28 transport is stateless."
    }
  ]
}
```

### Constraints that matter

- **stdout must be valid JSON and nothing else** when `--format json` is passed. No rich
  output, no progress spinner, no banner. Anything else and `| jq` breaks, which defeats
  the point. Send warnings to stderr.
- **Paths relative to the scan root**, as they already are internally, so output is stable
  across machines and diffable in CI.
- Exit codes stay exactly as they are (0 clean / 1 breaking / 2 could not check). The flag
  changes the format, not the semantics.
- A finding with no fix text should omit the key rather than emit `null`.

### How to do it

`Finding` is already a dataclass, so most of this is a serialiser plus a branch in the CLI.
Look at how `check` builds its output today and add the format switch as high up as you
can — the temptation is to thread a flag through the renderer, and it reads much better as
two renderers behind one interface.

### Done when

`--format json` emits parseable JSON on a project with findings and on a clean one, exit
codes are unchanged, and a test asserts `json.loads(stdout)` succeeds in both cases.

Good one to pick up if you want something with a clear contract and no regex judgment
calls. Ask in the thread and I will answer the same day.

PR #119, opened 2026-08-06
fix(cli): let a breaking finding fail the build on partially-covered languages

### LINKED ISSUE #98: Every TypeScript project exits 2, so CI cannot fail on a breaking finding
Exit codes are documented as **0 clean · 1 breaking findings · 2 could not check**. That holds for Python. For TypeScript, everything returns **2**.

Measured on `main` today:

| Project | Expected | Actual |
| --- | --- | --- |
| Python, clean | 0 | **0** ✓ |
| Python, breaking finding | 1 | **1** ✓ |
| TypeScript, clean | 0 | **2** ✗ |
| **TypeScript, breaking finding** | **1** | **2** ✗ |
| Empty directory | 2 | 2 ✓ |
| Path does not exist | 2 | 2 ✓ |

So a TypeScript server with a real `breaking` finding is indistinguishable — to a shell, to CI, to anything — from a typo'd path.

### Cause

[`cli.py:201`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/cli.py). The ungraded branch returns unconditionally:

```python
if reason:
    ...
    if findings:
        # prints them, correctly
    return EXIT_UNSCANNABLE          # <- even when findings is non-empty
```

TypeScript is `PARTIAL`, so `reason` is always set for a TS project, so that `return` always wins. The code immediately above it prints the findings and says *"Real, and worth fixing"* — and then the exit code says nothing was checked.

### Why this matters more than it reads

Most MCP servers are TypeScript. The whole point of a migration linter is to fail a build when a breaking change lands, and **no TypeScript user can currently do that.** They cannot even distinguish "clean" from "broken invocation", so there's no workaround short of parsing stdout.

It also makes #85 (`--format json`) less useful than it should be: even with parseable output, the exit code stays wrong.

### The fix, and the judgment call in it

Grading and checking are different questions, and right now one answer is doing both jobs. Withholding the *grade* on partial coverage is correct and should not change — that's a deliberate, well-argued decision and the printed explanation is good. But *findings* are findings regardless of whether we'll stand behind a letter.

```python
if reason:
    ...
    return 1 if any(rules[f.rule_id].severity == "breaking" for f in findings) else 0
```

...with `EXIT_UNSCANNABLE` still returned when there is genuinely nothing to read.

The thing to decide, and worth saying in the thread before you code: **should a clean-but-ungraded project exit 0?** Argument for: nothing breaking was found, which is what exit 0 means. Argument against: 0 reads as "verified clean" and we only ran 16 of 21 rules. I lean toward 0 with the existing on-screen caveat doing the honest work, but I hold it loosely and would rather hear the counter-argument first.

### Done when

A TypeScript project with a breaking finding exits 1; a clean one exits whatever we decide; an empty directory still exits 2; Python behaviour is unchanged. Tests for each row of the table above.

iphonekumar:
Fixes #98.

The ungraded branch returned `EXIT_UNSCANNABLE` unconditionally, so a TypeScript server with a real `breaking` finding was indistinguishable — to a shell, to CI, to anything — from a typo'd path.

| Project | Documented | Before | After |
|---|---|---|---|
| TypeScript, clean | 0 | 2 ✗ | **0** ✓ |
| TypeScript, breaking | 1 | 2 ✗ | **1** ✓ |
| Empty directory | 2 | 2 ✓ | 2 ✓ |
| Language with no backend | 2 | 2 ✓ | 2 ✓ |
| Path does not exist | 2 | 2 ✓ | 2 ✓ |

## What changed, and what deliberately didn't

Grading and checking were one answer doing two jobs.

**The grade is still withheld.** `grade: null`, `score: null`, `scannable: false`, the on-screen explanation — all unchanged, with a test pinning it. That decision is well-argued and this PR doesn't touch it.

**The exit code now carries findings.** They're findings whether or not we'll stand behind a letter.

The discriminator is `project.files` — did we open anything?

```python
def _checked_something(project) -> bool:
    return bool(project.files)
```

A TypeScript tree is `PARTIAL`, but we *did* read those files and run every rule that covers them. An empty directory or a language with no backend read nothing, and 2 remains the honest answer there.

## The open question in the issue

You asked whether a clean-but-ungraded project should exit 0, and said you held it loosely.

I've taken your lean, for the reason you gave: 0 means "nothing breaking was found", which is exactly what happened, and the printed caveat does the honest work of saying only 16 of 21 rules ran. The counter-argument — that 0 reads as "verified clean" — is real, but it applies equally to a Python project where a rule simply has no pattern for some construct, and we don't hedge the exit code there.

It's one branch in `_checked_something`, so flipping it to `EXIT_UNSCANNABLE`-when-clean is a one-line change if you decide the other way.

## One knock-on

The console printed `Nothing scannable here.` whenever there were no findings. Paired with exit 0 on a clean TypeScript tree, that's self-contradictory — we read those files. It now says that only when nothing was actually opened.

## About the three changed tests

`test_check_refuses_and_offers_no_grade`, `test_check_json_reports_no_grade` and `test_typescript_only_tree_reports_findings_without_a_grade` asserted the old exit code. They're **updated, not deleted** — every no-grade and no-badge assertion is intact, since that half is deliberately unchanged. What changed is that the exit code is now parametrised across the two cases, because an empty tree and a clean TypeScript tree differ in exactly that one respect and agree on everything else.

Flagging it explicitly since "PR changes existing tests" is worth a second look on principle.

## New tests

Breaking TS exits 1 · clean TS exits 0 · empty tree still 2 · no-backend language still 2 · the grade is still withheld · and the JSON and console paths agree, since they're two returns of one contract and they drifted apart once already.

Full suite: **324 passed**.

---

This also unblocks the exit-code half of #85 (`--format json`) — parseable output isn't much use while the exit code says nothing was checked.

### DIFF
--- a/src/mcp_migrate/cli.py
+++ b/src/mcp_migrate/cli.py
@@ -114,6 +114,26 @@ def _print_language_hint(console, counts) -> None:
         )
 
 
+def _exit_for(findings, rules) -> int:
+    """Exit code from findings alone: 1 if anything breaking, else 0."""
+    return EXIT_FINDINGS if any(
+        rules[f.rule_id].severity == "breaking" for f in findings
+    ) else EXIT_OK
+
+
+def _checked_something(project) -> bool:
+    """Did we actually read source, or is this an empty/unreadable tree?
+
+    The discriminator between "partial coverage" and "could not check".
+    A TypeScript project is `PARTIAL` -- we withhold the *grade*, which is
+    correct and deliberate, but we did open files and run rules against
+    them, so the exit code has real information to carry. An empty
+    directory or a tree in a language with no backend read nothing, and
+    EXIT_UNSCANNABLE is the honest answer there.
+    """
+    return bool(project.files)
+
+
 def cmd_check(args) -> int:
     console = Console()
     root = Path(args.path).resolve()
@@ -169,7 +189,8 @@ def cmd_check(args) -> int:
                     "location": f.location(),
                 } for f in findings],
             }, indent=2))
-            return EXIT_UNSCANNABLE
+            return _exit_for(findings, rules) if _checked_something(project) \
+                else EXIT_UNSCANNABLE
         print(json.dumps({
             "spec": SPEC,
             "path": str(root),
@@ -185,9 +206,7 @@ def cmd_check(args) -> int:
                 "location": f.location(),
             } for f in findings],
         }, indent=2))
-        return EXIT_FINDINGS if any(
-            rules[f.rule_id].severity == "breaking" for f in findings
-        ) else EXIT_OK
+        return _exit_for(findings, rules)
 
     console.print()
     console.print(f"[bold]mcp-migrate[/bold] [dim]v{__version__}[/dim]  ->  {root.name}")
@@ -235,7 +254,14 @@ def cmd_check(args) -> int:
         console.print()
         # Not .capitalize() -- that lowercases the rest, turning "24
         # TypeScript" into "24 typescript".
-        headline = "No grade for this one." if findings else "Nothing scannable here."
+        # "Nothing scannable" is only true when we opened nothing. A clean
+        # TypeScript tree now exits 0, and pairing that with "nothing
+        # scannable" would contradict it -- we did read those files and run
+        # every rule that covers them; what we withheld is the grade.
+        headline = (
+            "Nothing scannable here." if not _checked_something(project)
+            else "No grade for this one."
+        )
         console.print(
             f"[bold yellow]{headline}[/bold yellow] {reason[0].upper()}{reason[1:]}."
         )
@@ -258,7 +284,8 @@ def cmd_check(args) -> int:
             "not read.[/dim]"
         )
         console.print()
-        return EXIT_UNSCANNABLE
+        return _exit_for(findings, rules) if _checked_something(project) \
+            else EXIT_UNSCANNABLE
 
     n_python = sum(1 for f in project.files if f.language == "python")
     console.print(f"[dim]{n_python} Python files, {len(rules)} rules, spec {SPEC}[/dim]")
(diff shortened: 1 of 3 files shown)


dheerajjha:
Merged. This unblocks the thing v0.1.3 shipped without: **CI can now fail on a TypeScript regression.**

I verified the whole matrix rather than the two cases in the tests:

```
  empty dir        -> exit 2
  TS + breaking    -> exit 1
  TS clean         -> exit 0
  Python clean     -> exit 0
  Go only          -> exit 2
```

All five right. `_checked_something(project)` as the discriminator is the correct call — the question was never "is this language supported", it's "did we open any files", and those come apart exactly on the partial languages.

Fixing the `"Nothing scannable here."` headline in the same PR was the right instinct. A clean TypeScript tree exiting 0 under a banner saying nothing was scanned would have been a worse contradiction than the bug.

**Two things left behind, neither blocking, both mine to have caught earlier.**

**1.** I pushed a follow-up for the closing line, which you didn't touch:

> No grade and no badge: this tool has no opinion about code it could **not read**.

It printed that directly under the findings it had just read. Same class of contradiction as the headline you fixed, three lines further down. It now says the grade was withheld when files were read, and keeps the old wording for genuinely unreadable trees.

**2. The JSON is now internally inconsistent, and I'd like your view before anyone changes it.** On a TypeScript tree with a breaking finding:

```json
{ "scannable": false, "findings": [ ... ] }
```
exit code `1`.

A CI script reading `scannable: false` concludes nothing was checked, while the exit code says otherwise. The field predates you and your change is what makes the incoherence reachable — the honest fix is probably a separate `gradable` field, with `scannable` meaning what its name says. That's an interface change though, and you've just done the thinking on this area. **Want to file it?** I'd rather it came from you than from me.

**One process note that's about me, not you.** You checked #98 was unassigned before starting, which was correct. It was also *wrong* — I'd told @Shxnque on 2026-08-05 that #98 was theirs and then never ran the assignment, so it looked free because of my mistake. You did nothing wrong and I've said so on #88. Mentioning it only so it doesn't surface later as though you'd jumped a queue.

Welcome, and thank you — this was the top item on my list and I'd assumed it would sit another week.

dheerajjha:

```

1
```

@iphonekumar, you're named in the release notes and in `CHANGELOG.md`. **Every fix in this release came from outside the repo**, which is worth saying plainly.

PR #120, opened 2026-08-06
test: guard against the #67 quadratic scan — and fix the live one it found in R003

### LINKED ISSUE #86: Nothing stops the #67 quadratic scan from coming back
#67 was a real O(n²): R015's TypeScript path re-ran two `search_wire` scans of the whole
project inside its per-file loop. @li2631026381-alt hoisted them out in #70 and the
measured effect was **63.0 ms → 3.1 ms at 200 files, a 20× speedup.**

Nothing stops that from coming back. The rule interface makes it genuinely easy to do by
accident — `project.search_*` looks like a cheap accessor at the call site, and inside a
`for f in project.files` loop it is quadratic. We now have 21 rules and 14 of them have a
second language path, so there is a lot of surface for someone to reintroduce it in good
faith.

### What to build

A test that generates a synthetic project of N TypeScript files, runs every rule, and
asserts the whole scan stays under a wall-clock budget.

```python
def test_full_scan_stays_linear_ish(tmp_path):
    # 200 files is enough that a quadratic rule blows the budget by ~20x
    # while a linear one finishes in single-digit ms.
    ...
```

### The part that needs judgment

**A naive timing test is a flaky test**, and a flaky test in CI is worse than no test —
people learn to re-run it, and then it catches nothing. Options, and I do not have a
strong view:

- Assert a generous absolute ceiling (say 2 s for 200 files). Simple, but tied to runner
  speed, and our CI matrix is four Python versions.
- Measure the *ratio* between N=50 and N=200. Scale-free and mostly immune to a slow
  runner, but noisier on small numbers.
- Count `search_*` invocations with a counter rather than timing anything. Deterministic
  and fast, but tests the proxy rather than the thing.

I lean toward counting calls — it is the only one that cannot flake — but it needs the
counter to live somewhere that is not ugly. Say which you would pick in the thread.

### Done when

Reintroducing the #67 pattern in any rule makes the suite fail, and the test does not flake
across ten consecutive runs on the CI matrix.

Worth knowing: this is the kind of test that pays for itself once and then silently forever.
Ask in the thread if you want to talk through the approach first.

s35153:
Fixes #86.

## Which approach, and why

You floated three and said you didn't have a strong view. This counts `search_*` calls, taking your lean, for the reason you gave: **it's the only one that cannot flake.** A flaky perf test in CI is worse than no test — people learn to re-run it, and then it catches nothing. The matrix is four Python versions on shared runners, which is precisely where a wall-clock budget goes bad.

Your reservation was that counting "tests the proxy rather than the thing." I'd argue it's exact here, not a proxy:

`run_check` invokes each rule once per language, and every `search_*` already iterates all files internally. So a correctly-written rule issues **the same number of calls at 20 files as at 200** — the count is a constant of the rule, not a function of project size. A rule scanning per-file issues one call per file. That's a difference in kind, not degree, so the assertion needs no tolerance, no ratio, and no clock.

The counter lives in the test file, not on `Project` — putting bookkeeping in the hot path of the thing being measured, to serve a test, seemed like the wrong trade.

## It found a live one

R003's TypeScript path ran a whole-project `search_wire` **inside** `for f in project.files`, then discarded every match not belonging to the current file:

```python
for f in project.files:
    ...
    calls = [
        (line, text)
        for file_obj, line, text in project.search_wire(TS_HTTP_CALL)  # <- every file, every time
        if file_obj.path == f.path
    ]
```

Same shape as #67, still in the tree. Hoisted into one pass grouped by path:

| 200 TypeScript files, R003 TS path | |
|---|---|
| before | **1433.9 ms** |
| after | **37.6 ms** |
| | **38×** |

I fixed it here rather than filing it separately, because a guard that can't be switched on guards nothing — and the fix is the same hoist #70 applied to R015. Happy to split it into its own commit if you'd rather review them apart.

## Two bugs in the instrumentation, both now pinned

Writing this, the guard twice looked green while measuring nothing. Both failure modes have tests now, because a guard that silently stops guarding is worse than none:

- **Double-counting.** Re-reading the `Project` methods per call wrapped the previous wrapper, so counting twice in one test doubled the numbers. Originals are captured once at import.
- **Zero-counting.** The instrumented rule set and the executed rule set were built separately; when they diverged, every count came out zero and the guard passed over rules it never touched. They're one list now, used for both.

## Tests

- no rule's scan count grows with file count (the guard itself)
- a deliberately quadratic rule **is** caught — the guard's own regression test
- a correctly-hoisted rule is **not** flagged, so it can't just be asserting "everything is fine"
- counts are byte-identical across runs, which is the property that makes it CI-safe
- the synthetic fixture really trips several rules — a rule that returns early never reaches its per-file loop and would look linear however it was written

Full suite: **322 passed**.

Reintroducing the #67 pattern in any rule now fails the suite, deterministically, on any runner.

### DIFF
--- a/src/mcp_migrate/rules/r003_routing_headers.py
+++ b/src/mcp_migrate/rules/r003_routing_headers.py
@@ -160,6 +160,15 @@ class MissingRoutingHeaders(Rule):
         return out
 
     def _check_ts(self, project: Project) -> list[Finding]:
+        # Hoisted out of the per-file loop below. `search_wire` walks every
+        # file in the project on each call, so issuing it once per file --
+        # and then discarding all but the current file's matches -- is
+        # O(files**2). That is the #67 shape; see tests/test_scan_complexity.py,
+        # which now fails if it comes back here or anywhere else.
+        calls_by_path: dict = {}
+        for file_obj, line, text in project.search_wire(TS_HTTP_CALL):
+            calls_by_path.setdefault(file_obj.path, []).append((line, text))
+
         out: list[Finding] = []
         for f in project.files:
             if not TS_MCP_SURFACE_RX.search(f.text):
@@ -171,11 +180,7 @@ class MissingRoutingHeaders(Rule):
             if not missing_method and not missing_name:
                 continue
 
-            calls = [
-                (line, text)
-                for file_obj, line, text in project.search_wire(TS_HTTP_CALL)
-                if file_obj.path == f.path
-            ]
+            calls = calls_by_path.get(f.path)
             if not calls:
                 continue
 
--- a/tests/test_scan_complexity.py
+++ b/tests/test_scan_complexity.py
@@ -1,0 +1,247 @@
+"""A guard against #67 coming back.
+
+#67 was a real O(n**2): R015's TypeScript path re-ran two `search_wire`
+scans of the whole project inside its per-file loop. #70 hoisted them out
+and measured 63.0 ms -> 3.1 ms at 200 files, a 20x speedup.
+
+Nothing stopped it from returning. The rule interface makes it easy to do
+by accident -- `project.search_*` reads like a cheap accessor at the call
+site, and inside a `for f in project.files` loop it is quadratic. There
+are 21 rules and 14 of them carry a second language path, so there is a
+lot of surface for someone to reintroduce it in good faith.
+
+Of the three approaches floated in #86 -- an absolute time ceiling, a
+ratio between two sizes, or counting `search_*` calls -- this counts the
+calls, for the reason the issue gives: it is the only one that cannot
+flake, and a flaky perf test in CI is worse than no test because people
+learn to re-run it and then it catches nothing. The CI matrix is four
+Python versions on shared runners, which is exactly where a wall-clock
+budget goes bad.
+
+What makes counting *exact* rather than a proxy here: `run_check` invokes
+each rule once per language it declares, and every `search_*` method
+already iterates all files internally. So a rule that scans correctly
+issues **the same number of calls whether the project holds 20 files or
+200** -- the count is a constant of the rule, not a function of project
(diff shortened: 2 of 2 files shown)


dheerajjha:
Merged. This is the best-argued PR this repo has had, and I want to say why rather than just thank you.

**You picked the right instrument and you showed your work.** #86 floated three options: a time ceiling, a size ratio, or counting calls. You took counting, and the justification is exact rather than hand-wavy:

> `run_check` invokes each rule once per language it declares, and every `search_*` method already iterates all files internally. So a rule that scans correctly issues **the same number of calls whether the project holds 20 files or 200** — the count is a constant of the rule, not a function of project size.

That's what makes the assertion need no tolerance. A wall-clock budget across four Python versions on shared runners would have flaked within a week, and — your point, and the one I'd have missed — **a flaky perf test is worse than no perf test**, because people learn to re-run it and then it catches nothing.

**And it found a live one on its first run.** R003 was re-scanning the whole project inside its per-file loop, the same shape as #67, in a rule nobody was looking at. A regression guard that immediately catches a regression nobody knew about is the strongest possible argument for the guard.

Keeping the instrumentation test-only was right too. Putting a counter on `Project` would have added bookkeeping to the hot path of the exact thing being measured, to serve a test.

377 tests on `main`, green.

One thing worth knowing: the #67 fix this guards was @li2631026381-alt's, and they've gone quiet — their other two PRs I closed today as superseded or already-fixed. Their hoist is the reason R015 is fast, and your test is the reason it stays that way. I've said so on their thread.

If you want more: **#123** (R017 discards a compiled `re.IGNORECASE`) is small and verified, and **#91** is the meatiest correctness issue open.

dheerajjha:

```

1
```

@s35153, you're named in the release notes and in `CHANGELOG.md`. **Every fix in this release came from outside the repo**, which is worth saying plainly.

PR #121, opened 2026-08-06
feat(registry): generate live shields.io badge endpoints from the registry

### LINKED ISSUE #101: The badge is frozen at scan time and never goes stale, which makes it a lie
`mcp-migrate check` already prints a badge URL when a project grades cleanly:

```
Grade A. Nothing to fix. Add your badge:
![MCP 2026-07-28](https://img.shields.io/badge/MCP%202026--07--28-A-brightgreen)
```

That badge is **frozen at the moment you ran the command.** It keeps saying A after the spec moves, after you regress, after a new rule ships that you fail. A migration badge that cannot go stale is worse than no badge, because it's a claim that stops being true without anyone noticing.

### What to build

A [shields.io endpoint](https://shields.io/badges/endpoint-badge) — a JSON document at a stable URL that shields renders live:

```
https://img.shields.io/endpoint?url=https://<host>/badge/sooperset/mcp-atlassian
```

```json
{
  "schemaVersion": 1,
  "label": "MCP 2026-07-28",
  "message": "C",
  "color": "yellow"
}
```

We already have the data: [`registry/servers/*.yaml`](https://github.com/dheerajjha/mcp-migrate/tree/main/registry/servers) holds a grade per listed server, and it's regenerated when entries change. This is a transform from that, not new analysis.

### Why it's a good slot

Static hosting. No server, no database, no secrets. A script that reads `registry/servers/*.yaml` and writes `docs/badge/<owner>/<name>.json`, plus a workflow step that runs it on push to main and publishes to GitHub Pages. The whole thing is maybe 60 lines and a workflow edit.

### Details that matter

- **Colours must match the grades**: A/B `brightgreen`, C `yellow`, D/F `red`. If a badge says C in green, nobody trusts the next one.
- **Unknown repo → a real answer**, not a 404. `{"message": "not listed", "color": "lightgrey"}` renders; a 404 shows a broken image on someone's README, which is a bad thing to do to a stranger.
- **URL-safe paths.** Owner/name pairs contain characters that need care.
- **Say when it was checked.** `checked_with` and the scan date are in the YAML. A badge that renders live but reflects a scan from six weeks ago is only *slightly* better than the frozen one, so consider putting the date in the label or exposing a companion endpoint.

### Then update the CLI

Once endpoints exist, `check` should print the live URL for listed servers instead of the frozen one. Fine as a follow-up PR — say if you want to do both or just the generator.

### Done when

`registry/servers/mcp-atlassian.yaml` produces a JSON endpoint that shields renders as the grade in the right colour, an unlisted repo renders "not listed", and it regenerates automatically when the registry changes.

djubx:
Fixes #101.

`scripts/render_badges.py` transforms `registry/servers/*.yaml` into [shields.io endpoint](https://shields.io/badges/endpoint-badge) documents. The grade moves out of the image URL and into a JSON file we regenerate, so shields renders the current value at request time and every badge updates itself.

Static files, no server, no database, no secrets — wired into the existing Board workflow, which already reruns on registry changes.

```json
{
  "schemaVersion": 1,
  "label": "MCP 2026-07-28 (0.1.1)",
  "message": "C",
  "color": "yellow"
}
```

## The case the issue didn't anticipate: monorepos

Three registry entries share `modelcontextprotocol/servers`. Three share `awslabs/mcp`. **Keyed by repo alone, the last write wins** — the badge reports an arbitrary one of the three grades, with nothing on screen to say which.

I found this when the generator wrote 14 files and 10 landed.

So there are two addresses:

| Path | Reports |
|---|---|
| `badge/<name>.json` | one registry entry, exactly |
| `badge/<owner>/<repo>.json` | the repository as a whole |

`name` is unique by construction (it must match the filename), which makes it the precise address for a server inside a monorepo. The repo path is what people actually paste into a README, and where it covers several servers it reports the **worst** grade with the count in the label:

```json
{ "label": "MCP 2026-07-28 (0.1.1) (3 servers)", "message": "D", "color": "red" }
```

Worst rather than mean or best because this is a migration-readiness claim: *"something in here is an F"* is the fact a reader needs, and averaging it away would recreate exactly the quietly-untrue badge these endpoints exist to remove. Say if you'd rather the repo-level badge didn't exist at all and monorepos were addressed only per-server — that's a smaller surface, at the cost of the README case.

## The details from the issue

- **Colours match the grades.** A/B `brightgreen`, C `yellow`, D/F `red`, parametrised in tests.
- **Unknown repo renders.** `unknown.json` says `not listed` in grey. A 404 shows a broken image on a stranger's README, and anyone can point an endpoint URL at us.
- **URL-safe paths.** A `repo` that isn't a plain `owner/name` is *refused* rather than escaped — it's a registry error worth surfacing, and writing outside `docs/badge/` would be worse. Traversal attempts are tested.
- **Says when it was checked.** `checked_with` rides in the label. A badge that renders live but reflects a six-week-old scan is only slightly better than a frozen one, and the version is already in the YAML.

A malformed entry is warned about and skipped rather than raised on — `validate_registry.py` is where that gets rejected, and failing here would stop every *other* badge from refreshing, which is the bug being fixed.

## Tests (31)

Every grade's colour · the unlisted fallback · an out-of-range grade saying `unknown` rather than inventing a colour · monorepo aggregation · per-server addresses inside a monorepo · path traversal refused · one bad entry not blocking the rest · determinism (it commits in CI, so nondeterminism means endless no-op commits) · and **every generated grade matching its source YAML** — the badge is a transform, so if it ever disagrees, it's the badge that's wrong.

Full suite: **348 passed**.

## Scope

You asked whether to do both halves. This is the generator only. Not included, happy to follow up in a second PR:

- publishing `docs/` to GitHub Pages (needs a repo settings change on your side)
- switching the CLI's printed badge to the live URL for listed servers

The files are generated and committed but not yet served, so nothing renders until Pages is switched on.


dheerajjha:
Merged. The monorepo handling is the part that makes this correct rather than merely working.

I verified the worst-grade rule against the registry rather than trusting the tests:

```
awslabs/mcp                    -> A   (entries: A, A, A)
modelcontextprotocol/servers   -> D   (entries: B, D, B)
```

Right in both directions. And your reasoning for *worst* rather than *average* is the same argument the tool itself rests on:

> "something in here is an F" is the fact a reader needs, and averaging it away would recreate the quietly-untrue badge these endpoints exist to remove

A `not listed` badge instead of a 404 is a small thing that would have been a support burden forever — a broken image on someone else's README is the kind of bug that never gets reported and never gets fixed.

I also checked the generator is idempotent, since committed generated output that drifts from its source is a trap:

```
$ python scripts/render_badges.py && git status --short docs/badge
(clean)
```

348 tests at the time I checked; `main` is at 377 now with the rest of this morning's merges.

**One open item, which is mine, not yours:** you noted that publishing `docs/badge` to GitHub Pages is the remaining step and that until then the files are generated but not served. That's a repo-settings action I need to take — flagging it here so it's recorded and so nobody assumes the endpoints are live yet. I'll confirm on this thread when Pages is serving it.

This is your second merged PR here (R008's TypeScript port was the first). Thank you — and #101 had been sitting since before v0.1.3.

dheerajjha:
**Pages is enabled and your endpoints are live**, as promised on this thread.

Serving `docs/` from `main` at `https://dheerajjha.github.io/mcp-migrate/`. Verified end to end rather than just enabling it and hoping:

```
$ curl -s https://dheerajjha.github.io/mcp-migrate/badge/sooperset/mcp-atlassian.json
{ "schemaVersion": 1, "label": "MCP 2026-07-28 (0.1.1)", "message": "C", "color": "yellow", "isError": false }

$ curl -s https://dheerajjha.github.io/mcp-migrate/badge/modelcontextprotocol/servers.json
{ "schemaVersion": 1, "label": "MCP 2026-07-28 (0.1.1) (3 servers)", "message": "D", "color": "red" }

$ curl -s https://dheerajjha.github.io/mcp-migrate/badge/unknown.json
{ "schemaVersion": 1, "label": "MCP 2026-07-28", "message": "not listed", "color": "lightgrey" }
```

And shields.io actually renders it, which is the part that could have failed silently:

```
HTTP 200  type=image/svg+xml  1348 bytes
  >MCP 2026-07-28 (0.1.1)</text>
  >C</text>
```

MCP 2026-07-28

I pushed one follow-up commit (`aee5381`) replacing the `<pages-host>` placeholder in `docs/badge/README.md` and in `render_badges.py`'s docstring with the real host, and dropped the "not yet served" caveat you'd correctly written while it wasn't.

**This shipped in [v0.1.4](https://github.com/dheerajjha/mcp-migrate/releases/tag/v0.1.4), on PyPI as of a few minutes ago.** The monorepo worst-grade behaviour is called out in the release notes, because it's the design decision in here that someone will otherwise re-litigate later.

Thank you — #101 had been open since before v0.1.3.

PR #148, opened 2026-08-06
fix(r014): match TypeScript lastEventId case variants (#143)

### LINKED ISSUE #143: R014 TypeScript misses case variants: lastEventID, LastEventId
R014's TypeScript path (added in #127) matches the identifier form case-sensitively, so it finds `lastEventId` but misses `lastEventID`, `LastEventId`, and `LAST_EVENT_ID`.

```ts
// found
const lastEventId = req.headers.get("last-event-id");

// missed
transport.onresume = (lastEventID) => replay(lastEventID);
```

Verified on `main` (6688d16): the second line produces no R014 finding.

## Credit

**This was caught by @PuvaanRaaj's #60**, an independent R014 port that got there first on this specific point. I compared both against the same corpus rather than picking one on seniority, and they turned out complementary — #127 catches an object-literal header map (`{ "Last-Event-ID": cursor }`) that #60 misses, and #60 catches the case variants that #127 misses. #127 merged because #60 had been unmergeable and its author offline since 2026-08-04; the half #60 got right is this issue.

## The fix

`_check_ts` builds its identifier pattern without flags. Pass `re.IGNORECASE` through to `search_code`, the way #60 did:

```python
for f, line, text in project.search_code(TS_IDENT_RX, flags=re.IGNORECASE):
```

**Careful with the wire half.** The header-string search should stay as it is — `search_wire` on `last-event-id` already matches the header case-insensitively in practice because the pattern is written lowercase and headers are conventionally lowercased on the wire. Widening both halves risks a false positive on unrelated prose containing "last event".

## Watch for

This is the same shape as #123: a rule that *intends* case-insensitivity but never gets the flag to the search. `search_wire`/`search_code` recompile with `flags=0`, so a flag living on a compiled `re.compile(...)` object is silently dropped when only `.pattern` is passed. R014 is the second instance found. **Worth grepping the other 19 rules for the same trap while you're in here.**

## Test

`tests/test_typescript.py` — assert `lastEventID` and `lastEventId` both fire.

ujjwalprakash17:
Fixes #143.

R014's TypeScript identifier path matched case-sensitively, so it found
`lastEventId` and `LastEventId` but missed `lastEventID` and `LAST_EVENT_ID`.

- `TS_CODE_RX` is now a single bounded pattern, `\blast_?event_?id\b`,
  matched with `re.IGNORECASE` via `search_code(..., flags=...)`.
  The `_?` covers the SCREAMING_SNAKE const form, which the flag alone
  would still miss.
- The header/wire half is unchanged and stays case-sensitive, as the
  issue notes.
- Tests: parametrized over all five casings, plus a negative case
  asserting `lastEventIdentifier` / `lastEventIds` / `eventIdCache`
  do not fire. `\b` anchoring keeps this in the conservative direction.

418 tests pass locally.

### DIFF
--- a/src/mcp_migrate/rules/r014_sse_resumability_removed.py
+++ b/src/mcp_migrate/rules/r014_sse_resumability_removed.py
@@ -1,3 +1,5 @@
+import re
+
 from .base import Finding, Project, Rule
 
 PY_RX = r"Last-Event-ID|last_event_id|LAST_EVENT_ID"
@@ -7,7 +9,15 @@ PY_RX = r"Last-Event-ID|last_event_id|LAST_EVENT_ID"
 # valid identifier in either language -- it can only appear as a string
 # literal -- so that half needs search_wire, kept separate from the
 # identifier half the same way R001 splits header-string vs. identifier.
-TS_CODE_RX = r"\blastEventId\b|\bLastEventId\b"
+#
+# Matched case-insensitively (see TS_CODE_FLAGS): real servers write every
+# casing of the same identifier -- `lastEventId`, `lastEventID`,
+# `LastEventId` -- and a const carrying the header name is usually
+# `LAST_EVENT_ID`, so the optional underscores cover the SCREAMING_SNAKE
+# form too. The `\b` anchors keep it bounded: `lastEventIdentifier` and
+# `lastEventIds` are left alone, which is the conservative direction.
+TS_CODE_RX = r"\blast_?event_?id\b"
+TS_CODE_FLAGS = re.IGNORECASE
 TS_HEADER_RX = r"""["'`]last-event-id["'`]|["'`]Last-Event-ID["'`]"""
 
 MESSAGE = "Implements SSE resumability (Last-Event-ID) -- removed from the transport."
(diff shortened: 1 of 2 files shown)


dheerajjha:
Verified and this is a better fix than the issue asked for. **Welcome — this is your first PR here and it is a clean one.**

I ran both against the same corpus rather than reading the regex:

```
hits.ts    lastEventId / LastEventId / lastEventID / LAST_EVENT_ID / last_event_id
misses.ts  lastEventIdentifier / lastEventIds / eventIdCache / broadcastEventIdx
```

| | main | this PR |
|---|---|---|
| hits.ts | lines 1, 2 | **lines 1, 2, 3, 4, 5** |
| misses.ts | none | **none** |
| header forms | 1, 2 | 1, 2 |

Five casings caught, zero false positives, header half untouched.

**The `_?` is the part I want to call out**, because #143 did not ask for it and it is the difference between a partial fix and a complete one. `re.IGNORECASE` alone gets you `lastEventID`, but `LAST_EVENT_ID` and `last_event_id` have separators the flag cannot see. Collapsing four alternations into `\blast_?event_?id\b` covers all of them and is *shorter* than what it replaced.

The negative test matters just as much. `lastEventIdentifier` and `lastEventIds` are exactly the shape that #87 has been open about for a week — a widened pattern that starts matching unrelated identifiers. Anchoring on `\b` and proving it with a test is the right instinct, and it is the instinct this repo most needs from contributors.

Keeping the wire half case-sensitive is also correct and I am glad you did not widen it "for consistency" — headers are conventionally lowercased on the wire, and loosening that pattern risks matching prose containing "last event".

## Credit

This closes #143, which came out of @PuvaanRaaj's #60 — an independent R014 port that got case-insensitivity right when the port that merged did not. **Their name is on the issue and it belongs in this thread too.**

## One thing not your fault

CI shows a failure. It is not this PR: `test (3.11)` is being **cancelled** by GitHub without ever starting — 15 minutes queued, zero steps recorded — while 3.10, 3.12 and 3.13 all pass. Three separate runs on three different branches show the same shape since 15:07Z. Your branch passes 418 tests locally.

Issue #113, opened 2026-08-05
A comment saying `server/discover` is missing convinces R010 that it isn't


dheerajjha:
R010 reports "this project registers MCP request handlers but never implements `server/discover`". It decides the project *has* `server/discover` using raw `project.search`, which matches comments and docstrings — so writing a comment that says you haven't implemented it convinces the tool that you have.

`src/mcp_migrate/rules/r010_server_discover_missing.py:119`:

```python
def _has_discover(project: Project) -> bool:
    if any(project.search_code(DISCOVER_CODE_RX.pattern)):
        return True
    if any(project.search(r"server/discover")):      # <- matches prose
        return True
    return False
```

## Reproduce

Same server, three times:

```
R010 findings: 1   <- no mention of discover at all
R010 findings: 0   <- plus: # TODO: server/discover is not implemented yet
R010 findings: 0   <- plus: """We do not implement server/discover."""
```

The more clearly you document the gap, the more certain the tool is that there is no gap. A `# TODO` admitting the exact thing R010 checks for is the most likely sentence in the codebase to appear next to this defect, so this isn't a contrived input.

## The TypeScript twin already gets this right, and says why

Thirty lines below, `_ts_has_discover` does the same job and deliberately does not copy the Python version:

```python
    # search_wire rather than the Python side's raw search: the wire name
    # is a string literal, but a JSDoc block saying "server/discover is not
    # implemented yet" is prose, and letting prose satisfy the check would
    # suppress the finding for the whole project -- the worst failure mode
    # available here, since nobody ever sees a finding that was silently
    # withheld.
    if any(project.search_wire(r"server/discover")):
```

So the reasoning is already written down in this file. The Python branch predates it and was never brought along.

## Why this one is worth care

Most of our open false-positive bugs make the tool report something that isn't there. This is the opposite: it makes the tool go quiet about something that is. That's normally the safe direction here — a false positive is worse than a missed finding — but that rule is about *reporting*, not about *suppression logic*. A permissive check inside a "should I stay silent?" gate inverts it: the loose match doesn't cost a wrong finding, it costs the right one, and nobody ever sees a finding that was withheld.

Worth reading `Project.search_wire`'s docstring in `rules/base.py` before starting — it lays out the three-way split, and the whole of this fix is choosing correctly between them.

## Done when

- The reproduction above reports **1 finding in all three cases**.
- A fixture under `tests/fixtures/` covers the comment case and the docstring case for the Python path.
[...]

Issue #17, opened 2026-07-31
Fixer: Roots/Sampling/Logging deprecated core features (R007)


dheerajjha:
No fixer exists yet for R007 ([`src/mcp_migrate/rules/r007_deprecated_features.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r007_deprecated_features.py)),
which flags dependencies on Roots, Sampling or Logging as core capabilities.

**Spec:** "Roots, Sampling and Logging deprecated" -- https://modelcontextprotocol.io/specification/draft/changelog

**A reasonable scope:** rather than trying to migrate the functionality
(impossible without understanding what the handler does), consider the
`r001_session_id.py` fixer's pattern -- annotate the flagged line with a
`# TODO(mcp-migrate): ...` comment pointing at the spec and
`cookbook/18-roots-sampling-logging-deprecated.md`, `review` confidence,
without touching the code's behavior at all. That's mechanical, safe, and
genuinely useful (a loud, precise pointer at what needs human attention)
without pretending to migrate something this fuzzy automatically.

**What "done" looks like:** a `Fixer` subclass in `src/mcp_migrate/fixers/`
following that TODO-annotation pattern, with fixtures and tests per
`tests/test_fixers.py`.

See CONTRIBUTING.md#add-a-fixer-45-minutes.

PR #189, opened 2026-08-06
fix(scan): load JavaScript source files

### LINKED ISSUE #149: JavaScript is counted but never read: make the scanner load .js, then port the rules
Two extension maps disagree, and the result is that a plain-JavaScript MCP server gets scanned, counted, and reported on -- with zero findings, because none of its files were ever loaded.

## Reproduce

Same three lines, twice, one file each:

```js
// server.js AND server.ts -- byte-identical
const { SSEServerTransport } = require("@modelcontextprotocol/sdk/server/sse.js");
const sessionId = req.headers["Mcp-Session-Id"];
const e = { code: -32002, message: "resource not found" };
```

```
$ mcp-migrate check .
No grade for this one. Found 1 JavaScript, 1 TypeScript. ...

  breaking    R001  server.ts:2  Mcp-Session-Id was removed ...
  breaking    R017  server.ts:3  -32002 for resource-not-found is the old code ...
  deprecated  R006  server.ts:1  HTTP+SSE transport is deprecated.

3 finding(s) from the rules that do cover this language.
```

**Three findings for `.ts`, zero for the identical `.js`.** The output says "Found 1 JavaScript", so it looks like JavaScript was examined. It was counted, not read.

## Cause

There are two extension tables and only one of them knows about JavaScript.

```python
# src/mcp_migrate/languages.py -- used by survey(), which produces "Found 1 JavaScript"
".js": "javascript", ".jsx": "javascript", ".mjs": "javascript", ".cjs": "javascript",
```

```python
# src/mcp_migrate/scan.py -- used by load_project(), which decides what rules see
SOURCE_EXTENSIONS = {
    ".py": "python",
    ".ts": "typescript", ".mts": "typescript", ".cts": "typescript",
    ".tsx": "typescript",
}
```

`load_project` calls `_language_of(path)`, gets `None` for `.js`, and skips the file entirely.

## Why it matters

A large share of MCP servers are plain JavaScript or ESM. Today those users get a clean-looking run that checked nothing. **That is the exact failure this tool exists to refuse**: reporting on code it did not read.

---

# Step 1 -- make the scanner read JavaScript

1. Add `.js`, `.jsx`, `.mjs`, `.cjs` to `scan.SOURCE_EXTENSIONS` mapping to `"javascript"`.
2. Route the content-span logic to the TypeScript implementation. JS and TS are identical for what the scanner needs -- `//` and `/* */` comments, single/double/backtick strings. **Check whether the span code keys off the language string; if it does, `"javascript"` must reach the same branch as `"typescript"` rather than fall through to the Python path.**
3. `scan.DECLARATION_SUFFIXES` excludes generated `.d.ts`/`.d.mts`/`.d.cts`. There is no `.d.js`, so probably nothing to mirror -- but say so in the PR rather than leaving it unconsidered.

**This step alone changes no user-visible output**, because no rule declares `"javascript"` yet. The deliverable is a test asserting a `.js` file now appears in `project.files` with `language == "javascript"`.

# Step 2 -- port the rules

Twenty rules read TypeScript. Each needs `"javascript"` added to `languages` **and a JavaScript fixture proving it works** -- the tuple edit without the fixture is not a port, it is a claim.

Three of them have no TypeScript-specific branch at all (their patterns are wire strings and literals spelled identically in every language), so they are genuinely one-line changes once Step 1 lands:

- [ ] R006 -- deprecated HTTP+SSE transport
- [ ] R017 -- old `-32002` resource-not-found code
- [ ] R021 -- older JSON Schema dialect pinned

The other seventeen have dedicated TypeScript branches built on `TS_*` constants, and each needs those patterns checked against JavaScript idioms rather than assumed:

- [ ] R001 `TS_IDENT_RX`, `TS_HEADER_RX`
- [ ] R003 `TS_HTTP_CALL`, `TS_MCP_SURFACE_RX`
- [ ] R004 `TS_HANDLER_RX`, `TS_SORT_RX`
- [ ] R005 `TS_CAPS_TYPE_RX`, `TS_SDK_IMPORT_RX`
- [ ] R007 `TS_OWNER`, `TS_MCP_SURFACE_RX`
- [ ] R008 `TS_OTEL_SPECIFIER_RX`, `TS_TRACEPARENT_RX`
- [ ] R009 `TS_HANDSHAKE_CODE_RX`
- [ ] R010 `TS_LOWLEVEL_HANDLER_RX`, `TS_MCPSERVER_REGISTER_RX`
- [ ] R011 `TS_PING_CODE_RX`, `TS_PING_DISPATCH_RX`
- [ ] R012 `TS_SET_LEVEL_CODE_RX`
- [ ] R013 `TS_SUBSCRIBE_CODE_RX`
- [ ] R014 `TS_CODE_RX`, `TS_HEADER_RX`
- [ ] R015 `TS_SDK_IMPORT_RX`, `TS_JSONRPC_ENVELOPE_RX`
- [ ] R016 `TS_CACHEABLE_HANDLER_RX`, `TS_CACHEABLE_METHOD_RX`
- [ ] R018 `TS_FEATURES_CODE`
- [ ] R019 `TS_TASKS_CODE_RX`
- [ ] R020 `TS_DCR_CODE_RX`

(R002 reads neither TypeScript nor JavaScript yet -- see #32.)

## What actually differs between JS and TS

This is where a careless port goes wrong:

- **`require()` vs `import`.** A CommonJS server writes `const { Server } = require("@modelcontextprotocol/sdk/server/index.js")`. Any pattern anchored on `import ... from` misses it entirely.
- **No type annotations.** Patterns matching `: SomeType`, `as SomeType`, `interface`, or a generic parameter cannot fire in JavaScript. A rule leaning on those for its "is this MCP?" evidence needs another signal or it will never fire on JS.
- **The SDK is identical.** Zod schema names (`...RequestSchema`) and wire method strings are spelled the same, so those halves port for free.
- **`.mjs` and `.cjs`** are in scope, not just `.js`.

**Write a failing JavaScript fixture first, then make it pass.**

## How to take this on

Step 1 is one PR and blocks everything else. After that, the rule ports are independent -- **claim a few rules in a comment and open one PR per small group**, rather than one PR per rule or one giant PR. Comment here and I will tick them off as they land.

Originally filed as 21 separate issues; collapsed because a one-line tuple change does not deserve its own tracking item.

aryansk:
## Summary
- load js, jsx, mjs, and cjs files as JavaScript source
- use the TypeScript comment and string span scanner for JavaScript
- keep generated d.ts declarations excluded
- add a regression test for all supported JavaScript extensions

Closes #149

## Validation
- PYTHONPATH=src uv run --with pytest --with pyyaml --with rich pytest
- 411 tests passed

### DIFF
--- a/src/mcp_migrate/rules/base.py
+++ b/src/mcp_migrate/rules/base.py
@@ -161,7 +161,8 @@ class Project:
         key = id(f)
         if key not in self._span_cache:
             self._span_cache[key] = (
-                _ts_spans(f.text, prose_only=False) if f.language == "typescript"
+                _ts_spans(f.text, prose_only=False)
+                if f.language in {"typescript", "javascript"}
                 else _content_spans(f.text)
             )
         return self._span_cache[key]
@@ -170,7 +171,8 @@ class Project:
         key = ("prose", id(f))
         if key not in self._span_cache:
             self._span_cache[key] = (
-                _ts_spans(f.text, prose_only=True) if f.language == "typescript"
+                _ts_spans(f.text, prose_only=True)
+                if f.language in {"typescript", "javascript"}
                 else _prose_spans(f.text)
             )
         return self._span_cache[key]
--- a/src/mcp_migrate/scan.py
+++ b/src/mcp_migrate/scan.py
@@ -59,6 +59,8 @@ SOURCE_EXTENSIONS = {
     ".py": "python",
     ".ts": "typescript", ".mts": "typescript", ".cts": "typescript",
     ".tsx": "typescript",
+    ".js": "javascript", ".jsx": "javascript",
+    ".mjs": "javascript", ".cjs": "javascript",
 }
 
 
--- a/tests/test_scan.py
+++ b/tests/test_scan.py
@@ -24,3 +24,34 @@ def test_generated_declarations_are_skipped_for_every_module_form(tmp_path):
     project = load_project(tmp_path)
     scanned = {f.path.name for f in project.files}
     assert scanned == {"impl.ts", "impl.mts", "impl.cts"}
+def test_load_project_reads_javascript_and_skips_declaration_output(tmp_path):
+    (tmp_path / "server.js").write_text("const answer = 42;\n")
+    (tmp_path / "component.jsx").write_text("export default () => null;\n")
+    (tmp_path / "module.mjs").write_text("export const answer = 42;\n")
+    (tmp_path / "common.cjs").write_text("module.exports = {};\n")
+    (tmp_path / "types.d.ts").write_text("export declare const answer: number;\n")
+
+    project = load_project(tmp_path)
+
+    assert [(file.path.name, file.language) for file in project.files] == [
+        ("common.cjs", "javascript"),
+        ("component.jsx", "javascript"),
+        ("module.mjs", "javascript"),
+        ("server.js", "javascript"),
+    ]
+
+
+def test_javascript_uses_the_typescript_comment_and_string_spans(tmp_path):
+    """Loading `.js` is only half the job. If `javascript` did not reach the
+    TypeScript span branch it would fall through to the Python one, and `//`
+    would stop being recognised as a comment -- so a rule would read
+    commented-out code as live. That is #113's failure, one language over,
+    and the extension map alone does not catch it."""
+    (tmp_path / "a.js").write_text(
+        '// const sessionId = req.headers["Mcp-Session-Id"];\n'
+        'const real = 1;\n'
+    )
+    project = load_project(tmp_path)
+    assert list(project.search_code(r"Mcp-Session-Id")) == [], (
+        "a // comment must not read as code on a .js file"
+    )


dheerajjha:
Verified -- this is Step 1 of #149 done correctly, and **you got the half that is easy to miss.** Welcome; first PR here.

Adding `.js`/`.jsx`/`.mjs`/`.cjs` to `SOURCE_EXTENSIONS` is the obvious part. Routing `javascript` into `_ts_spans` in `rules/base.py` is the part that actually matters:

```python
_ts_spans(f.text, prose_only=False)
if f.language in {"typescript", "javascript"}
```

Without it, `.js` files would fall through to the Python span path and `//` would stop being recognised as a comment -- so every rule would read commented-out code as live. **That is exactly #113, one language over**, and the extension map alone would have shipped it.

I added a test asserting precisely that, since neither your test nor mine covered it:

```python
def test_javascript_uses_the_typescript_comment_and_string_spans(tmp_path):
    (tmp_path / "a.js").write_text(
        '// const sessionId = req.headers["Mcp-Session-Id"];\n'
        'const real = 1;\n'
    )
    assert list(load_project(tmp_path).search_code(r"Mcp-Session-Id")) == []
```

## Rebase

I rebased this onto `main` and pushed rather than bouncing it back. Two things had landed since you branched, both touching the same file:

- `scan.DECLARATION_SUFFIXES` now excludes `.d.mts` and `.d.cts`, not just `.d.ts`
- `tests/test_scan.py` did not exist when you started; we both created it

The resolution was additive -- **both tests kept, nothing of yours changed**, plus the span test above. Your authorship is on the commit.

## Confirmed it changes no findings

As #149 said it should not -- no rule declares `"javascript"` yet:

```
before:  files_scanned: 1   findings: server.ts x3
after:   files_scanned: 2   findings: server.ts x3
         languages: {javascript: 1, typescript: 1}
```

The `.js` file is now read. Nothing reads *into* it yet -- that is Step 2, the per-rule checklist on #149, and it is now unblocked. **Three of those rules (R006, R017, R021) are one-line changes** if you want to keep going; the other seventeen need their `TS_*` patterns checked against `require()` and the absence of type annotations.

## CI

Shows a failure that is not yours: `test (3.11)` is being cancelled by GitHub without ever starting -- queued 15 minutes, zero steps recorded -- while 3.10, 3.12 and 3.13 all pass. Same shape on three unrelated branches since 15:07Z. 413 tests pass locally on the rebased branch.

dheerajjha:
**This is ready from my side — it is marked as a draft, so I have not merged it.** Mark it ready for review and it goes in; I did not want to flip that flag for you in case you were still working.

Reviewed, rebased onto `main`, and verified above. It is green locally (413 tests on the rebased branch) and does exactly what Step 1 of #149 asked.

Everything else from tonight has landed around it, so when you un-draft it will merge cleanly onto a `main` that now has:

- `.d.mts` / `.d.cts` excluded alongside `.d.ts`
- three more fixers (R009, R011, R012), taking the set to 16 of 21
- R014 catching all five `lastEventId` casings, from @ujjwalprakash17's first PR

If you have gone off the idea, say so and I will pick it up — no hard feelings, and the review stands either way.

dheerajjha:
Merging this. Two things you should know, since I did something I said I would not.

**I flipped the draft flag myself.** I said I would wait for you to do it. Then a second PR arrived doing the same fix (#192), and waiting would have meant your work getting overtaken by someone who started later. You filed first and the fix is right, so this goes in first, under your name. If you were still mid-thought on it, say so and I will revert — but it is green and complete, so I judged landing it was the better call.

Rebased onto current `main` again (it had moved: R002 landed, so all 21 rules now read TypeScript). Clean rebase, **454 tests pass**.

## One thing this does not cover

I tested it in isolation and found a gap neither of us wrote a test for. On a pure-JavaScript project containing a live breaking pattern:

```
$ cat server.js
const sessionId = req.headers["Mcp-Session-Id"];

$ mcp-migrate check .    # with this PR alone
exit=0
```

Before this PR that was `exit=2`. `_checked_something()` in `cli.py` is `bool(project.files)` — so the moment `.js` files start loading, a project where **zero rules ran** starts reporting *clean*. Nothing checked that file; the tool now says it is fine.

That is not your bug — it is a latent one your fix is the first to reach, and it is the tool telling someone their broken server is healthy. It is worse than the exit 2 it replaces, because exit 2 is a refusal and exit 0 is a verdict.

@waterlemonnn found it independently in #192 and wrote the fix. I am landing that immediately after this, reduced to just that delta, so the two of you each keep what you actually did. Main will be in the regressed state for about a minute.

**Welcome — you're contributor #17,** and you picked the half of this that was easy to get wrong. The `_ts_spans` routing is the part that mattered; the extension map alone would have shipped #113 one language over.

Step 2 of #149 (porting rules to declare `"javascript"`) is now unblocked if you want it. **R006, R017 and R021 are one-line changes.**

aryansk:
Thanks for the thorough review, the extra JavaScript span regression test, and the careful rebases. I appreciate the context about Step 2 as well. This PR is complete from my side, and I am glad you were able to land it cleanly under the original contribution.

PR #190, opened 2026-08-06
docs: add executable check --json schema contract

### LINKED ISSUE #188: The --json schema is documented nowhere, and it just changed in a breaking way
`check --json` is a supported output that other tools are expected to consume, and its shape is documented nowhere. The README mentions the flag exists:

> `--json` for machine-readable, uncapped output

That is the whole specification.

## Why this needs fixing now specifically

The schema **just changed in a breaking way**. #144 replaced `location` (a `"path:line"` string) with separate `path` and `line` fields, and added `tool`, `version` and `counts`. That shipped in 0.1.0 through 0.1.4 and is gone.

Nothing in the repository still reads `location` -- I traced every consumer before merging -- but external scripts are invisible from here, and there was no document to check against or to deprecate from. The next change will have the same problem unless the shape is written down.

## The work

1. **Document the current shape** in the README, next to the `--json` mention: every top-level key, every finding key, which are always present and which are conditional (`fix` is only emitted when the rule has remediation text).
2. **A JSON Schema in the repository**, and a test validating real `check --json` output against it. That makes the contract executable rather than aspirational, and it means the next breaking change is a deliberate schema edit rather than a surprise.
3. **State the compatibility policy.** Pre-1.0, "this may change and the CHANGELOG will say so under Changed" is a perfectly good policy -- but say it, so consumers know what they are building on.

## Current shape, for reference

```json
{
  "tool": "mcp-migrate",
  "version": "0.1.4",
  "path": "...", "spec": "...",
  "scannable": true, "files_scanned": 3,
  "languages": {"python": 3},
  "grade": "C", "score": 72,
  "counts": {"breaking": 1, "deprecated": 0, "advisory": 0},
  "findings": [
    {"rule": "R017", "severity": "breaking", "path": "s.py", "line": 3,
     "message": "...", "fix": "..."}
  ]
}
```

`grade` and `score` are absent for trees that got no grade -- worth confirming and documenting rather than assuming.

slegarraga:
## What & why

Makes the `check --json` shape a documented, executable contract instead of a README hint. External scripts now have a stable schema to validate against and a stated pre-1.0 compatibility policy.

Closes #188.

## Changes

- Added `schemas/check-json.schema.json` covering the normal, SDK, and unscannable output variants.
- Added `tests/test_check_json_schema.py`, which validates real output from clean, legacy, empty, and SDK fixtures against the schema.
- Documented the always-present keys, conditional keys, and finding shape in the README, next to the `--json` mention.
- Added `jsonschema` to the dev extra and noted the contract in the CHANGELOG under Unreleased.

## Verification

- `uv sync --extra dev`
- `uv run pytest -q`: 449 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,12 @@ contributed.
   the `extensions` map, trace context propagation, and Roots/Sampling/
   Logging deprecation. All eighteen slots are now written; there are no
   stubs left. (@waterlemonnn)
+- **`check --json` now has an executable schema contract** in
+  `schemas/check-json.schema.json`, documented in the README and validated
+  against clean, legacy, unscannable, and SDK outputs. Pre-1.0 compatibility
+  policy is stated next to the contract so external consumers know breaking
+  shape changes will be documented under Changed.
+  ([#188](https://github.com/dheerajjha/mcp-migrate/issues/188))
 
 ### Changed
 
--- a/README.md
+++ b/README.md
@@ -56,6 +56,28 @@ Zero findings prints `Grade A` and a ready-to-paste badge instead. Add
 rule at 5 rows with a "+N more" line so it stays readable; JSON always has
 every finding).
 
+### `check --json` contract
+
+`--json` emits one JSON object with these always-present top-level keys:
+`tool`, `version`, `spec`, `path`, `scannable`, `languages`, `grade`,
+`score`, `files_scanned`, `counts`, and `findings`.
+
+Conditional keys:
+
+- `is_sdk` and `sdk_reason` only when the tree is a protocol SDK. In that case `grade` and `score` are `null`.
+- `reason` only when `scannable` is `false`. In that case `grade` and `score` are `null`.
+- `grade` and `score` are also `null` when a tree was read but is not gradable, for example TypeScript-only trees.
+
+Each finding always has `rule`, `severity`, `path`, `line`, and `message`.
+`path` and `line` may be `null` for project-level findings. `fix` is present
+only when the rule has remediation text; it is omitted, never `null`.
+
+The executable contract is
+[`schemas/check-json.schema.json`](schemas/check-json.schema.json). Pre-1.0,
+this shape may change in a breaking way; such changes are documented under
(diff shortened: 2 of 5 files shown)


dheerajjha:
Verified and merged. **Welcome — first PR here, and it is a careful one.**

I checked the schema against outputs your tests do not cover, because a schema that only validates the fixtures it was written against is decoration:

| output | result |
| --- | --- |
| Python, graded | valid |
| **TypeScript-only (ungraded)** | valid |
| **mixed `.js` + `.ts`** | valid |
| unscannable / empty tree | valid |
| post-`fix --write` tree | valid |

The TypeScript case is the one I expected to break it and it did not.

## You corrected the issue, which is the part I want to flag

#188 as I filed it said:

> `grade` and `score` are absent for trees that got no grade -- worth confirming and documenting rather than assuming.

**They are not absent. They are present and `null`**, on every ungraded path — SDK trees, unscannable trees, and TypeScript-only trees alike. You checked instead of inheriting my guess, and the README now says so in three separate cases rather than one. That is the difference between documentation and transcription.

Two more details you got right that I would have missed:

- **`path` and `line` may be `null` for project-level findings.** Confirmed: R010 reports against the project, not a line — `{"rule": "R010", "path": null, "line": null}`. A consumer assuming a string there would crash on any tree missing `server/discover`.
- **`fix` is omitted, never `null`.** Verified across a mixed tree. Those are different contracts and conflating them is exactly the kind of thing a schema exists to pin down.

## Mechanics

`jsonschema` goes in the `dev` extra, and CI runs `uv sync --extra dev`, so it installs — I checked rather than assumed, since a test that silently skips is worse than no test. The schema is vendored rather than fetched, so CI stays offline; that was the one thing #188 asked for explicitly and you did it without being reminded.

449 tests pass locally, and your branch is green on all four Python versions.

## Note on timing

CI was broken for about six hours last night — `test (3.11)` was being cancelled by GitHub without ever starting. **It recovered before your PR ran**, so your green checks are real. Mentioning it only because you may have seen odd red marks on other PRs in the meantime.

Issue #100, opened 2026-08-04
No GitHub Action, so nobody runs this in CI


dheerajjha:
There is no way to run this in CI without writing the glue yourself: install Python, `pip install mcp-migrate`, run it, interpret the exit code. For a TypeScript repo — which is most MCP servers — that means adding Python to a Node toolchain for one command. Most people won't.

A published action turns it into four lines:

```yaml
- uses: dheerajjha/mcp-migrate-action@v1
  with:
    path: ./src
    fail-on: breaking
```

### Why this is worth doing

The tool's whole value is catching a breaking change *before* it ships. Today it's a thing you remember to run manually, which means it runs once, at migration time, and never again. In CI it runs on every PR forever.

It's also the highest-visibility surface we have. An action appears in the GitHub Marketplace, in other people's workflow files, and in their PR checks — in front of exactly the people who need this tool and don't know it exists.

### Depends on two open issues

Please don't start this before they land, or you'll build on sand:

- **#98** — every TypeScript project currently exits `2`, so `fail-on` cannot work at all for the majority language.
- **#85** — `--format json`, so the action can annotate individual lines rather than dumping a table into the log.

Comment here if you want it and I'll ping you when both are merged.

### Shape

- Composite action (no Docker) — faster, and this is a `pip install` plus one command.
- `path` (default `.`), `fail-on` (`breaking` | `deprecated` | `advisory` | `never`), `include-tests`.
- **Annotations on the diff.** `::error file=…,line=…::message` is what makes a CI check useful rather than something people learn to ignore. This is the part that matters most; a check that just goes red teaches people to skip it.
- Pin the `mcp-migrate` version by default and allow an override. An action that silently changes behaviour when we ship a rule is an action people disable.

### Two real questions

**Separate repo or this one?** Marketplace publishing wants its own repo and its own tags. That's the conventional answer, but it splits the work across two places. I lean separate; convince me either way.

**What should `fail-on: breaking` do about an ungraded project?** Once #98 is fixed the exit code is honest, but "16 of 21 rules ran" is still a caveat the action should probably surface in the check summary rather than swallow.

Say which parts you want. Happy for this to arrive in pieces — the action skeleton alone is useful before annotations exist.

PR #202, opened 2026-08-07
feat(cli): add SARIF 2.1.0 output for GitHub code scanning

### LINKED ISSUE #182: Add SARIF output so findings land in GitHub code scanning
`check --json` emits a shape of our own design. GitHub code scanning, and most other tooling that would consume this, speaks SARIF.

## Why it matters

There is an open issue (#100) for shipping a GitHub Action. An Action that can only print to a log is a fraction as useful as one that files findings into the Security tab, annotates the diff in a pull request, and tracks them across commits. SARIF is what unlocks that, and it is a pure output-format addition -- no rule or scanner changes.

## The work

`mcp-migrate check --format sarif` producing SARIF 2.1.0.

The mapping is mostly mechanical, and the pieces already exist:

| SARIF | source |
| --- | --- |
| `tool.driver.name` / `.version` | the `tool` / `version` keys added in #144 |
| `rules[]` with `id`, `shortDescription` | `all_rules()` -- `id`, `title` |
| `rules[].helpUri` | `rule.spec_ref` (populated on all 21) |
| `rules[].fullDescription` | `rule.fix` -- the remediation text |
| `results[].ruleId` / `level` / `message` | finding `rule`, `severity`, `message` |
| `results[].locations[]` | finding `path` + `line` |

Two things to decide rather than guess:

- **Severity mapping.** SARIF has `error`/`warning`/`note`. `breaking` to `error` and `advisory` to `note` are obvious; `deprecated` could go either way, and the choice determines whether a deprecation blocks a PR by default. State the reasoning in the PR.
- **`--format` vs `--json`.** `--json` already exists and is now depended on. Adding `--format {text,json,sarif}` and keeping `--json` as an alias is friendlier than a second boolean flag -- but that is a CLI-surface decision worth agreeing before implementing.

## Test

Validate the output against the SARIF 2.1.0 schema. `jsonschema` as a dev dependency with a vendored copy of the schema keeps CI offline, which matters -- a test that fetches a schema over the network is a flake waiting to happen.

ankitverma31:
Fixes #182.

```bash
mcp-migrate check . --format sarif > mcp-migrate.sarif
```

Pure output format — no rule or scanner changes, just a projection of the findings `--json` already carries into the shape code scanning speaks. Findings annotate the diff and land in the Security tab instead of scrolling past in a log.

### `deprecated` → `warning`, not `error`

`breaking` → `error` and `advisory` → `note` are forced. This one is the real choice, because **code scanning's default gate fails on `error` alone** — so the mapping decides whether a deprecation blocks a pull request.

The spec gives deprecated features at least twelve months. That makes a deprecation categorically *"you have time, but start"*, which is what `warning` means. Mapping it to `error` would make every server using Roots or Sampling unmergeable **today** over a change that doesn't break until next year — and the predictable response to that is switching the whole integration off, at which point it catches nothing.

### `--format {text,json,sarif}`, with `--json` as an alias

Taking your suggestion over a second boolean. `--json` predates this and is depended on, so it stays — and both resolve through one function, which is what stops them from ever disagreeing. There's a test asserting `--json` and `--format json` produce byte-identical output.

## Three details that decide whether this actually works

**Paths are relative to the scanned root, forward-slashed.** Code scanning matches results to the diff by path. An absolute path from the scanning machine matches nothing and the annotations silently never appear — no error, just no annotations.

**The driver declares all 21 rules, not only the ones that fired.** That's how a consumer distinguishes "this rule ran and found nothing" from "this rule doesn't exist", which is what lets code scanning **close** a resolved alert rather than leaving it open forever. Same reason a clean run still emits a full document with an empty `results` array.

**A prose `spec_ref` doesn't become a `helpUri`.** Rules like R007 have `spec_ref = "Roots, Sampling and Logging deprecated"` — not a URI. Emitting it as one makes consumers reject the entire document, so it rides in `properties` instead and only real URLs become `helpUri`.

Project-level findings (R010 asks a question about the whole tree and has no file) keep an empty `locations` array rather than being dropped.

## Schema validation

The SARIF 2.1.0 schema is vendored at `schemas/sarif-2.1.0.schema.json` and validated against in CI — as you asked, a test that fetches a schema over the network is a flake waiting to happen, and this one runs on four Python versions per push. `jsonschema` was already a dev dependency for `check-json.schema.json`, so no new dep.

Real output on `tests/fixtures/legacy_server`: 53 results, 21 declared rules, levels `error`/`warning`/`note`, validates clean.

## Tests (20)

Schema validation on clean and legacy fixtures · the severity mapping · every result's level matching its rule's severity · `ruleIndex` pointing at the right descriptor (off by one and every finding is attributed to the wrong rule, with no error anywhere) · relative paths · line numbers · project-level findings surviving · prose `spec_ref`s not becoming URIs · `--json` unchanged · `--json` ≡ `--format json` · and exit codes unchanged by format — the format decides what is printed, never what is returned.

Full suite: **479 passed**.

---

This is the piece #100 needs to be worth shipping: an Action that can only print to a log is a fraction as useful as one that files findings into the Security tab. Happy to follow up with the Action itself now that #98 and #85 have both landed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,42 @@ All notable changes to this project are documented here.
 
 ## [Unreleased]
 
+### Added
+
+- **SARIF 2.1.0 output: `check --format sarif`.** Findings land in GitHub
+  code scanning as annotations on the diff and entries in the Security tab,
+  tracked across commits instead of re-read from a CI log every time.
+  ([#182](https://github.com/dheerajjha/mcp-migrate/issues/182))
+
+  `--format {text,json,sarif}` is the new surface; `--json` keeps working as
+  an alias for `--format json` and its output is byte-identical. The format
+  decides what is printed and never what is returned — exit codes are
+  unchanged across all three.
+
+  Two decisions worth stating rather than burying:
+
+  - **`deprecated` maps to `warning`, not `error`.** Code scanning's default
+    gate fails on `error` alone, so this mapping decides whether a
+    deprecation blocks a merge. The spec gives deprecated features 12+
+    months; making a server unmergeable today over something that breaks
+    next year gets the integration switched off, and then it catches
+    nothing. `breaking` → `error` and `advisory` → `note`.
+  - **The driver declares all 21 rules, not only those that fired.** That is
+    what lets code scanning tell "this rule ran and found nothing" from
+    "this rule does not exist", and therefore close a resolved alert instead
+    of leaving it open forever. A clean tree still emits a full run with
+    `results: []`.
+
+  Paths are repo-relative and forward-slashed on every platform, because
+  code scanning matches results to the diff by path and an absolute path
+  from the scanning machine matches nothing — the annotations would silently
+  never appear. Project-level findings (R010 asks about the whole tree) are
+  kept with an empty `locations` array rather than dropped.
+
+  Validated in CI against a vendored copy of the official schema at
+  `schemas/sarif-2.1.0.schema.json`; it is checked in rather than fetched so
+  the suite does not depend on the network across four Python versions.
+
 ### Fixed
 
 - **R004 no longer fires on a wire method name that returns no tools.**
--- a/README.md
+++ b/README.md
@@ -78,6 +78,28 @@ this shape may change in a breaking way; such changes are documented under
 Changed in [CHANGELOG.md](CHANGELOG.md), so consumers should pin a version
 and watch that section.
 
+### `check --format sarif`
+
+SARIF 2.1.0, for GitHub code scanning and anything else that speaks it:
+
+```yaml
+- run: mcp-migrate check . --format sarif > mcp-migrate.sarif
+- uses: github/codeql-action/upload-sarif@v3
+  with:
+    sarif_file: mcp-migrate.sarif
+```
+
+Findings land in the Security tab and annotate the diff, instead of scrolling
+past in a log. Severities map to SARIF levels as `breaking` -> `error`,
+`deprecated` -> `warning`, `advisory` -> `note`. `deprecated` is deliberately
+not `error`: code scanning's default gate fails on `error` alone, and the spec
+gives deprecated features 12+ months, so blocking a merge today over a change
+that doesn't break until next year is how an integration gets switched off.
+
+`--format {text,json,sarif}` is the general flag; `--json` remains as an alias
+for `--format json`. Output is validated against the vendored
+[SARIF 2.1.0 schema](schemas/sarif-2.1.0.schema.json) in CI.
+
 Exit codes, so it drops straight into CI:
 
 | code | meaning |
--- a/schemas/sarif-2.1.0.schema.json
+++ b/schemas/sarif-2.1.0.schema.json
@@ -1,0 +1,3012 @@
+{
+  "$id": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "additionalProperties": false,
+  "definitions": {
+    "address": {
+      "additionalProperties": false,
+      "description": "A physical or virtual address, or a range of addresses, in an 'addressable region' (memory or a binary file).",
+      "properties": {
+        "absoluteAddress": {
+          "default": -1,
+          "description": "The address expressed as a byte offset from the start of the addressable region.",
+          "minimum": -1,
+          "type": "integer"
+        },
+        "fullyQualifiedName": {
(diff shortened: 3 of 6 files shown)


dheerajjha:
Reviewed and verified — **this is ready apart from one small thing.**

I rebased it onto current `main` locally (clean, **552 tests pass**) and checked every claim in your description against real output rather than reading the diff, because SARIF fails in ways that are invisible until GitHub silently drops your annotations.

## What I verified

```
$ mcp-migrate check tests/fixtures/legacy_server --format sarif
```

| | |
|---|---|
| validates against the vendored SARIF 2.1.0 schema | ✅ |
| 21 rules declared, 53 results | ✅ |
| `ruleIndex` → correct descriptor for every result | ✅ **no mismatches** |
| paths relative, forward-slashed, none absolute | `auth.py`, `handshake.py`, … |
| R010 (project-level) kept with `locations: []` | ✅ not dropped |
| clean tree → full document, `results: []`, 21 rules declared | ✅ |
| `--json` vs `--format json` | **byte-identical** |
| exit code by format (text/json/sarif) | `1` / `1` / `1` |
| prose `spec_ref` not emitted as `helpUri` | ✅ none non-URL |

The `ruleIndex` one is the check I most expected to fail — an off-by-one there attributes every finding to the wrong rule with no error anywhere, and it's correct.

## On `deprecated` → `warning`

Agreed, and for your reason rather than as a default. Code scanning's PR gate fails on `error` alone, so this mapping decides whether a deprecation blocks a merge. The spec gives deprecated features 12+ months; making a server unmergeable **today** over something that breaks next year gets the integration switched off, and then it catches nothing. `warning` is the honest level for "you have time, but start."

Declaring all 21 rules rather than only those that fired is the other detail that matters and is easy to miss — it's what lets code scanning **close** a resolved alert instead of leaving it open forever.

## The one ask

**Add a CHANGELOG entry** under `## [Unreleased]` → `### Added`. Nothing else. The PR touches README, `sarif.py`, `cli.py`, the schema and tests, but not `CHANGELOG.md`.

Worth knowing: as of #193, the docs are asserted against the code, so drift between the README and `all_rules()`/`all_fixers()` now fails CI. The CHANGELOG isn't machine-checked, so it's the one surface still on the honour system.

## Timing

**v0.2.0 shipped today** (every rule reads TypeScript, JavaScript is read, 19 of 21 fixers, a pre-commit hook — your hook is in it). This is targeted at **0.3.0**, together with #203, because that one changes the `--json` contract again and I'd rather not stack two breaking shape changes into one release without a gap.

Push the CHANGELOG entry and I'll merge.

ankitverma31:
CHANGELOG entry pushed, under `## [Unreleased]` → `### Added`. Rebased onto current `main` first; 561 tests pass locally.

I put the two decisions in the entry rather than just naming the feature — `deprecated` → `warning` and the driver declaring all 21 rules. Neither is recoverable from the diff without reading `sarif.py`, and both are the kind of thing someone revisits in six months asking "why was this done that way".

Thanks for exercising it against real output rather than reading the diff. `ruleIndex` was the one I most wanted independently checked for exactly the reason you gave: it fails silently and attributes every finding to the wrong rule, and no schema validation catches it because the document is still perfectly valid.

Noted on 0.3.0 alongside #203, and agreed on the gap — two `--json` shape changes in same-day releases is more churn than a pre-1.0 contract should ask for.

dheerajjha:
Verified and merging. The CHANGELOG entry was the only ask and it does more than close the gap — it records the two decisions that aren't recoverable from the diff, which is exactly what that file is for.

What I checked, beyond re-reading the diff:

- **561 tests pass** locally; green on 3.10–3.13 in CI.
- **`--json` is genuinely byte-identical to `--format json`** — `cmp` on both outputs, not just equal-looking JSON. That's the compatibility claim that matters most here and it holds.
- **Exit codes are invariant across all three formats.** 1 on a tree with breaking findings, 0 on a clean one, 2 on a tree with nothing scannable — same in `text`, `json` and `sarif`. The format decides what is printed and never what is returned, as the entry says.
- **A tree the tool can't read still emits a valid SARIF document** and exits 2, rather than emitting nothing. Code scanning gets a well-formed run either way.
- **All 21 rules declared in the driver**, on a run with 53 results and on an empty one.

The `deprecated` → `warning` mapping is the right call and I want it on the record that it was argued rather than defaulted: a deprecation that breaks in 12+ months shouldn't make a repo unmergeable today, because the response to that is switching the integration off, and a disabled integration catches nothing. Same reasoning that put R003 at `advisory`.

Vendoring the schema rather than fetching it is also right — a network dependency in the suite would fail on someone's laptop and pass in CI, or the reverse, which is the worst kind of test.

Thanks — this is your fourth landed change and the standard hasn't moved.

Issue #181, opened 2026-08-06
No baseline file, so an existing project cannot adopt this in CI incrementally


dheerajjha:
There is no way to adopt this tool on an existing project without either fixing everything first or ignoring the exit code.

A `breaking` finding exits `1`. A mature MCP server hitting this tool for the first time will have several. So the CI story today is all-or-nothing: fix every breaking finding in one pass, or do not wire it into CI at all.

That is the single biggest thing standing between this tool and being *used*, and it is a solved problem elsewhere -- ESLint, mypy, and Ruff all have some form of it.

## The work

A baseline file recording currently-known findings so `check` can fail only on **new** ones.

```
$ mcp-migrate check --write-baseline .mcp-migrate-baseline.json
recorded 27 findings

$ mcp-migrate check          # exits 0 -- nothing new
$ mcp-migrate check          # after someone adds a new Mcp-Session-Id read
  breaking  R001  server.ts:14  ...
  1 new finding (27 baselined)                                  exit 1
```

Design questions that matter more than the plumbing:

- **What identifies a finding across edits?** `path:line` is the obvious key and the wrong one -- inserting a line at the top of a file invalidates every entry below it. Keying on `(rule, path, normalised line content)` survives reformatting and reordering. Getting this wrong makes the feature worse than useless: a baseline that goes stale on every commit trains people to regenerate it blindly.
- **Findings that disappear.** If a baselined finding is gone, say so and offer to prune. A baseline that only grows is a place where fixed things are recorded as broken forever.
- **The grade must ignore the baseline entirely.** The letter is a claim about the code, not about what the team has agreed to tolerate. If a baseline can lift a D to an A, the board becomes fiction.

`check --json` gained stable `path`/`line`/`rule` fields in #144, so the serialisation format is mostly decided already.

## Related

Inline `# mcp-migrate: ignore[R001]` suppression is the other half and is filed separately. Baseline is "not yet"; inline is "this one is wrong". A project wants both, and they should not be the same mechanism.

dheerajjha:
Unclaimed again, so nobody assumes otherwise: PR #233 no longer exists. The
fork it came from was deleted, which took the pull request with it — the number
now 404s rather than showing as closed, so there is nothing to reopen or rebase.

It was close. It had `--write-baseline` and `--baseline` working, and was stuck
on a merge conflict in `cmd_check` against #228 plus three integration questions
I had left. Recording the answers here so the next person does not have to
rediscover them:

[...]

PR #203, opened 2026-08-07
feat: inline suppression, auditable by construction

### LINKED ISSUE #180: No inline suppression, so one false positive means turning a rule off globally
There is no way to tell this tool that a finding is wrong or deliberate. No `# noqa`, no per-line pragma, nothing:

```
$ grep -rn "noqa\|mcp-migrate:" src/mcp_migrate/*.py
(nothing)
```

## Why it matters

This tool ships with **three open false-positive classes** (#87, #88, #89) and says so in its own README: *"Treat a finding as a prompt to look, not as a verdict."* That is honest, and it is also unworkable in CI -- a breaking finding exits `1`, and a user with one false positive has exactly two options: turn the rule off globally, or stop running the tool.

There is also the legitimate case: code that genuinely reads the way a rule matches, deliberately, and is not going to change.

## The work

An inline suppression comment, checked at the line the finding is reported on:

```python
sid = request.headers["Mcp-Session-Id"]  # mcp-migrate: ignore[R001] -- proxy shim, not MCP session state
```

Things worth getting right:

- **Rule-scoped, not blanket.** `ignore` with no rule id suppresses everything on that line and rots silently. Require the id.
- **A reason is not optional.** Or if it is optional, `check` should be able to list every suppression and its reason -- a suppression nobody can audit is just a deleted finding.
- **Both comment syntaxes.** `#` for Python, `//` for TypeScript/JavaScript. `fixers/base.py` already has `comment_prefix(path)` and `is_commented(line)` -- reuse them rather than re-deriving.
- **Suppressions must not be invisible.** At minimum a count in the summary: `3 finding(s), 1 suppressed`. Otherwise this becomes a way to make a bad grade look good, which is precisely what the badge work (#101) was fixing.
- **Does a suppressed finding still count against the grade?** It should not, or suppression is pointless -- but that means the grade is now partly self-reported, and the output has to be honest about that.

That last question is the real design work here. Worth discussing in the issue before writing code.

## Related

A file-level or project-level baseline is the other half of this and is filed separately -- inline suppression is for "this line is fine forever", a baseline is for "not today".

ankitverma31:
Fixes #180.

```python
mcp_session_id = req.headers["X-Sid"]  # mcp-migrate: ignore[R001] -- proxy shim, not MCP session state
```
```ts
const mcpSessionId = req.headers["x-sid"];  // mcp-migrate: ignore[R001] -- proxy shim
```

## The design question you flagged as the real work

> **Does a suppressed finding still count against the grade?**

**It doesn't.** A suppression that still costs the grade isn't a suppression — the user's only remaining move would be to stop running the tool, which is precisely the outcome this feature exists to prevent.

You also named the cost: the grade becomes partly self-reported. That's real, and burying it would be the wrong trade, so three things make it auditable:

| | |
|---|---|
| **The count always prints** | never behind a flag — it's part of reading the grade honestly |
| **`--show-suppressions`** | lists every one with file, rule and reason |
| **Unused suppressions are reported** | matched nothing → either fixed or the code moved, and stale ones can't silently accumulate |

That's the same standard #101 applied to the stale badge: the claim may be wrong, but it can't be *quietly* wrong. If you'd rather suppressed findings still counted, it's one line in `suppress.apply` — but then I think the feature doesn't do its job.

## Four things deliberately refused

Each because the sloppy version rots:

- **Blanket `ignore` with no rule id** — it would also silence rules written next year. Refused **and reported**, not silently dropped: someone wrote it believing it worked, and the finding is about to appear anyway.
- **A bad rule id** — reported.
- **A missing reason** — reported, but still honoured. Failing closed there would mean a malformed comment silently reintroduces a finding somebody thought was handled.
- **Project-level findings can't be suppressed at all.** R010 asks a question about the whole tree and has no line to attach a comment to. Letting an arbitrary line elsewhere silence it would be a blanket ignore wearing a disguise.

Suppression is both rule-scoped and line-scoped — `ignore[R006]` won't silence an R001 on the same line, and a directive on its own line applies to nothing. Both are tested.

## Compatibility

`run_check` keeps its 5-tuple signature. Widening it breaks callers silently, and several tests unpack exactly five values — the full result moved to `run_check_detailed` / `CheckResult`, and `run_check` delegates.

`--json` gains a **required** `suppressed` array, each entry carrying its `reason`. `schemas/check-json.schema.json` is updated to match so the executable contract stays executable. That's a `Changed` entry for the CHANGELOG — say the word and I'll add it here rather than leaving it to you.

## Tests (24)

More than half assert on **visibility** rather than on suppression working, because that's the half that can be quietly wrong: the count printing without a flag, `--show-suppressions` showing reasons, malformed directives surfacing, unused suppressions surfacing, JSON carrying reasons, suppressed findings absent from `counts`, and the grade/exit-code consequences stated explicitly.

Full suite: **483 passed**.

---

The baseline file (#181) is the other half of this — inline suppression is "this line is fine forever", a baseline is "not today". Happy to take that next if this shape is agreeable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,48 @@ All notable changes to this project are documented here.
 
 ## [Unreleased]
 
+### Added
+
+- **Inline suppression: `# mcp-migrate: ignore[R001] -- reason`.** Silence a
+  single wrong finding without switching a rule off for the whole project.
+  Both comment syntaxes, so it reads naturally in Python and TypeScript.
+  ([#180](https://github.com/dheerajjha/mcp-migrate/issues/180))
+
+  A suppressed finding does not count against the grade. A suppression that
+  still costs you the grade is not a suppression — the only move left would
+  be to stop running the tool. That makes the grade partly self-reported, so
+  it is auditable by construction rather than by convention:
+
+  - the rule id is required; a blanket `ignore` would also silence rules that
+    do not exist yet, and nobody revisits it
+  - a reason is expected, and directives without one are reported
+  - the suppression count prints on every run, never behind a flag
+  - `--show-suppressions` lists each one with its file, rule and reason
+  - directives that matched nothing are reported, so stale ones cannot
+    quietly accumulate
+
+  Malformed directives are reported rather than dropped: someone wrote it
+  believing it worked, and the finding it was meant to silence is about to
+  appear anyway.
+
+  `R005`, `R015` and `R016` report at most one finding per file, so
+  suppressing their line silences those rules for that whole file. Documented
+  in the README; every other rule is genuinely per-line.
+
+### Changed
+
+- **`check --json` gained two required keys, `suppressed` and
+  `unused_suppressions`.** Both are always present, empty array included. A
+  consumer validating against the 0.2.0 schema will reject 0.3.0 output until
+  it is updated; `schemas/check-json.schema.json` is the executable contract.
+  Findings that were suppressed are absent from `findings` and from `counts`,
+  so a consumer that sums `counts` still gets a total matching `findings`.
+
+  `unused_suppressions` carries one entry per rule id (`rule`, `path`, `line`,
+  `reason`). It is in the JSON and not only on the console because stale
+  suppressions accumulate in CI, and CI is exactly the consumer that reads
+  `--json` and never sees a console line.
+
 ### Fixed
 
 - **R004 no longer fires on a wire method name that returns no tools.**
--- a/README.md
+++ b/README.md
@@ -60,13 +60,17 @@ every finding).
 
 `--json` emits one JSON object with these always-present top-level keys:
 `tool`, `version`, `spec`, `path`, `scannable`, `languages`, `grade`,
-`score`, `files_scanned`, `counts`, and `findings`.
+`score`, `files_scanned`, `counts`, `findings`, and `suppressed`.
 
 Conditional keys:
 
 - `is_sdk` and `sdk_reason` only when the tree is a protocol SDK. In that case `grade` and `score` are `null`.
 - `reason` only when `scannable` is `false`. In that case `grade` and `score` are `null`.
 - `grade` and `score` are also `null` when a tree was read but is not gradable, for example TypeScript-only trees.
+
+`suppressed` holds findings silenced by an inline `mcp-migrate: ignore[R0NN]`
+comment; they are excluded from `findings`, from `counts`, and from the grade,
+and each carries the `reason` from its comment.
 
 Each finding always has `rule`, `severity`, `path`, `line`, and `message`.
 `path` and `line` may be `null` for project-level findings. `fix` is present
@@ -77,6 +81,36 @@ The executable contract is
 this shape may change in a breaking way; such changes are documented under
 Changed in [CHANGELOG.md](CHANGELOG.md), so consumers should pin a version
 and watch that section.
+
+### Suppressing a finding
+
+This tool has open false-positive classes and says so above. When a finding
+is wrong, or the code is deliberate and not changing, silence that one line
+rather than the whole rule:
+
+```python
+mcp_session_id = req.headers["X-Sid"]  # mcp-migrate: ignore[R001] -- proxy shim, not MCP session state
+```
+
+```ts
+const mcpSessionId = req.headers["x-sid"];  // mcp-migrate: ignore[R001] -- proxy shim
+```
+
+The rule id is required — a blanket `ignore` would also silence rules that
+don't exist yet, and nobody ever revisits it. A reason after `--` is expected;
+`check` reports directives that lack one.
+
+A suppressed finding **doesn't count against the grade** — a suppression that
+still costs you the grade isn't one, and the only move left would be to stop
+running the tool. That does make the grade partly self-reported, so three
+things keep it auditable:
+
+- the count is always printed, never behind a flag
+- `--show-suppressions` lists every one with its file, rule and reason
+- `check` reports suppressions that matched nothing, so stale ones don't
+  quietly accumulate
+
+`--json` carries them under `suppressed`, each with its `reason`.
 
 Exit codes, so it drops straight into CI:
 
--- a/README.md
+++ b/README.md
@@ -60,7 +60,8 @@ every finding).
 
 `--json` emits one JSON object with these always-present top-level keys:
 `tool`, `version`, `spec`, `path`, `scannable`, `languages`, `grade`,
-`score`, `files_scanned`, `counts`, `findings`, and `suppressed`.
+`score`, `files_scanned`, `counts`, `findings`, `suppressed`, and
+`unused_suppressions`.
 
 Conditional keys:
 
@@ -71,6 +72,12 @@ Conditional keys:
 `suppressed` holds findings silenced by an inline `mcp-migrate: ignore[R0NN]`
 comment; they are excluded from `findings`, from `counts`, and from the grade,
 and each carries the `reason` from its comment.
+
+`unused_suppressions` holds directives that matched no finding — either it was
+fixed or the code moved. Each entry has `rule`, `path`, `line`, and `reason`,
+one per rule id. It is always present, empty array included: stale suppressions
+accumulate in CI, and CI is the consumer that reads `--json` and never sees a
+console line.
 
 Each finding always has `rule`, `severity`, `path`, `line`, and `message`.
 `path` and `line` may be `null` for project-level findings. `fix` is present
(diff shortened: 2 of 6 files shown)


dheerajjha:
Reviewed and verified. **Two things to change, one to decide, then this goes in.**

Rebased onto current `main` locally — clean, **556 tests pass**. I exercised the behaviour rather than reading the diff, because the whole feature is a claim about what *doesn't* get reported.

## What I verified

| | |
|---|---|
| suppressed finding leaves the grade | ✅ Grade A, count printed unprompted |
| `--show-suppressions` lists file, rule, reason | `R001  a.py:3  proxy shim` |
| suppression is rule- **and** line-scoped | ✅ a directive on line 4 does not touch line 3 |
| blanket `ignore` refused **and reported** | ✅ *"needs a rule id … would also silence rules that do not exist yet"* |
| unused suppression reported | ✅ *"a.py:4 ignore[R001] matched nothing — fixed, or the code moved?"* |
| `run_check` still returns 5 values | ✅ `run_check_detailed` / `CheckResult` alongside |
| `--json` validates against the updated schema | ✅ `suppressed` present and required |

You were right to make the count print unconditionally. A suppression that costs the grade isn't a suppression, and the only honest way to buy that is making the self-reporting impossible to miss.

## Ask 1 — the audit story stops at the text output

`unused_suppressions` is **text-only**. It's absent from `--json` and from `schemas/check-json.schema.json`:

```
top-level keys: counts, files_scanned, findings, grade, languages, path,
                scannable, score, spec, suppressed, tool, version
```

That inverts the design. Stale suppressions accumulate in **CI**, not on someone's terminal — and CI is exactly the consumer that reads `--json` and never sees the text. A team could carry a dozen suppressions that stopped matching months ago and have nothing in their pipeline able to notice.

Please add `unused_suppressions` to the JSON document and to the schema, with the same shape as `suppressed` (path, line, rule, reason). Given `suppressed` is already required, making this one required too keeps the contract uniform — and an empty array is the common case, which is the right default to make visible.

## Ask 2 — the CHANGELOG entry

> say the word and I'll add it here rather than leaving it to you

Saying the word. Under `## [Unreleased]`: `### Added` for the feature, and **`### Changed` for the required `suppressed` key**, since that breaks any consumer validating against the old schema. 0.2.0 shipped today with the `location` → `path` + `line` break documented the same way; that entry is the template.

## Decide — one interaction worth documenting

Some rules report only the **first** occurrence per file. R001 is one:

```python
mcp_session_id = req.headers["Mcp-Session-Id"]   # line 3 -- reported
other = req.headers["Mcp-Session-Id"]            # line 4 -- never reported separately
```

Two occurrences, one finding. So suppressing line 3 silences R001 **for the whole file**, and line 4 becomes invisible — not because it was suppressed, but because it was never a separate finding to suppress.

That's not a bug in your code, and I don't think it should block. But the README will tell people suppression is line-scoped, and for these rules it is effectively file-scoped, which is a meaningfully bigger promise than the reader will take from the docs. A sentence in the README saying so is enough. If you'd rather handle it differently, say so — I'd rather you decide than have me pick.

## Timing

Targeted at **0.3.0** alongside #202. Two breaking `--json` shape changes in consecutive same-day releases is more churn than a pre-1.0 contract should ask of consumers, so I'd like a gap after today's 0.2.0.

**Your pre-commit hook is in 0.2.0, live on PyPI now — thanks, that was a strong first PR and this is a stronger second one.**

ankitverma31:
All three done and pushed, rebased onto current `main` — **569 tests pass**. One of them needed correcting rather than doing, details below.

## Ask 1 — `unused_suppressions` in `--json`

Added to the document and to `schemas/check-json.schema.json`, required, in all three JSON shapes (normal, SDK, unscannable). You were right that text-only inverted the design, and the argument that convinced me is the one you made: the audit story has to work for the consumer that *cannot* read the console.

One shape decision worth flagging since it isn't quite what you specified. A directive can name several rules, so instead of one entry with a `rules` array I emit **one entry per rule id** — `rule`, `path`, `line`, `reason` — identical to `suppressed`, so a consumer reads both with the same code path. Nothing is lost expanding them: if the directive matched nothing, every rule id in it is stale.

```
$ mcp-migrate check . --json
"unused_suppressions": [
  { "rule": "R006", "path": "server.py", "line": 3, "reason": "stale" },
  { "rule": "R011", "path": "server.py", "line": 3, "reason": "stale" }
]
```

Made it required for the reason you gave — the empty array is the common case and that's exactly the one worth making visible, since a key that only appears when non-empty forces every consumer to write the same `.get()`.

`tests/test_check_json_schema.py` already validates all three shapes, so a block I'd missed would have failed CI rather than shipped. Four behaviour tests added on top: present-when-empty, not-reported-when-matched, the two-rule expansion, and reason/line/path round-tripping.

## Ask 2 — CHANGELOG

Added, with `### Changed` covering **both** now-required keys rather than only `suppressed` — `unused_suppressions` lands in the same release and breaks a 0.2.0-schema consumer identically, so documenting one and not the other would be the more confusing half-truth. Also noted that suppressed findings are absent from `counts`, so a consumer summing `counts` still matches `findings`.

## Decide — and here I think the example is wrong

Documented, but not for R001, because **R001 isn't one of these rules**. It dedups on `(path, line)`, not per file, so it reports every occurrence separately. Verified before writing anything:

```python
mcp_session_id = req.a        # line 2
other_mcp_session_id = req.b  # line 3
```
```
R001: [('server.py', 2), ('server.py', 3)]
```

Both fire, so suppressing line 2 leaves line 3 reported. (Worth noting the trap I hit reproducing your snippet: `req.headers["Mcp-Session-Id"]` finds nothing at all, because `search_code` skips string literals by design — the identifier spelling is what R001 matches.)

The rules that genuinely do this are **R005, R015 and R016** — the three that guard on `seen_files` and ask a question about a *file* rather than a line ("does this file declare capabilities without `extensions`", "…build a result without `resultType`", "…implement a list/read handler without cache metadata"). Each reports at most one finding per file, on the first matching line, so suppressing that line silences the rule for the whole file. That's now in the README, phrased as you suggested — a sentence, not a behaviour change:

> `R005`, `R015` and `R016` ask a question about a *file* … Suppressing that line therefore silences the rule **for the whole file**, not just that line. Every other rule reports each occurrence separately, and there suppression really is per line. If you suppress one of these three, you are accepting the rule's verdict on that file.

## One gap I found and deliberately did not fix here

`Suppression.used` is a single flag set when **any** of its rule ids matches. So `ignore[R001, R006]` where only R001 ever fires is marked used, and the stale `R006` half is invisible — it won't appear in `unused_suppressions` either.

It's a real hole in the same audit story, but fixing it means per-rule usage tracking, which changes what `unused()` means for the text output too. That felt like it belonged in its own PR rather than widening this one at review stage. Happy to file it, or fold it in if you'd rather it not ship with a known gap — your call.

Thanks for the note on the hook being live in 0.2.0.

dheerajjha:
All three follow-ups are in, and one of them corrects me. Merging.

**On the README caveat: you're right and I was wrong.** I named R001 as the file-scoped case in review. I checked it rather than take it on faith, and R001 reports every occurrence separately — two triggers on lines 2 and 3 both fire, and suppressing line 2 leaves line 3 reported. The rules that really do report once per file are R005, R015 and R016, and `grep -l seen_files src/mcp_migrate/rules/*.py` returns exactly those three. The README now says the accurate thing because you checked instead of implementing what I asked for.

What else I verified:

- **569 tests pass** locally; green on 3.10–3.13.
- **Suppression is line-scoped and the grade moves.** One suppressed R001 and one live one in the same file: `findings` carries only line 3, `suppressed` carries line 2 with its reason attached, and the grade is C/75 — one breaking finding, not two. The suppressed one is genuinely out of `counts` and out of the score, not just hidden from the table.
- **`unused_suppressions` carries `rule`, `path`, `line`, `reason`**, and both it and `suppressed` are `required` in the schema. A directive for a rule that never fires shows up there instead of rotting silently. That was the gap and it's closed properly — CI reads `--json` and never sees a console line, so text-only really did invert the design.
- **A blanket `ignore` is refused**, with a message that explains why rather than just rejecting: *"A blanket ignore would also silence rules that do not exist yet."* The finding still fires. That is what earns the "auditable by construction" in your title.
- **An unknown rule id fails loudly.** `ignore[R099]` silences nothing — R001 still fires on that exact line — and the directive lands in `unused_suppressions`. A typo'd rule id that silently silenced nothing *and* reported nothing would be the worst outcome here, and it doesn't.

The `Changed` entry for the two now-required keys is the right call: a consumer validating against the 0.2.0 schema does break on 0.3.0 output, and that belongs in the section people read before upgrading.

Both of your PRs land today. Thanks.

dheerajjha:
Merged — with a note on *how*, since it isn't a plain squash.

#202 landing an hour earlier conflicted this one: both PRs touched `CHANGELOG.md`, `README.md` and `cli.py`. All three conflicts were additive — two independent entries under a shared `### Added`, two independent README sections, two unrelated helpers in `cli.py` — so both sides are kept in full. I checked the resolution rather than trusting the marker strip: `Unreleased` still reads Added / Changed / Fixed in order, all three entries survive, `cli.py` parses, 589 tests pass.

I merged rather than squashed so **both commits keep your authorship** in the history instead of collapsing into a trailer.

One thing the merge created that neither PR could have caught: **nothing exercised suppression and SARIF together.** Each branch passed on its own. The failure mode there is quiet in the worst way — a finding a maintainer had deliberately silenced would still open a code-scanning alert on their pull request, which is exactly what suppression exists to prevent.

It works (`0450ba7` pins it): a suppressed finding produces no SARIF result, and all three formats agree on what suppression removed — same hits, same exit code, with the suppressed and unused directives still accounted for in `--json` rather than silently gone. That's a test I owed you both, not a defect in either PR.

591 tests, and 5 of your changes are now in. Thanks.

PR #205, opened 2026-08-07
feat: add a pre-commit hook

### LINKED ISSUE #184: Add a pre-commit hook
The cheapest place to catch a spec regression is before it is committed. There is no pre-commit hook, so the earliest anyone finds out is CI -- and there is no GitHub Action either (#100).

## The work

Add `.pre-commit-hooks.yaml` at the repository root so projects can do:

```yaml
repos:
  - repo: https://github.com/dheerajjha/mcp-migrate
    rev: v0.1.4
    hooks:
      - id: mcp-migrate
```

The file itself is short:

```yaml
- id: mcp-migrate
  name: mcp-migrate
  description: Check an MCP server against the 2026-07-28 spec revision
  entry: mcp-migrate check
  language: python
  types_or: [python, ts, tsx, javascript]
  pass_filenames: false
```

## What actually needs thought

- **`pass_filenames: false`.** Several rules are whole-project checks -- R010 asks "does this project implement `server/discover` anywhere?", R015 looks for evidence across a file set. Handing them only the staged files would make them fire wrongly on a partial view. Starting with `false` (scan the project) is the safe default, and the reasoning belongs in a comment in the yaml, because the next person will want to "optimise" it.
- **Exit codes.** `1` is breaking findings, `2` is could-not-check. Pre-commit treats any non-zero as failure, so a project with no scannable files would block every commit. That is wrong and needs handling -- likely the hook should not fail on `2`.
- **Speed.** A hook that takes several seconds gets uninstalled. `scripts/` has a complexity guard but no timing benchmark (filed separately); worth a rough measurement on a large tree before claiming this is usable.
- Document it in the README next to the CI section.

Good first issue if you have used pre-commit before; the yaml is easy and the two decisions above are where the value is.

ankitverma31:
Fixes #184.

```yaml
repos:
  - repo: https://github.com/dheerajjha/mcp-migrate
    rev: v0.1.4
    hooks:
      - id: mcp-migrate
```

The yaml is the easy part. Both things you flagged as the actual work:

## Exit codes

pre-commit treats **any** non-zero exit as a failed hook, and `check` exits `2` for "no readable source in a supported language". Wired directly, the hook blocks every commit in a repository the tool can't read — which is every repository, right up until someone adds their first Python or TypeScript file. A hook that blocks commits for a reason the user can't act on gets uninstalled within the hour, and then it catches nothing ever again.

So `entry` is `mcp-migrate-precommit`, a wrapper whose entire job is mapping `2` → `0`:

| | `check` | hook |
|---|---|---|
| clean | 0 | **0** |
| breaking finding | 1 | **1** — still blocks, that's the point |
| nothing readable | 2 | **0** |

`check` itself is unchanged and still exits 2, because that's the honest answer for a CLI. Only pre-commit's *reading* of it is wrong, so the adjustment lives in the hook. The "could not read this" message still prints — not fatal isn't the same as not shown.

A wrapper rather than a `--exit-zero-on-unscannable` flag on `check`: the flag would be one more thing to get wrong on a command line, and the distinction exists solely because of how pre-commit interprets exit codes.

## `pass_filenames: false`

With the reasoning in a comment **in the yaml**, because — as you predicted — the next person will want to optimise it.

Several rules are whole-project questions. R010 asks whether `server/discover` exists *anywhere*; R015/R016 look for evidence across a file set. Handed only the staged files they answer about a partial tree and fire wrongly: stage one unrelated file and R010 reports a missing handler that exists two files away.

## Speed

You asked for a measurement rather than a claim. 600 files (300 Python + 300 TypeScript), three consecutive runs:

```
run 1: 0.34 s
run 2: 0.31 s
run 3: 0.32 s
```

That's the cost of rescanning the project each time, and it's well under where a hook starts getting uninstalled.

## One bug caught while testing

The wrapper originally decided "is this a subcommand?" by testing for a leading dash. A bare path doesn't start with `-`, so it was parsed as a subcommand name, argparse exited `2`, and the wrapper dutifully turned that into a **pass** — silently disabling the hook it exists to enable. Fixed, and pinned by `test_a_bare_path_is_treated_as_an_argument_not_a_subcommand`.

## Tests (13)

Each exit code · the message still printing · flags passing through · the bare-path regression · and the hook definition itself: that it runs the wrapper rather than `check` directly, that `pass_filenames` stays false, and that the entry point is declared in `pyproject.toml`.

Full suite: **472 passed**. Documented in the README next to the CI section.

---

Note the `rev: v0.1.4` in the README example — it points at a tag that doesn't exist yet, so this wants a release before anyone can actually use it. Say if you'd rather I pinned it to the current tag instead.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,31 @@
+# Hook definitions consumed by projects that add this repo to their
+# .pre-commit-config.yaml. See the "pre-commit" section of README.md.
+
+- id: mcp-migrate
+  name: mcp-migrate
+  description: Check an MCP server against the 2026-07-28 spec revision
+  entry: mcp-migrate-precommit
+  language: python
+  # Gate on the languages the scanner can actually read, so the hook stays
+  # silent in repositories it has nothing to say about.
+  types_or: [python, ts, tsx, javascript, jsx]
+
+  # Scan the project, not the staged files -- and do not "optimise" this
+  # to true.
+  #
+  # Several rules are whole-project questions. R010 asks "does this project
+  # implement server/discover *anywhere*?" and R015/R016 look for evidence
+  # across a file set. Handed only the staged files, they would answer that
+  # question about a partial view of the tree and fire wrongly: stage one
+  # unrelated file and R010 reports a missing server/discover handler that
+  # exists two files away.
+  #
+  # The cost is that the hook re-reads the project each time. That is the
+  # right trade for a scanner whose rules are project-scoped.
+  pass_filenames: false
+
+  # `mcp-migrate-precommit` is a thin wrapper over `check` that maps exit
+  # 2 ("could not check -- no readable source") to 0. pre-commit treats
+  # every non-zero exit as a failure, so without this the hook blocks
+  # every commit in a repository the tool cannot read. Exit 1, a breaking
+  # finding, still fails the hook -- that is the point of installing it.
--- a/README.md
+++ b/README.md
@@ -231,6 +231,30 @@ correct, it leaves the source untouched rather than guess. A wrong fix that
 silently corrupts your server is worse than reporting the finding and doing
 nothing.
 
+## Catch it before it is committed
+
+```yaml
+# .pre-commit-config.yaml
+repos:
+  - repo: https://github.com/dheerajjha/mcp-migrate
+    rev: v0.1.4
+    hooks:
+      - id: mcp-migrate
+```
+
+A breaking finding fails the hook, which is the point. A repository with no
+readable source does **not** — the hook runs `mcp-migrate-precommit`, a thin
+wrapper that maps `check`'s exit `2` ("could not check it") to `0`. pre-commit
+treats every non-zero exit as a failure, so without that the hook would block
+every commit in a repo the tool cannot read, which is a reason the user can do
+nothing about. The message still prints; it just isn't fatal.
+
+The hook scans the **project**, not the staged files (`pass_filenames: false`),
+and that isn't an oversight to optimise away: several rules are whole-project
+questions — R010 asks whether `server/discover` exists *anywhere* — and handed a
+partial view they fire wrongly. Cost of that choice, measured on 600 files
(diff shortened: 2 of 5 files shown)


dheerajjha:
Merging this. **Welcome — you're contributor #17,** and this is an unusually complete first PR.

@waterlemonnn opened #204 for the same issue 39 seconds before you did, so this was a genuine collision. I'm taking yours, for reasons that are about the work rather than the clock:

- **You measured what the issue asked for.** 600 files, three runs, 0.31–0.34s. #204 said "well under a second" and flagged the measurement as a follow-up. The issue specifically asked whether hook latency was acceptable; a number answers that and an impression doesn't.
- **`types_or` includes `jsx`.** #204's list stops at `[python, ts, tsx, javascript]`. A `.jsx` server would have silently skipped the hook.
- **You found a real bug in your own wrapper and pinned it.** Deciding "is this a subcommand?" by testing for a leading dash meant a bare path parsed as a subcommand name, argparse exited 2, and the wrapper turned that into a **pass** — silently disabling the hook it exists to enable. That is the worst possible failure mode for this feature, and it would not have been obvious in review.

I verified the exit-code matrix rather than taking the table on faith:

```
                wrapper   check
unscannable        0        2
breaking           1        1
clean              0        0
```

`check` is untouched, which is the right call — 2 is the honest answer for a CLI, and only pre-commit's reading of it is wrong.

## One correction

> Note the `rev: v0.1.4` in the README example — it points at a tag that doesn't exist yet

It does exist, as both a tag and a [release](https://github.com/dheerajjha/mcp-migrate/releases). The real problem is subtler and you were right to flag it: **v0.1.4 predates `.pre-commit-hooks.yaml`**, so pre-commit would clone that tag successfully and find no hook to run. Same broken outcome, different cause.

Fixed by pointing at `v0.2.0`, which I am cutting immediately after this merges — so the example works the moment it is published rather than being aspirational.

**522 tests pass.**

@waterlemonnn's `--allow-unscannable` argument is worth keeping in view: a flag on `check` helps anyone piping it through something that treats non-zero as failure, not just pre-commit. That is a real use case your wrapper does not cover, and it can land separately without conflicting with this.

PR #222, opened 2026-08-08
Show suppression counts in the registry

### LINKED ISSUE #220: A suppressed grade reaches the board with no record that anything was suppressed
Inline suppression shipped in #203: a suppressed finding does not count against the grade, and the tool reports every suppression on the console and in `--json`.

**None of that reaches the board.**

```
$ grep -rn suppress scripts/
$ grep -n suppress registry/schema.yaml
```

Both empty. `render_badges.py`, `render_board.py`, `validate_registry.py` and `ecosystem_scan.py` never read `suppressed` or `unused_suppressions`, and `registry/schema.yaml` has no field that could hold them.

## What that allows today

1. Suppress some findings.
2. Take the improved grade.
3. `mcp-migrate entry --repo owner/name`.
4. Submit the entry.

At no point does any artifact record that suppressions were involved — not the YAML, not the badge, not the board row. The grade is presented identically to one reached with no suppressions at all.

This is not hypothetical dishonesty on anyone's part; it's that the tool gives an honest local answer and then drops the caveat on the floor at exactly the boundary where the number stops being local. The board's whole claim is that its grades are reproducible (#92) — a grade you cannot reproduce without also knowing which findings were silenced does not meet that bar.

## Scope

- `mcp-migrate entry` should emit the suppression count into the generated YAML when a scan had any, rather than silently omitting it.
- `registry/schema.yaml` needs a field for it. Optional, integer, defaulting to absent — an entry with no suppressions should look exactly as it does now, so this doesn't churn the 16 existing files.
- `validate_registry.py` should accept it; `render_board.py` should show it on rows that have it. What that looks like is open — a footnote, a column, a marker.
- A test that a suppressed scan produces an entry carrying the count.

## Explicitly not in scope

Whether the **badge** should change when suppressions are active. That's the genuinely contested half and it's in [discussion #210](https://github.com/dheerajjha/mcp-migrate/discussions/210) — please argue it there rather than deciding it inside this PR. This issue is only the part I think has an obvious answer: the registry should record what the grade was computed from.

Good first issue if you want one with a clear boundary: the schema change and `entry` change are small, and the board rendering has room for judgment.

soltonigiri:
## Summary

- Add an optional `suppressed` count to generated registry entries when non-zero.
- Validate the field and show the count on the registry board.

## Tests

- `uv run pytest -q`
- `uv run python scripts/validate_registry.py --offline`

Closes #220

### DIFF
--- a/registry/schema.yaml
+++ b/registry/schema.yaml
@@ -4,6 +4,7 @@ repo:          {required: true,  type: str,  desc: "owner/name on GitHub"}
 language:      {required: true,  type: str,  enum: [python, typescript, go, rust, java, csharp, ruby, other]}
 grade:         {required: true,  type: str,  enum: [A, B, C, D, F]}
 score:         {required: true,  type: int,  range: [0, 100]}
+suppressed:    {required: false, type: int,  desc: "non-negative count of findings suppressed when the grade was computed"}
 checked_with:  {required: true,  type: str,  desc: "mcp-migrate <version>"}
 spec:          {required: true,  type: str,  desc: "spec revision checked against"}
 status:        {required: true,  type: str,  enum: [ready, migrating, unmaintained]}
--- a/scripts/render_board.py
+++ b/scripts/render_board.py
@@ -28,8 +28,9 @@ def main() -> int:
     for e in entries:
         repo = e.get("repo", "")
         notes = str(e.get("notes", "")).strip().replace("|", "\\|")
+        suppression_note = f" ({e['suppressed']} suppressed)" if e.get("suppressed", 0) > 0 else ""
         rows.append(
-            f"| [{e.get('name')}](https://github.com/{repo}) | **{e.get('grade')}** "
+            f"| [{e.get('name')}](https://github.com/{repo}) | **{e.get('grade')}**{suppression_note} "
             f"| {EMOJI.get(e.get('status'), e.get('status'))} | {e.get('language')} | {notes} |"
         )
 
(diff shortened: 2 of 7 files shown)


dheerajjha:
Merged — and welcome, this is your first change here and it's a good one.

I verified it past the diff rather than just reading it:

- A suppressed scan round-trips **`entry` → `validate_registry`** with no errors, producing `suppressed: 1` in the right place.
- An unsuppressed scan omits the key entirely.
- Re-rendering the board over the existing sixteen entries is **byte-identical** — no churn, which was the constraint in the issue and is easy to get wrong.
- 610 tests pass.

**The `isinstance(bool)` guard is the detail most people miss.** `True` is an `int` in Python, so `suppressed: true` would have validated as 1 without it. You caught that *and* wrote the test for it. That's the difference between a validator and a validator that works.

Two things I added rather than sending back, since neither is worth a round trip on a first PR:

**1. A crash in `render_board.py`.** `suppressed:` written with no value parses as `None`, and `None > 0` raises `TypeError`. `validate_registry` rejects such an entry — but `render_board.py` runs independently of it, so someone hand-editing a YAML and rendering before validating got a traceback instead of a board. Changed to `(e.get("suppressed") or 0) > 0` with a test. Worth knowing as a general shape: `.get(k, default)` doesn't protect you when the key exists and the *value* is `None`, which is exactly what YAML does with a bare `key:`.

**2. A CHANGELOG entry**, which the PR was missing. Every user-visible change here gets one; I wrote it this time and credited you.

For context on why this mattered: suppression deliberately doesn't cost the grade, and that was fine while the number lived in a terminal. It stopped being fine the moment the same grade could reach a public board with nothing recording that findings were silenced — the board's entire claim is that its grades are reproducible. Your change is what makes that claim survive the suppression feature.

The badge half is still open and genuinely contested — [discussion #210](https://github.com/dheerajjha/mcp-migrate/discussions/210) if you want to argue a position. Nobody has made the case for the strictest option properly yet.

PR #226, opened 2026-08-08
fix: consolidate grade→color map into shared constants

### LINKED ISSUE #224: Two grade→colour maps disagree: a B is green from the CLI and brightgreen on the board, where A and B are indistinguishable
Following on from #211, where I was trying to work out how to read the letter.
Turns out the *colour* has a related problem, and this one is live on the board
right now.

## There are two grade→colour maps and they disagree

`src/mcp_migrate/grade.py`, used by `badge_url()`, which is what `check` prints
for you to paste (`cli.py:575`):

```python
BADGE_COLOR = {"A": "brightgreen", "B": "green", "C": "yellow", "D": "orange", "F": "red"}
```

`scripts/render_badges.py`, used to generate the live endpoints under
`docs/badge/`:

```python
GRADE_COLOR = {"A": "brightgreen", "B": "brightgreen", "C": "yellow", "D": "red", "F": "red"}
```

They agree on A, C and F, and disagree on **B** and **D**.

## What that does to a real project

A project graded B gets told by the CLI to paste a badge that renders
**green**. Its entry on the board renders **brightgreen**. Same project, same
grade, same day, two different colours depending on which of the two published
paths you came through.

## The part I think matters more

On the live board, **A and B are the same colour**. Counted across the
committed endpoints in `docs/badge/`:

```
  8  B  brightgreen
  6  A  brightgreen
  1  C  yellow
  1  D  red
  1  not listed  lightgrey
```

Fourteen of the sixteen entries are A or B, and nothing in the colour
distinguishes them. `D` and `F` collapse the same way, both `red`.

So on the board the colour carries about one bit — "green, yellow or red" —
while the letter carries five values. Which reads oddly against the comment
sitting directly above `GRADE_COLOR`:

> a C rendered in green teaches the reader that the colour means nothing, and
> then the next badge is not trusted either

I'd have said collapsing B into A's colour does the same thing from the other
direction: it makes a B look like a clean bill of health, and it is the more
flattering error of the two, which is the one this project is usually careful
to avoid.

## Not sure which one is right

It may well be deliberate that the board uses a three-colour scheme — that's a
defensible call, traffic-light schemes are easier to scan in a table. But then
`badge_url()` should use the same one, and right now the tool tells a B-graded
maintainer to paste a colour their own board wouldn't give them.

The minimal fix is to have one map and import it in both places, whichever
scheme wins. If the three-colour scheme is the intended one for the board and
five is intended for the pasted badge, that's worth a comment saying so,
because it currently reads as drift rather than a decision.

Happy to send a PR once you've said which way it should go — I didn't want to
pick, since the answer changes 15 of the committed endpoints either way.

Jah-yee:
## Summary

Both `grade.py` (`badge_url`) and `render_badges.py` (live board) defined their own `GRADE_COLOR` map independently, and they disagreed on B and D:

| Grade | `grade.py` | `render_badges.py` |
|-------|-------------|---------------------|
| B | green | brightgreen |
| D | orange | red |

A B-graded project was told to paste a 'green' badge while its entry on the live board showed 'brightgreen'. On the board A and B were indistinguishable (both brightgreen) and D and F were indistinguishable (both red), collapsing a 5-value signal into a 3-value one.

## Fix

Consolidate into a single canonical map in `src/mcp_migrate/constants.py` and import it from both `grade.py` and `render_badges.py`.  The five-step traffic-light is preserved: each grade keeps its own colour so the signal is maximally informative.

## Changes

- **`src/mcp_migrate/constants.py`**: new file with the canonical `GRADE_COLOR` and `UNKNOWN_COLOR` constants
- **`src/mcp_migrate/grade.py`**: import `GRADE_COLOR` from constants; remove local `BADGE_COLOR` definition
- **`scripts/render_badges.py`**: import `GRADE_COLOR` and `UNKNOWN_COLOR` from constants; add `sys.path` manipulation to locate `src`
- **`tests/test_badges.py`**: update test expectations to match the five-step colour scheme

Closes #224.

### DIFF
--- a/scripts/render_badges.py
+++ b/scripts/render_badges.py
@@ -28,22 +28,14 @@ from pathlib import Path
 import yaml
 
 ROOT = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(ROOT / "src"))
+
 SERVERS = ROOT / "registry" / "servers"
 OUT = ROOT / "docs" / "badge"
 
 LABEL = "MCP 2026-07-28"
 
-# Must agree with how the grades read elsewhere, or the badge undermines
-# the thing it is reporting: a C rendered in green teaches the reader that
-# the colour means nothing, and then the next badge is not trusted either.
-GRADE_COLOR = {
-    "A": "brightgreen",
-    "B": "brightgreen",
-    "C": "yellow",
-    "D": "red",
-    "F": "red",
-}
-UNKNOWN_COLOR = "lightgrey"
+from mcp_migrate.constants import GRADE_COLOR, UNKNOWN_COLOR  # noqa: E402,F401
 
 # A 404 renders as a broken image on a stranger's README. Anyone can point
 # an endpoint URL at us -- including at a repo we have never scanned -- so
--- a/src/mcp_migrate/constants.py
+++ b/src/mcp_migrate/constants.py
@@ -1,0 +1,20 @@
+"""Shared constants used across the CLI and the badge renderer.
+
+The grade→color map must agree everywhere it is used, or the badge
+undermines the thing it is reporting: a C rendered in green teaches the
+reader that the colour means nothing, and then the next badge is not
+trusted either.
+"""
+
+# Five-step traffic-light: each grade has a distinct color.
+# This is the canonical map; both cli.py (badge_url) and
+# scripts/render_badges.py import from here.
+GRADE_COLOR = {
+    "A": "brightgreen",
+    "B": "green",
+    "C": "yellow",
+    "D": "orange",
+    "F": "red",
+}
+
+UNKNOWN_COLOR = "lightgrey"
--- a/src/mcp_migrate/grade.py
+++ b/src/mcp_migrate/grade.py
@@ -1,6 +1,7 @@
 """Turn findings into a single letter you can put in a README badge."""
 from __future__ import annotations
 
+from .constants import GRADE_COLOR
 from .rules.base import Finding, Rule
 
 WEIGHT = {"breaking": 25, "deprecated": 8, "advisory": 3}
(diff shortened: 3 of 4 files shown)


dheerajjha:
Verified and merging. Thank you — and welcome, this is your first PR here.

I reproduced the disagreement on \`main\` before reading the patch:

\`\`\`
grade.py          B: green        D: orange
render_badges.py  B: brightgreen  D: red
\`\`\`

So a B-graded maintainer really was being handed a \`green\` badge URL while their board row rendered \`brightgreen\`, and on the board A/B and D/F were each a single colour. Your table is exactly right.

What settled the direction for me is that `render_badges.py` already carried this comment above its map:

> Must agree with how the grades read elsewhere, or the badge undermines the thing it is reporting.

The intent was already written down — the code just never honoured it. Consolidating to one map in `constants.py` is what that comment was always asking for, and keeping the five-step scale is the reading its rationale supports: collapsing D into F tells a D-graded server it is as bad as an F, which is the same lie in the other direction.

One thing I checked specifically, because it would have broken the board silently: `.github/workflows/board.yml` installs **only** `pyyaml` and never installs the package, so `from mcp_migrate.constants import GRADE_COLOR` had to work with `mcp_migrate` not importable. Your `sys.path` shim handles it — I ran the script in a fresh venv with only pyyaml and no package installed, and it wrote all 27 endpoints cleanly.

Full suite: 610 passed.

Following up in a separate commit rather than asking you to rebase, since neither is really your PR's job:

1. The regenerated `docs/badge/*.json` — 9 of 16 change colour, and the PR doesn't include them.
2. That Board workflow only triggers on `paths: registry/**`, so it would never have re-run for a change to the renderer itself. The committed badges would have sat disagreeing with your new map indefinitely — the same bug you just fixed, one layer out.

Issue #239, opened 2026-08-10
A well-formed but nonexistent rule id is dropped from config in silence, and the README says it isn't


dheerajjha:
A rule id that is well-formed but doesn't exist is dropped from config in silence — and the README says it isn't

## Reproduce

```toml
[tool.mcp-migrate.rules]
R999 = "off -- typo, meant R010"
```
```
score 97   disabled_rules: []   config_warnings: []
```

Nothing. Compare a *malformed* id in the same place, which is handled properly:

```toml
[tool.mcp-migrate.rules]
R10 = "off"
```
```
config_warnings: ["pyproject.toml: not a rule id, ignoring [rules] entry 'R10'"]
```

## This contradicts what we promise

README.md:283-286:

> Malformed config (bad TOML, **an unrecognised rule id**, `rules.R008 = "sometimes"`)
> doesn't fail the run -- it falls back to defaults and **reports what it
> couldn't understand**, under `config_warnings` in `--json` output and as a
> `config warning` line in text output.

`R999` is an unrecognised rule id by any ordinary reading of that sentence, and
it is reported nowhere.

## Cause

`RULE_ID_RX` is `^R\d{3}$` (`suppress.py:63`) — a check on the *shape* of the id.
`_parse_table` (`config.py:90`) applies it and nothing else, so `R999` is accepted
into `disabled_rules`. Whatever filters against the real rule set downstream drops
it without comment. Shape is validated; existence never is.

The same hole exists for inline `mcp-migrate: ignore[R999]` suppressions, which
share `RULE_ID_RX`. Worth confirming and fixing in the same pass rather than
leaving the two halves inconsistent.

## Why it matters

The whole point of putting a rule toggle in config is that it is written once
and then nobody looks at it again. A typo'd id leaves the rule **on**, and the
only symptom is a grade slightly lower than the team expected — which reads as
"we have work to do", not as "our config is broken". It is a silent no-op with
a plausible-looking cause, which is the hardest kind to ever notice.

## Fix

Validate configured ids against the registered rules and warn on the rest.
The design question is where, because `config.py` does not currently import the
rule registry and probably should not — a config parser that depends on the
whole rule set is a heavier module than it needs to be, and there is an import
cycle risk worth checking before committing to it.

Two options:

1. Keep `config.py` pure and warn at the point where `run_check_detailed`
   filters `disabled_rules` against `all_rules()`, where both are already in
   hand. Cheapest, and the warning still reaches `config_warnings`.
[...]

Issue #214, opened 2026-08-07
RULE_CAP and WEIGHT were picked independently: breaking is the most aggressively capped severity


dheerajjha:
Raised in #211. `WEIGHT` was chosen on its own terms and `RULE_CAP` was picked
later as round numbers that looked severe enough; the ratio between them was
never considered. Expressed as "how many firings until a rule hits its cap",
the result is inverted:

| severity | weight | cap | firings to cap |
| --- | --- | --- | --- |
| breaking | 25 | 25 | 1 |
| deprecated | 8 | 12 | 1.5 |
| advisory | 3 | 6 | 2 |

So `breaking` is the most aggressively capped of the three: the second
occurrence of a breaking finding is already free, while an advisory gets two
before it stops counting. Concretely, a server with 14 breaking R001 sites and
one stray advisory is penalised more for the advisory's second hit than for
R001's thirteenth. That is not a position anyone holds — it is an artifact.

The best available defence of the current numbers is that repetition of a
systemic breaking finding carries the least new information, so capping it
hardest is right. That doesn't survive the counterexample: if repetition
carries little information for `breaking`, it carries less for `advisory`, and
the table has that backwards.

Preferred fix: pick a firings-to-cap multiple and derive the caps from the
weights, rather than hand-tuning three more round numbers. A flat 2x would give
`breaking` 50, `deprecated` 16, `advisory` 6 (unchanged).

This moves published grades, so it is not a patch-release change:

- [ ] agree the multiple
- [ ] re-verify the 14 board entries against the new numbers, and regenerate
- [ ] `Changed` entry in CHANGELOG.md

dheerajjha:
Closed by #227 (@waterlemonnn), landed with `CAP_MULTIPLE = 1` rather than the 2 the PR proposed.

The inversion this issue reported is gone — firings-to-cap is now 1.0 for every severity, and `RULE_CAP` is derived from `WEIGHT` rather than hand-picked, so the two cannot drift apart again.

On the multiple: all 16 board servers were re-scanned at 1x, 2x and 3x. Every multiple fixes the inversion equally; what it actually selects is how harshly we grade. 14 servers do not move at any setting. At 2x the other two would each have dropped two letters to F — on code that did not change, only our arithmetic — so 1x it is. `mcp-atlassian` 60 → 64, `mcp-server-git` 54 → 58, both keeping their letters. Full table in the CHANGELOG.

[...]

PR #241, opened 2026-08-10
docs: add getting-started guide and full rule reference

sinhphamvan:
The `docs/` directory currently only ships `badge/` and `banner.jpg`; everything is crammed into the README. These two pages move the reference material out so the README can stay a pitch and the docs can be linked, embedded, and translated.

**`docs/GETTING_STARTED.md`** — install via `uvx`, `check` and its report flags, exit codes, `fix` semantics (`safe` vs `review`, dry-run default), config precedence, suppressions and their three file-scoped rule caveats, pre-commit/CI/SARIF integration, and the Python/TypeScript grading situation.

**`docs/RULES_REFERENCE.md`** — the 21-rule table (severity, spec change, fixer confidence) plus the grading caps and score-to-letter mapping.

Content is drawn from the current README and `mcp-migrate rules` output; no behavior changes.

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -1,0 +1,175 @@
+# Getting Started
+
+`mcp-migrate` finds and fixes what the MCP 2026-07-28 spec revision breaks in
+your server. It is the only tool that edits your server's code for you: the
+official Python SDK ships no codemod for the protocol changes, and the
+TypeScript codemod only handles the v1→v2 package rename.
+
+## Install
+
+`mcp-migrate` is a Python CLI published on PyPI. The fastest way to run it
+without installing anything into your project is `uvx`:
+
+```bash
+uvx mcp-migrate --help
+```
+
+Alternatively, install it into your environment:
+
+```bash
+pip install mcp-migrate
+# or
+uv tool install mcp-migrate
+```
+
+## Check a server
+
+Point the tool at the root of your server project:
+
+```bash
+uvx mcp-migrate check .
+```
+
+The report lists every finding with its rule id, severity (`breaking`,
+`deprecated`, `advisory`), file, and line, then prints a letter grade and a
+score out of 100. A clean tree prints `Grade A` and a ready-to-paste badge.
+
+Useful report flags:
+
+- `--json` — machine-readable output with the full, uncapped finding list
+  (the terminal table caps each rule at 5 rows).
+- `--format sarif` — SARIF 2.1.0 for GitHub code scanning.
+- `--rule R001` — run only specific rules (repeatable). The grade is
+  suppressed when a subset of rules runs, because a partial grade is not a
+  grade.
+- `--severity breaking` — only change what is *displayed*; the grade and
+  exit code are still computed from every finding.
+- `--fail-on {breaking,deprecated,advisory,never}` — the minimum severity
+  that fails the run (default `breaking`).
+- `--include-tests` — scan tests, fixtures, and examples (skipped by
+  default: backward-compat tests are evidence of good testing, not of a
+  broken server).
+
+### Exit codes
+
+| code | meaning |
+| ---- | ------- |
+| `0`  | checked it, nothing `breaking` |
+| `1`  | checked it, found something `breaking` |
+| `2`  | **could not check it** — no readable source in a supported language |
+
+Exit `2` means "we did not read your code", not "your code is fine". An
+empty finding set is only meaningful when the tool actually scanned the tree.
+
+## Fix a server
+
+`fix` runs in **dry-run mode by default** — it prints the exact unified diff
+and writes nothing:
+
+```bash
+uvx mcp-migrate fix .            # preview only
+uvx mcp-migrate fix . --write    # apply the changes
+```
+
+Every change is tagged in the diff output:
+
+- **`safe`** — the transformation cannot change behavior. Apply with
+  confidence.
+- **`review`** — the fixer did the mechanical part it is sure of and left a
+  `# TODO(mcp-migrate): ...` where a human must finish the job.
+
+Other fix flags:
+
+- `--safe-only` — apply only `safe` fixers.
+- `--rule R006` — restrict to specific rules (repeatable, same as `check`).
+- `--include-tests` — also fix test/fixture paths.
+
+Fixers are deliberately conservative: when a fixer cannot be certain a
+transformation is correct, it leaves the source untouched rather than guess.
+A wrong fix that silently corrupts your server is worse than reporting the
+finding and doing nothing.
+
+## Configure it
+
+Everything that is a flag can live in config so a team and CI share it.
+Put it in `[tool.mcp-migrate]` in `pyproject.toml`:
+
+```toml
+[tool.mcp-migrate]
+skip = ["vendor/", "generated/"]
+include-tests = false
+
+[tool.mcp-migrate.rules]
+R008 = "off"                                          # no reason recorded
+R016 = "off -- ttl is enforced by the gateway in front of this"
+```
+
+No `pyproject.toml`? Use a standalone `.mcp-migrate.toml` at the project
+root instead (most TypeScript projects fall here).
+
+Rules:
+
+- **A flag beats config, config beats the built-in default.**
+- **A disabled rule never runs** — it costs nothing against the grade and
+  `check` reports how many rules config switched off in every run.
+- **Malformed config** (bad TOML, unknown rule id, invalid rule value) does
+  not fail the run — it falls back to defaults and reports what it could
+  not understand (`config_warnings` in `--json`).
+
+## Suppress a false positive
+
+When a finding is wrong, or the code is deliberate and not changing, silence
+that one line rather than the whole rule:
+
+```python
+mcp_session_id = req.headers["X-Sid"]  # mcp-migrate: ignore[R001] -- proxy shim, not MCP session state
+```
+
+The rule id is required and a reason after `--` is expected. Suppressed
+findings **don't count against the grade**. `--show-suppressions` lists every
+one; stale suppressions that matched nothing are reported as
+`unused_suppressions`, so they don't accumulate silently in CI.
+
+**Caveat:** `R005`, `R015`, and `R016` report at most one finding per file,
+so suppressing their line silences the rule for the whole file.
+
+## Integrate with CI and pre-commit
+
+Pre-commit hook (a `breaking` finding fails the hook):
+
+```yaml
+# .pre-commit-config.yaml
+repos:
+  - repo: https://github.com/dheerajjha/mcp-migrate
+    rev: v0.2.0
+    hooks:
+      - id: mcp-migrate
+```
+
+The hook maps exit `2` ("could not check it") to `0` so a repo the tool
+cannot read does not block every commit.
+
+GitHub code scanning via SARIF:
+
+```yaml
+- run: mcp-migrate check . --format sarif > mcp-migrate.sarif
+- uses: github/codeql-action/upload-sarif@v3
+  with:
+    sarif_file: mcp-migrate.sarif
+```
+
+Severities map to SARIF levels: `breaking` -> `error`, `deprecated` ->
+`warning`, `advisory` -> `note`.
+
+## Understand the language support
+
+- **Python is graded.** TypeScript is scanned — every rule reads TypeScript —
+  but a letter grade is withheld pending the decision tracked in
+  [dheerajjha/mcp-migrate#172](https://github.com/dheerajjha/mcp-migrate/issues/172).
+- The exit code works on TypeScript regardless: a `breaking` finding exits
+  `1` with or without a grade, so TypeScript CI gates work today.
+
+## Next steps
+
+- See the full rule reference in [RULES_REFERENCE.md](RULES_REFERENCE.md).
+- Add your server to the board with `mcp-migrate entry --repo owner/name`.
--- a/docs/RULES_REFERENCE.md
+++ b/docs/RULES_REFERENCE.md
@@ -1,0 +1,61 @@
+# Rule Reference
+
+Every rule shipped by `mcp-migrate`, with its severity, what the 2026-07-28
+spec change breaks, and whether a fixer exists. Run `mcp-migrate rules` for
+the exact list of your installed version and `mcp-migrate fixers` for the
+fixer confidence table.
+
+## The 21 rules
+
+| Rule | Severity | What breaks | Fixer |
+| ---- | -------- | ------------ | ----- |
+| [R001](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r001_session_id_removed.py) | breaking | `Mcp-Session-Id` is gone from the Streamable HTTP transport ([SEP-2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567)). | yes (`review`) |
+| [R002](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r002_connection_state.py) | breaking | Servers are required to be stateless ([SEP-2567](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2567)); a module-level dict keyed by connection breaks behind a load balancer or a restart. | yes (`review`) |
+| [R003](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r003_routing_headers.py) | advisory | Hand-rolled HTTP clients that skip the new required `Mcp-Method`/`Mcp-Name` [routing headers](https://modelcontextprotocol.io/specification/2026-07-28/basic/transports/streamable-http) get rejected by anything enforcing the new transport. | yes (`review`) |
+| [R004](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r004_tool_ordering.py) | advisory | [`tools/list` order](https://modelcontextprotocol.io/specification/draft/changelog) is not guaranteed; non-deterministic ordering defeats caching. | yes (`safe`) |
+| [R005](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r005_extensions.py) | advisory | Optional capabilities negotiate through an [`extensions`](https://modelcontextprotocol.io/specification/draft/changelog) map that isn't declared. | yes (`safe`) |
+| [R006](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r006_sse_transport_deprecated.py) | deprecated | [HTTP+SSE](https://modelcontextprotocol.io/specification/draft/changelog) is deprecated in favor of Streamable HTTP; stays in the spec 12+ months, then leaves. | yes (`review`) |
+| [R007](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r007_deprecated_features.py) | deprecated | [Roots, Sampling and Logging](https://modelcontextprotocol.io/specification/draft/changelog) are deprecated as core capabilities. | yes (`review`) |
+| [R008](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r008_trace_context.py) | advisory | Trace context (`traceparent`, `tracestate`, `baggage`) now travels in `_meta` ([SEP-414](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414)); OpenTelemetry breaks at your server if it's never read. | yes (`review`) |
+| [R009](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r009_initialize_handshake_removed.py) | breaking | The `initialize`/`notifications/initialized` handshake ([SEP-2575](https://modelcontextprotocol.io/specification/2026-07-28/changelog)) is gone; a server still implementing it never becomes usable to a 2026-07-28 client. | yes (`review`) |
+| [R010](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r010_server_discover_missing.py) | advisory | Servers must implement [`server/discover`](https://modelcontextprotocol.io/specification/2026-07-28/changelog) ([SEP-2575](https://modelcontextprotocol.io/specification/2026-07-28/changelog)); registering handlers without it leaves clients with no way to learn what you support. | no |
+| [R011](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r011_ping_removed.py) | breaking | `ping` ([SEP-2575](https://modelcontextprotocol.io/specification/2026-07-28/changelog)) is removed from the protocol; liveness rides on the transport now. | yes (`review`) |
+| [R012](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r012_logging_set_level_removed.py) | breaking | `logging/setLevel` ([SEP-2575](https://modelcontextprotocol.io/specification/2026-07-28/changelog)) is removed; log level is per-request via `_meta` now. | yes (`review`) |
+| [R013](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r013_subscriptions_replaced.py) | breaking | `resources/subscribe`/`resources/unsubscribe` ([SEP-2575](https://modelcontextprotocol.io/specification/2026-07-28/changelog)) are replaced by `subscriptions/listen`. | yes (`review`) |
+| [R014](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r014_sse_resumability_removed.py) | breaking | SSE resumability via `Last-Event-ID` ([SEP-2575](https://modelcontextprotocol.io/specification/2026-07-28/changelog)) is removed; a dropped connection is just dropped now. | yes (`review`) |
+| [R015](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r015_result_type_required.py) | advisory | Every result now requires `resultType` ([SEP-2322](https://modelcontextprotocol.io/specification/2026-07-28/changelog)). Fires only on servers that build their own JSON-RPC envelopes. | no |
+| [R016](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r016_cacheable_result_required.py) | advisory | List/read results require `ttlMs`/`cacheScope` ([SEP-2549](https://modelcontextprotocol.io/specification/2026-07-28/changelog)). Advisory: nothing has adopted the new API yet. | yes (`review`) |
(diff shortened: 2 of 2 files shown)


dheerajjha:
Good idea and mostly accurate — I checked the claims rather than the prose,
because a docs page that is wrong is worse than no docs page. Most of it holds
up. Three things need changing before I merge, and two of them are my fault
rather than yours.

## What I verified as correct

- SARIF mapping `breaking -> error`, `deprecated -> warning`, `advisory -> note`
  (matches `sarif.py:42-44`)
- the pre-commit hook mapping exit 2 to 0 (`precommit.py:43`)
- `unused_suppressions` being the actual `--json` key
- `--severity` not moving the exit code — I ran it, still exit 1 with
  `--severity advisory` when a breaking finding is present, exactly as you say
- `fix` printing a real unified diff, dry-run by default
- R005/R015/R016 being the three file-level rules
- "19 of 21 rules ship a fixer (R010 and R015 do not)"
- every rule id, severity and fixer-confidence cell in `RULES_REFERENCE.md`

That is a lot of surface to get right. Thank you for not guessing at any of it.

## 1. The cap table is wrong, and you got it from me

`RULES_REFERENCE.md` says the per-rule cap is -12 for `deprecated` and -6 for
`advisory`. It is -8 and -3. The caps are derived — `RULE_CAP = WEIGHT *
CAP_MULTIPLE` with `CAP_MULTIPLE = 1` — so each cap simply equals its cost, and
a rule costs the same whether it fires once or twenty times:

```
deprecated x1 -> 92    x20 -> 92
advisory   x1 -> 97    x20 -> 97
```

You copied this faithfully from README, which had been wrong since the multiple
changed. I have fixed README on main (ea6d5b9) and added a test that reads
`RULE_CAP` instead of restating it, so it cannot drift again. Please pull and
take the corrected numbers.

## 2. `rev: v0.2.0` — same story

Two releases behind; current is v0.4.0. Also copied from README, also now fixed
and now tested against `__version__`.

## 3. One claim is true only after #243 lands

> **Malformed config** (bad TOML, unknown rule id, invalid rule value) does not
> fail the run — it falls back to defaults and reports what it could not
> understand

An unknown-but-well-formed rule id (`R999`) is currently accepted in silence —
that is #239, and #243 fixes it. README makes the same claim and has the same
problem, which is how I found it. No change needed on your side; just be aware
the sentence is a promise the code keeps only once #243 merges, and I would
rather merge these in that order than publish a new page asserting it today.

## Two smaller things, your call

- The exit-code table reads `0` = "nothing breaking" and `1` = "found something
  breaking". That is the default, but `--fail-on` is documented four lines above
  and changes it — `--fail-on never` exits 0 with breaking findings. Worth
  qualifying with "at the default `--fail-on breaking`" so the two sections do
  not contradict each other.
- Consider whether `RULES_REFERENCE.md` should exist as prose at all, given
  `mcp-migrate rules` prints the same table from the code and cannot drift. I am
  not asking you to drop it — a linkable page has real value for people deciding
  whether to install — but the file now needs a test like the README tables have,
  or it becomes the next thing to quietly go stale. If you would rather not write
  that test, say so and I will add it after merge.

Everything else is ready. Pull main, fix the two numbers and the `rev`, and I
will merge.

dheerajjha:
Not chasing you — one of the three things I flagged has gone away on its own,
so there is less to do than my review implied.

**Item 3 is resolved.** #243 merged this morning, so an unknown rule id in
config now produces a warning. This sentence in `GETTING_STARTED.md` is
accurate as written and needs no change:

> **Malformed config** (bad TOML, unknown rule id, invalid rule value) does
> not fail the run — it falls back to defaults and reports what it could not
> understand

Still needed before merge, both from README rather than from you:

1. the cap table — `deprecated` is **-8** and `advisory` is **-3**, not -12/-6
2. `rev: v0.4.0`, not `v0.2.0`

Both are fixed on main now, so `git pull` and copy across.

One addition since my review, which affects a line in your page. `fix` can now
exit **1** — #244 landed, and a fix that would leave a Python file unparseable
is refused rather than written. Your exit-code table is under "Check a server"
so it is not strictly wrong, but the `fix` section says nothing about exit
codes at all and now probably should, given CI will start seeing a 1 it never
saw before. A sentence is enough.

No rush on any of this.

dheerajjha:
@sinhphamvan — checking in, and I re-verified my asks against main before sending this so you aren't chasing something that has since moved.

Both still stand, both still one-line copies from README:

1. cap table — `deprecated` is **-8**, `advisory` is **-3** (`grade.py`: `WEIGHT = {"breaking": 25, "deprecated": 8, "advisory": 3}`, `CAP_MULTIPLE = 1`)
2. `rev: v0.4.0` (README line 407, and `pyproject.toml` says `version = "0.4.0"`)

And the third, optional one: a sentence in the `fix` section noting it can now exit 1, since #244 made `fix` refuse a change that would leave a Python file unparseable.

These pages are genuinely worth having and the research in them was good — I'd rather land them than let them age out. If you'd prefer, say the word and I'll push those three edits to your branch myself and merge; I won't touch your branch without you asking. No pressure either way, and no deadline.

dheerajjha:
Thanks @sinhphamvan — the split is the right shape, and I checked the operational claims rather than skimming them. Almost all of it holds. One table does not, and it is the one that decides published grades, so I am leaving this open rather than merging.

**What I verified as correct**

| claim | checked against | result |
| --- | --- | --- |
| exit `0` clean / `1` breaking / `2` unreadable | ran all three | correct (`--fail-on never` → `0` too) |
| 21 rules, 19 fixers, R010 and R015 have none | `all_rules()` / `all_fixers()` | correct |
| SARIF `breaking`→`error`, `deprecated`→`warning`, `advisory`→`note` | `sarif.py:42` | correct |
| score→letter 95/80/60/40 | `grade.letter()` | correct |
| `mcp-migrate entry --repo owner/name` | `--help` | exists |
| `--rule` suppresses the grade | `cli.py` + run | correct |

Full suite on a merged tree (`main` @ cd55c27 + this branch): **722 passed**.

**The problem: the `Cap per rule` column in `docs/RULES_REFERENCE.md` is wrong.**

```
                 documented here    actual (mcp_migrate.grade)
breaking              -25                    -25   ✓
deprecated            -12                     -8   ✗
advisory               -6                     -3   ✗
```

The caps are not hand-written any more — `grade.py` derives them:

```python
CAP_MULTIPLE = 1
RULE_CAP = {severity: weight * CAP_MULTIPLE for severity, weight in WEIGHT.items()}
# -> {'breaking': 25, 'deprecated': 8, 'advisory': 3}
```

`25/12/6` is the **pre-#214** table. Confirmed by real board arithmetic rather than by reading — `mcp-atlassian` at its pinned `sha` fires R001 ×3 (breaking), R020 ×2 (deprecated), R010 ×1 (advisory):

```
with caps 25/ 8/3  ->  score 64      <- what the tool actually produces
with caps 25/12/6  ->  score 60
```

The board publishes **C/64**. The documented caps would put it at 60.

**Why CI did not catch it.** `main` fixed exactly this drift in ea6d5b9 ("docs: correct the per-rule cap table, and pin both drifted numbers with tests"), which landed *after* this branch's merge-base (7082dab). `README.md` on `main` now correctly reads `-25 / -8 / -3`, and `tests/test_docs.py:325` pins that table against `RULE_CAP` — but it reads `README.md` only, so a new `docs/RULES_REFERENCE.md` is unguarded. Green CI, wrong numbers.

Two things to do:

1. Correct the table to `-25 / -8 / -3`. `Cost per finding` is already right, and with `CAP_MULTIPLE = 1` the two columns are now identical by construction — worth saying so in the prose, since a reader will wonder why both columns exist.
2. Worth extending `test_docs.py:325` to check both files rather than just `README.md`, so this cannot drift again. Reading the caps out of `RULE_CAP` rather than restating them is the existing pattern.

One smaller thing while you are in there: `docs/RULES_REFERENCE.md` describes R010 as "Servers must implement `server/discover` … registering handlers without it leaves clients with no way to learn what you support." That restates R010's current `fix` text faithfully, so it is not your error — but #257 has since established the premise is wrong (the 2.x SDK registers `server/discover` in `Server.__init__`; 1.x has no such handler to register). I reproduced that independently today. Since a rule reference is exactly where someone will go to act on the finding, it may be worth holding that row until #257 settles, or landing it with a pointer to the issue.

Rebasing onto current `main` will also pick up ea6d5b9's corrected README, which is the source you drew the table from.

dheerajjha:
Reviewed this in a maintainer sweep. I checked the two pages claim by claim against the code rather than reading them, and almost everything holds — but there are two stale numbers I cannot merge past. Both are one-line fixes.

## What I ran

`pr-241` is 9 commits behind `main` (merge-base `7082dab`), so this is a real merge of `aeb8f0a` into `origin/main` (`cd55c27`) — clean, 722 passed, CI green on 3.10–3.13.

## Blocking: the cap table is wrong by two rows

`docs/RULES_REFERENCE.md`, "The grade":

| Severity | doc: cap per rule | `RULE_CAP` in `src/mcp_migrate/grade.py` |
|---|---|---|
| `breaking` | -25 | -25 ✅ |
| `deprecated` | **-12** | **-8** ❌ |
| `advisory` | **-6** | **-3** ❌ |

`CAP_MULTIPLE = 1`, so `RULE_CAP` is derived as `WEIGHT * 1` and every cap equals its weight. Read out of the code:

```
WEIGHT:   {'breaking': 25, 'deprecated': 8, 'advisory': 3}
RULE_CAP: {'breaking': 25, 'deprecated': 8, 'advisory': 3}
```

This is not new drift — it is the exact drift `ea6d5b9` ("docs: correct the per-rule cap table, and pin both drifted numbers with tests", 2026-08-11) fixed in the README, one day after you opened this PR. You copied a README that was correct at the time and has since been corrected. Nothing you did wrong; the numbers just need refreshing.

Worth knowing for the fix: `tests/test_docs.py::test_readme_cost_and_cap_table_matches_the_code` reads the caps out of `RULE_CAP` and asserts against `README.md` only. It will not see `docs/RULES_REFERENCE.md`, so if this page carries a hand-copied table it re-opens the same hole one file over. Pointing that test at both files — or having it walk a list of doc paths — would be a real improvement over what is on `main` today.

## Also stale: the pre-commit `rev`

`docs/GETTING_STARTED.md` pins `rev: v0.2.0`. `README.md` says `v0.4.0`, which is also the newest tag and the version in `pyproject.toml`. `v0.2.0` is a real tag so nothing breaks, but it points new users two releases back.

## Everything else I checked, and it was right

Verified by running the tool, not by reading it:

- **All 21 rows** of the rule table — severity and the `yes (safe)` / `yes (review)` / `no` fixer cell — match `all_rules()` and `all_fixers()` exactly. Zero mismatches.
- **"Only 19 of the 21 rules ship a fixer (R010 and R015 do not)."** Correct: `set(rules) - set(fixers) == {'R010', 'R015'}`.
- **The score-to-letter table.** Correct at every boundary: 95→A, 94→B, 80→B, 79→C, 60→C, 59→D, 40→D, 39→F.
- **"one `Mcp-Session-Id` read and fourteen of them across four files both score 75 and both grade C."** Correct — R001 is `breaking`, capped at 25, so 100−25=75.
- **Exit codes.** `0` clean, `1` on a breaking finding, `2` on a tree with no readable source — all three observed.
- **`--fail-on never`** on a breaking tree → exit `0`.
- **`--severity` changes display only** — `check --severity advisory` on `mcp-server-git` shows 3 findings and still reports D/58 and still exits `1`.
- **`--rule` suppresses the grade** — `rule_filtered: true`, `grade: null`, `score: null`.
- **"the terminal table caps each rule at 5 rows"** — `MAX_SHOWN_PER_RULE = 5` in `cli.py`.
- **SARIF.** `2.1.0`, and the mapping is exactly as documented: `breaking`→`error`, `deprecated`→`warning`, `advisory`→`note`.
- **The suppression syntax** matches `suppress.py`'s own documented form, including the required rule id and the `--` reason.
- **`fix` flags** — `--dry-run` is the default, `--write`, `--safe-only`, `--rule`, `--include-tests` all exist as described.
- **`entry --repo owner/name`** exists as described.

That is a high hit rate for 236 lines of reference material, and the two failures are both "the README moved after you copied it" rather than anything you got wrong.

## To merge

1. `deprecated` cap → `-8`, `advisory` cap → `-3`.
2. `rev: v0.2.0` → `v0.4.0`.
3. Optional but very welcome: extend the `test_docs.py` cap assertion to cover `docs/RULES_REFERENCE.md`, so this page is guarded the way the README now is.

Ping me when those are in and I will re-run the whole check and merge. Thanks for this, Sinh — moving the reference material out of the README is the right shape, and the `--json`/`counts` note at the end of the grade section ("the letter counts kinds of problem, not amount of work") is a better explanation of the cap than what the README currently gives.

dheerajjha:
@sinhphamvan — housekeeping on my previous comment: it restates the review already in this thread rather than adding to it. **There are no new asks.** The two blockers are the same two that have been standing since 11 Aug: the cap table (`deprecated` → `-8`, `advisory` → `-3`) and `rev: v0.4.0`. Please work from the 8 Sep review, which is more complete than mine — it also flags the R010 row against #257, which mine missed.

Apologies for the extra wall of text. I would have removed it if I could.

dheerajjha:
Thanks Sinh — I checked these two pages claim by claim against the code rather than against the README, and almost all of it holds. Two numbers are stale, and they are stale through no fault of yours: `main` fixed both in `ea6d5b9` ("docs: correct the per-rule cap table, and pin both drifted numbers with tests") ten commits after your branch point, so you copied the README faithfully and the README moved.

## What checks out

Verified against `all_rules()` / `all_fixers()` directly rather than the rendered table (the `rules` output line-wraps, so I read the objects):

- all 21 severities in `RULES_REFERENCE.md` — correct, all 21
- all 21 fixer confidences (`safe`/`review`/none) — correct, all 21
- "Only 19 of the 21 rules ship a fixer (R010 and R015 do not)" — correct; `len(all_fixers()) == 19`, and the set difference is exactly `{R010, R015}`
- the score-to-letter table — matches `grade.letter()` exactly at every boundary
- exit codes 0/1/2, and the hook mapping 2 -> 0 — matches `_exit_for` and `precommit.py`
- `--severity` changes display but not grade or exit code — confirmed: on mcp-server-git, `--severity advisory` still reports `D/58` and still exits 1
- `--rule` suppresses the grade — confirmed: `"grade": null, "rule_filtered": true`
- SARIF `breaking`->`error`, `deprecated`->`warning`, `advisory`->`note` — matches `sarif.py`
- the R005/R015/R016 one-finding-per-file caveat — confirmed empirically: two `ServerCapabilities(...)` constructions in one file report a single R005
- "every rule reads TypeScript" — confirmed, 21/21
- `pip install mcp-migrate` — PyPI returns 200

## The two stale numbers

**1. The per-rule cap table.** `RULES_REFERENCE.md` says:

| Severity | Cap per rule |
| --- | --- |
| `deprecated` | -12 |
| `advisory` | -6 |

Actual, from `grade.py`: `CAP_MULTIPLE = 1`, so `RULE_CAP == WEIGHT == {'breaking': 25, 'deprecated': 8, 'advisory': 3}`. 25/12/6 were the values when `CAP_MULTIPLE` was 1.5.

It changes predictions on real repos. mcp-server-git at its pinned `76d64c82` fires `{R004: 1, R007: 2, R010: 1, R016: 1, R018: 2}`:

```
with real RULE_CAP {'breaking': 25, 'deprecated': 8, 'advisory': 3}  -> penalty 42, score 58
with the table in this PR {'breaking': 25, 'deprecated': 12, 'advisory': 6} -> penalty 46, score 54
```

The tool says 58. A reader doing the arithmetic from this page gets 54.

**2. The pre-commit `rev:`.** `GETTING_STARTED.md` pins `rev: v0.2.0`; `main`'s README now pins `v0.4.0`. This is the quieter of the two — pre-commit resolves v0.2.0 happily and the user just runs two-releases-old rules forever.

## Why CI did not catch it

`tests/test_docs.py` pins the cap table, but it locates it by matching README's exact header line, so it only guards `README.md`. I merged your branch into current `main` and the full suite passes — 722 — with the wrong table sitting in `docs/RULES_REFERENCE.md`. Nothing reads the new files.

That is worth fixing in this PR rather than after it: these pages are going to drift exactly the way the README did, and the mechanism to stop it already exists. Pointing the existing cap-table test at both files (it reads `RULE_CAP` rather than restating it, so it keeps working when `CAP_MULTIPLE` changes) would cover it.

## One omission

The language-support section explains Python and TypeScript but does not mention JavaScript, which is also in `PARTIAL` and currently sits at 3 of 21 rules (R006, R017, R021, ported in #251 — again, after your branch point). Since the section's job is to tell someone whether their project gets a grade, a JS reader currently finds no answer there.

Rebase on `main`, fix the two numbers, and add JavaScript to that section, and I will merge it. The structure is right and moving this out of the README is the correct call.

dheerajjha:
Sinh — I checked these two pages claim by claim against the code on current `main` (`d55fca5`) rather than against the README they were drawn from, since `main` has moved 13 commits since you opened this. Almost all of it holds up. **One number is wrong**, and it is wrong in a way that will not be caught by CI.

## The blocker: the cap table in `RULES_REFERENCE.md`

```
| `deprecated` | -8  | -12 |
| `advisory`   | -3  | -6  |
```

Both caps are stale. `src/mcp_migrate/grade.py` derives the cap rather than hand-picking it:

```python
CAP_MULTIPLE = 1
RULE_CAP = {severity: weight * CAP_MULTIPLE for severity, weight in WEIGHT.items()}
# -> {'breaking': 25, 'deprecated': 8, 'advisory': 3}
```

So every cap equals its weight. Measured, not read off the source:

```
2 deprecated findings of one rule -> score 92   (doc's -12 cap would give 88)
2 advisory   findings of one rule -> score 97   (doc's  -6 cap would give 94)
```

This is not your mistake — the README said `-12` and `-6` when you wrote this. `ea6d5b9` ("docs: correct the per-rule cap table, and pin both drifted numbers with tests") corrected the README afterwards, and `tests/test_docs.py:324` now reads `RULE_CAP` directly so it can never drift again. Its docstring names your two numbers specifically: *"makes every cap equal its weight, and the README went on claiming -12"*.

The reason it matters more than a typo: **that test pins `README.md` only.** Nothing in `tests/` reads `docs/RULES_REFERENCE.md` or `docs/GETTING_STARTED.md` — I grepped. So merging as-is reintroduces the exact drift `ea6d5b9` fixed, in a new file, with a green suite. Worth extending `test_readme_cost_and_cap_table_matches_the_code` to cover the new page in the same PR, which would make this whole class of problem someone else's next-time problem rather than yours.

## Everything else I checked, and it was right

Verified against `all_rules()` and `mcp-migrate fixers` on `main`:

- **21 rules**, and all 21 severities in your table match the code exactly.
- **19 fixers, R010 and R015 without one** — correct.
- **Fixer confidences** — `safe` for R004, R005, R017, R021 and `review` for the rest. Correct.
- **All 21 source filenames** in your links resolve, including the long ones like `r019_tasks_polling_replaces_blocking_result.py`.
- **Letter boundaries** 95/80/60/40 — match `letter()` at every edge I probed.
- **Exit codes** — ran all three: clean `0`, a real breaking finding `1`, a tree with no supported source `2`. Your "exit 2 means we did not read your code" framing holds even under `--fail-on never`, which still exits `2`. That is a genuinely easy thing to get wrong and you got it right.
- **SARIF levels** `breaking→error`, `deprecated→warning`, `advisory→note` — match `sarif.py:42`.
- **Malformed config does not fail the run** — fed it an unknown top-level key and a nonexistent `R999`; both surfaced in `config_warnings` and the run exited `0`.
- **"every rule reads TypeScript"** — true, 21 of 21 declare `typescript`.

## One suggestion, not a blocker

The language section covers Python and TypeScript but not JavaScript. Since #251 landed, 3 rules (R006, R017, R021) declare `javascript`, and the rest do not — so a `.js` project is scanned by an eighth of the rule set. A reader of this page would reasonably assume JS is either fully covered or not read at all, and it is neither. One sentence would cover it; #149 is the tracking issue.

Fix the cap table and I will merge this. The pages are a real improvement over having it all in the README.

dheerajjha:
Thanks for this, @sinhphamvan — the structure is right and most of it is accurate. I checked the factual claims against the code rather than reading them, and found two numbers that are wrong plus one omission that will bite people. All three are small; none of them are design problems with the PR.

Merged `aeb8f0a` into `main` @ `2d9ed43` and tested that. **Full suite: 734 passed** — which is itself part of the problem, see the last section.

## What I verified as correct

Mechanically, against `all_rules()` and `all_fixers()`, not by eye:

- **All 21 rules**: every `severity` and every fixer confidence in `RULES_REFERENCE.md` matches the code. R001–R021, `breaking`/`deprecated`/`advisory`, `safe`/`review`. No exceptions.
- **"Only 19 of the 21 rules ship a fixer (R010 and R015 do not)"** — correct.
- **Grade boundaries** 95-100/80-94/60-79/40-59/0-39 match `letter()` exactly.
- **Exit codes**: `0`/`1` confirmed on the `fixer_roundtrip` fixture; `2` confirmed on a tree with no supported source.
- **`--severity` does not change the exit code** — confirmed, `--severity advisory` still exits `1` on a tree with `breaking` findings. This is a subtle claim and you got it right.
- **`--rule` suppresses the grade** — confirmed, `--format json --rule R004` returns `grade: null, score: null, rule_filtered: true`.
- **SARIF mapping** `breaking`→`error`, `deprecated`→`warning`, `advisory`→`note` — matches `sarif.py:41`.
- **The R005/R015/R016 file-scoped caveat** — matches the README and the rules' actual behaviour.
- **Malformed config warns rather than failing**, surfacing in `config_warnings` — confirmed by feeding it a bad file.

## Defect 1 — the cap table is wrong

`RULES_REFERENCE.md` says:

| Severity | Cost per finding | Cap per rule |
| --- | --- | --- |
| `breaking` | -25 | -25 |
| `deprecated` | -8 | **-12** |
| `advisory` | -3 | **-6** |

The code says otherwise:

```python
>>> from mcp_migrate.grade import WEIGHT, CAP_MULTIPLE, RULE_CAP
>>> WEIGHT,  CAP_MULTIPLE
({'breaking': 25, 'deprecated': 8, 'advisory': 3}, 1)
>>> RULE_CAP
{'breaking': 25, 'deprecated': 8, 'advisory': 3}
```

`RULE_CAP` is derived (`WEIGHT * CAP_MULTIPLE`, and `CAP_MULTIPLE` is `1`), so the caps equal the weights. `-12`/`-6` is the old hand-picked table from before #214 — the README carried exactly those numbers until `ea6d5b9` ("docs: correct the per-rule cap table, and pin both drifted numbers with tests") corrected it to `-8`/`-3`. The README says `-25`/`-8`/`-3` today, so "content is drawn from the current README" isn't quite what happened here; this table came from a pre-`ea6d5b9` copy.

It isn't academic — it changes a published number. `mcp-atlassian` fires R001 ×3 (breaking), R020 ×2 (deprecated), R010 ×1 (advisory):

```
under RULE_CAP (code):        25 + 8 + 3  = 36  ->  64   (C)
under the doc's table:        25 + 12 + 3 = 40  ->  60   (C)
```

`registry/servers/mcp-atlassian.yaml` says `score: 64`. A reader who does the arithmetic from this page gets 60 and concludes the board is wrong.

## Defect 2 — the pre-commit `rev:` is two releases stale

`GETTING_STARTED.md` has `rev: v0.2.0`. The README has `rev: v0.5.0`, and it is pinned by `tests/test_docs.py::test_readme_precommit_rev_is_the_current_version`, whose docstring is worth quoting because it is about this exact string:

> It had drifted to `v0.2.0` while the project shipped 0.3.0 and 0.4.0. A stale `rev` is quieter than a wrong one: pre-commit resolves it happily, so the user gets a working hook running two releases of rules behind, with nothing anywhere saying so.

So this is the precise value that test was written to catch, reintroduced in a file the test does not read. `v0.2.0` also predates `action.yml`, so it is stale in a second way. Should be `v0.5.0`.

## Gap 3 — the standalone config example is missing, and it silently no-ops

The guide shows the `pyproject.toml` form, then says:

> No `pyproject.toml`? Use a standalone `.mcp-migrate.toml` at the project root instead (most TypeScript projects fall here).

The keys are spelled differently in that file — top level, with a `[rules]` table, not `[tool.mcp-migrate]`. A reader following this page will copy the block above it, and get nothing:

```bash
$ printf '[tool.mcp-migrate.rules]\nR999 = "off"\n' > /tmp/badcfg/.mcp-migrate.toml
$ mcp-migrate check --format json /tmp/badcfg | jq .config_warnings
["…/.mcp-migrate.toml: unknown setting 'tool', ignoring -- `[tool.mcp-migrate]` is the
  pyproject.toml spelling; in .mcp-migrate.toml the settings are top level
  (`skip`, `include-tests`, and a `[rules]` table)"]
```

The run succeeds, the rule stays on, and the only signal is a `config_warnings` entry most people will not look at. README:270 shows the standalone form explicitly for this reason. Worth copying it across — along with README:277's rule that **a `pyproject.toml` wins if it exists at all**, whether or not it sets `[tool.mcp-migrate]`, which this page also omits.

## The structural point

The suite passed 734 on the merged tree with all three of the above in it. `tests/test_docs.py` matches on the README's literal table header and the README's `rev:` line, so `docs/` is entirely unguarded — the same "a documented number with no test holding the docs to it" that #265 is about, and the reason `ea6d5b9` added tests rather than just editing prose.

If these pages are going to be the reference material, the cap table and the `rev:` line in `docs/` want the same treatment: parameterise the two existing tests over `[README.md, docs/GETTING_STARTED.md, docs/RULES_REFERENCE.md]` rather than hard-coding `README`. Cheap, and it stops this file drifting the way the README did. The 21-rule table could be generated from `all_rules()`/`all_fixers()` in a test too, though I would not block on that.

## To land it

1. Cap column → `-25` / `-8` / `-3`.
2. `rev: v0.2.0` → `v0.5.0`.
3. Add the standalone `.mcp-migrate.toml` spelling and the pyproject-wins precedence rule.
4. Ideally, extend the two `test_docs.py` tests to cover the new files.

Everything else I checked holds up, and the `--severity` and `--rule` semantics in particular are stated more clearly here than in the README. Re-ping me when it's pushed and I'll re-verify all 21 rows plus the four items above.

dheerajjha:
Merging this. I applied the two outstanding fixes myself rather than asking a fifth time — they were one-line copies, and a good docs PR sitting open for a month over two numbers is my failure to unblock it, not yours.

**What I changed** (commit `89f4fff` on your branch):

- **Cap column: 25/12/6 → 25/8/3.** The cap is `WEIGHT × CAP_MULTIPLE` and `CAP_MULTIPLE` is `1`. Worth saying plainly: your numbers were not invented, they are what the caps actually were before #214 derived them from `WEIGHT`. You copied a README that was correct when you read it.
- **`rev: v0.2.0` → `v0.5.0`.**

**What I verified rather than re-read**, since this page will be the thing people cite:

| claim | checked against | verdict |
| --- | --- | --- |
| 19 of 21 rules ship a fixer, R010 and R015 do not | `fixers.all_fixers()` — 19 entries, no R010/R015 | correct |
| the `safe` / `review` split in the Fixer column | R004, R005, R017, R021 are `safe`; the other 15 `review` | correct |
| every A–F boundary in the score table | `letter()` at 100/95/94/80/79/60/59/40/39/0 | all ten correct |
| "one read and fourteen both score 75 and grade C" | `letter(75)` | correct |

674 tests pass on the branch.

**One thing you should know, because it changes what this page is.** Since you opened this, `docs/` became the GitHub Pages source — <https://dheerajjha.github.io/mcp-migrate/> is live, and it renders the rule list generated from `all_rules()`. So there are now two rule tables: yours, maintained by hand, and one that cannot drift because it is generated.

That is not a reason to hold this — yours carries the "what breaks" prose and the fixer column, which the generated one does not, and that prose is the valuable part. But the drift is real and it is exactly the drift that produced the cap-table problem above. I am adding a test that pins your table's rule ids and fixer column to the code, the same way `test_docs.py` already pins the README's. That is my follow-up, not yours.

Thanks for the patience on this one — and for the structure, which was right from the first push.

PR #243, opened 2026-08-10
fix(config): warn about unknown rule ids

### LINKED ISSUE #239: A well-formed but nonexistent rule id is dropped from config in silence, and the README says it isn't

adity982:
Closes #239

## Summary

- validate configured rule IDs where the loaded config and live registry are both available
- preserve the config parser's independence from the rule registry
- surface a file-qualified warning with either a close-match suggestion or `mcp-migrate rules` guidance
- exclude unknown IDs from the effective disabled-rule map and apply the same validation to `fix`
- cover unknown and registered IDs with regression tests

## Validation

- `.venv/Scripts/python.exe -m pytest tests/test_config.py -q` — 21 passed
- `.venv/Scripts/python.exe -m pytest -q` — 676 passed

### DIFF
--- a/src/mcp_migrate/cli.py
+++ b/src/mcp_migrate/cli.py
@@ -71,6 +71,23 @@ class CheckResult:
     disabled_rules: dict  # rule id -> reason, for rules a project config switched off
 
 
+def _validate_configured_rules(config: Config, known_rule_ids) -> dict[str, str]:
+    """Return configured rule disables that name a registered rule."""
+    known = sorted(known_rule_ids)
+    known_set = set(known)
+    for rule_id in sorted(set(config.disabled_rules) - known_set):
+        suggestion = difflib.get_close_matches(rule_id, known, n=1)
+        if suggestion:
+            guidance = f"did you mean {suggestion[0]}?"
+        else:
+            guidance = f"see `mcp-migrate rules` for the {len(known)} that exist"
+        source = config.source or "config"
+        warning = f"{source}: no rule {rule_id} -- {guidance}"
+        if warning not in config.warnings:
+            config.warnings.append(warning)
+    return {rid: reason for rid, reason in config.disabled_rules.items() if rid in known_set}
+
+
 def run_check_detailed(
     root: Path, *, include_tests: bool = False, rule_ids: frozenset[str] | None = None, config: Config | None = None,
 ) -> "CheckResult":
@@ -83,7 +100,7 @@ def run_check_detailed(
     effective_include_tests = include_tests or config.include_tests
     project = load_project(root, include_tests=effective_include_tests, extra_skip=config.skip)
     rules = {r.id: r for r in all_rules()}
-    disabled_rules = {rid: reason for rid, reason in config.disabled_rules.items() if rid in rules}
+    disabled_rules = _validate_configured_rules(config, rules)
     findings = []
     # One pass per (rule, language it declares), against a project view
     # filtered to that language. Views are cached because building one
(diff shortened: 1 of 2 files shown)


dheerajjha:
This is ready as far as I can tell — it is marked draft, so I have not merged it.
Mark it ready and I will. Nothing below is a change request.

You took option (1) from the issue, and the reasoning in your summary is the
reason it is the right one: the config parser stays a config parser. I checked
the behaviour rather than the diff, on all three cases that matter:

```
R999  ->  no rule R999 -- see `mcp-migrate rules` for the 21 that exist
R022  ->  no rule R022 -- did you mean R021?
R010  ->  no warning, disabled, score 100
```

The `R022` case is the one I want to call out. I suggested `get_close_matches`
in the issue without checking whether the default cutoff would actually fire on
ids this short and this similar — it would have been easy for that suggestion to
degrade into either silence or a wrong guess on every id. It lands correctly on
both sides: a near miss gets the suggestion, a wild id gets the pointer to
`mcp-migrate rules` instead of a nonsense "did you mean R019?".

Things I specifically went looking for and did not find:

- **The warning reaching every surface.** It is in `config_warnings` in `--json`,
  and on the console for both `check` and `fix`. A warning that only appears in
  one of the three would have been half a fix, since the CI case this protects
  is the one reading JSON.
- **Double-printing.** `_validate_configured_rules` mutates `config.warnings`
  and is called from two places, so I checked the count rather than assuming the
  `if warning not in config.warnings` guard was enough. One line per run from
  both `check` and `fix`.
- **`fix` honouring the same validation.** You caught that `_select_fixers` and
  the `skipped` set both had to move to the validated map. I would have missed
  the second one: a typo'd id previously landed in `skipped` and printed
  "1 fixer(s) skipped, disabled by config" for a rule that was never disabled and
  has no fixer.
- **`difflib` already imported** at `cli.py:5`, so no new dependency and no
  NameError waiting on the one code path that reaches it.

Full suite: 676 passing locally, and CI green on 3.10-3.13.

One thing this fixes that the issue did not mention: README:283 has been
promising since the config feature landed that an unrecognised rule id is
reported. It now is, so the docs went from wrong to right without touching them.

Note that main has moved under you — #240 and #242 merged this morning, both in
`cli.py`. GitHub still reports this mergeable, so it should be a clean squash,
but pull before you touch it again.

Issue #236, opened 2026-08-10
Config is read from the scan path only, so `check src/` silently ignores the repo's own config


dheerajjha:
Config is read from the scan path only, so `check src/` silently ignores the repo's own config

## Reproduce

A repo configured at its root, scanned two ways:

```
repo/
  pyproject.toml        [tool.mcp-migrate.rules] R010 = "off -- gateway handles discovery"
  src/server.py
```

```
$ mcp-migrate check .        ->  Grade A (100/100)   1 rule(s) disabled by config
$ mcp-migrate check src      ->  Grade A  (97/100)   disabled_rules: []
```

Two grades for the same code, and nothing on the second run says the config was not read.

## Cause

`load_config(root)` in `src/mcp_migrate/config.py:132` looks for `pyproject.toml` and then `.mcp-migrate.toml` in `root` and nowhere else. `root` is whatever path was passed to `check`. A repo's config lives at the repo root; a scan very often does not start there.

This is not an exotic invocation. `mcp-migrate check src/` is the natural thing to type in a repo whose Python lives under `src/`, and a CI job that scans one package of a monorepo has no other option.

## Why it matters more than the usual config-discovery papercut

Everything else in the ecosystem walks up — ruff, mypy, eslint and black all find the config from a subdirectory. So the failure is not just "config not applied", it is "config not applied in the one way every user has been trained to expect it to work". The result is a grade that differs between a developer's machine and CI depending on which directory each one happened to point at, which is precisely the class of thing this project fixed for board entries by recording `checked_with` and a sha.

## Fix

Walk up from `root` looking for the first directory that has a `pyproject.toml` containing `[tool.mcp-migrate]`, or a `.mcp-migrate.toml`. Three things to settle while doing it:

1. **Where to stop.** A `.git` directory is the obvious ceiling, with the filesystem root as the backstop. Walking past the repo root into `$HOME` would let a stray file in a parent directory quietly change someone's grade, which is the bug this issue reports, pointed the other way.

[...]

PR #246, opened 2026-08-14
fix(config): warn on unrecognised top-level keys instead of dropping them silently

### LINKED ISSUE #235: `[tool.mcp-migrate]` silently ignores every key it doesn't recognise
`[tool.mcp-migrate]` drops every key it does not recognise, without a word.

I found this by getting it wrong myself: I wrote the rule toggle at the top level instead of under `[rules]`, and the tool told me nothing at all.

## Reproduce

```toml
# pyproject.toml
[tool.mcp-migrate]
R010 = "off -- our gateway answers server/discover"
```

```
$ mcp-migrate check .
Grade A (97/100)  0 breaking, 0 deprecated, 1 advisory
```

No warning. `--json` reports `"disabled_rules": []`. The correct spelling gives what I asked for:

```toml
[tool.mcp-migrate.rules]
R010 = "off -- our gateway answers server/discover"
```

```
Grade A (100/100)   1 rule(s) disabled by config
```

Same for a typo in either of the other two keys — `include-test = true` (missing `s`) and `skips = ["vendor/"]` both parse, both do nothing, both say nothing.

## Why this is a bug and not a preference

The module already holds itself to the opposite standard one level down. Put a bad key *inside* `[rules]` and it speaks up:

```toml
[tool.mcp-migrate.rules]
R10 = "off"
```
```
config warning  not a rule id, ignoring [rules] entry 'R10'
```

So `_parse_table` warns about an unrecognised key in the rules table and about an unrecognised *value* in the rules table, but says nothing about an unrecognised key in the table it is actually parsing. `src/mcp_migrate/config.py:67-115` reads exactly three keys — `skip`, `include-tests`/`include_tests`, `rules` — and everything else falls off the end.

The cost is worse here than for a normal typo'd config, because of what this tool is for. A misspelled key means the grade you publish is not the grade you configured, and you have no way to notice: there is no output that differs. A `skip` that never took effect scans vendored code and lowers the grade; a rule toggle that never took effect leaves a rule on. Either way the number is wrong in a tool whose entire proposition is that the number is trustworthy.

## Fix

In `_parse_table`, collect the keys that were actually consumed and warn for the rest:

```python
KNOWN = {"skip", "include-tests", "include_tests", "rules"}
for key in table:
    if key not in KNOWN:
        warnings.append(f"{source}: unknown setting {key!r}, ignoring")
```

Two things worth getting right rather than just emitting the warning:

- **A rule id at the top level deserves a better message than "unknown setting".** `R010 = "off"` is not a typo, it is the right idea in the wrong place, and the warning can say so: ``did you mean [tool.mcp-migrate.rules]?``. `RULE_ID_RX` is already imported in this module.
- **A near-miss on a real key deserves the same.** `include-test` and `skips` are one character from correct; `difflib.get_close_matches(key, KNOWN)` turns the warning into a fix.

Warn, do not fail. An unknown key is not a reason to refuse to run — the existing `warnings` list already goes to the console without stopping the check, and that is the right shape.

## Tests

`tests/test_config.py` has the pattern to copy. Worth covering:

- an unknown top-level key produces a warning and does not change the outcome
- a rule id at the top level produces the `[rules]` hint specifically
- a valid config produces **no** warnings (the regression that actually bites — a warning that fires on correct config is worse than no warning at all)

---

Good first issue: one function, a clear standard already set by the code next to it, and the tests are small. Ask in the thread if you want the fixture set up for you.

Shxnque:
Fixes #235.

## What was wrong

`_parse_table` read exactly three keys (`skip`, `include-tests`/`include_tests`, `rules`) and let every other top-level key fall off the end in silence, even though the same function already warns about an unrecognised key one level down inside `[rules]`. As the issue lays out, that silence is worse here than for an ordinary typo, because a `skip` that never took effect or a rule toggle that never applied leaves the published grade different from the configured one, with no output that differs to reveal it, in a tool whose whole proposition is that the grade is trustworthy.

## The fix

One added block in `_parse_table`: collect the keys actually consumed and warn (never fail) for the rest, with a targeted hint for each of the three mistakes that actually happen:

1. **A rule id at the top level** (`R010 = "off"`) points at the rules table, spelled for the file it is in (`[tool.mcp-migrate.rules]` in `pyproject.toml`, bare `[rules]` in the standalone file).
2. **`tool` in a standalone `.mcp-migrate.toml`** (the case @dheerajjha added in the thread, copying the `pyproject.toml` shape into the standalone file) names both shapes so the reader knows which one belongs where.
3. **A near miss on a real key** (`skips`, `include-test`) gets a `difflib` "did you mean" suggestion.

## One related fix

The warning text names TOML tables like `[rules]`, and the two `console.print` sites were rendering warnings through Rich markup, which silently swallowed those brackets into empty backticks. That is a pre-existing bug (the existing `` `[rules]` must be a table `` warning shows it) that the new messages would otherwise inherit, so the warning text is now escaped at both sites. The stored strings and `--json` output are untouched.

## Evidence (running the thing)

```
# 1. rule id at the top level (pyproject.toml)
config warning  pyproject.toml: 'R010' is a rule id at the top level, ignoring -- rule toggles go in the [tool.mcp-migrate.rules] table

# 2. pyproject-shaped table in .mcp-migrate.toml
config warning  .mcp-migrate.toml: unknown setting 'tool', ignoring -- `[tool.mcp-migrate]` is the pyproject.toml spelling; in .mcp-migrate.toml the settings are top level (`skip`, `include-tests`, and a `[rules]` table)

# 3. near miss on a real key
config warning  pyproject.toml: unknown setting 'skips', ignoring -- did you mean 'skip'?

# regression: a fully valid config prints no config warning at all
1 rule(s) disabled by config (pyproject.toml): R001
```

## Tests

8 added in `tests/test_config.py`, one per case above plus the two regression guards the issue calls out specifically: a fully valid config emits **no** warnings, and the underscore alias `include_tests` stays accepted silently. Also one CLI test asserting a table name survives to rendered output (guards the escape fix). Full suite green:

```
698 passed
```

Warn, never fail, matching the existing `warnings` shape. Happy to adjust any of the hint wording.

### DIFF
--- a/src/mcp_migrate/cli.py
+++ b/src/mcp_migrate/cli.py
@@ -12,6 +12,7 @@ from dataclasses import dataclass
 from pathlib import Path
 
 from rich.console import Console
+from rich.markup import escape
 from rich.table import Table
 
 from . import __version__
@@ -417,7 +418,7 @@ def _report_config(console, result: "CheckResult") -> None:
     makes for inline `ignore` comments applies here, one config-load earlier.
     """
     for warning in result.config.warnings:
-        console.print(f"[yellow]config warning[/yellow]  {warning}")
+        console.print(f"[yellow]config warning[/yellow]  {escape(warning)}")
 
     if not result.disabled_rules:
         return
@@ -930,7 +931,7 @@ def cmd_fix(args) -> int:
     cfg = load_config(root)
     disabled_rules = _validate_configured_rules(cfg, (rule.id for rule in all_rules()))
     for warning in cfg.warnings:
-        console.print(f"[yellow]config warning[/yellow]  {warning}")
+        console.print(f"[yellow]config warning[/yellow]  {escape(warning)}")
 
     rule_ids = frozenset(args.rule) if args.rule else None
     fixers = _select_fixers(safe_only=args.safe_only, rule_ids=rule_ids, disabled=disabled_rules)
--- a/src/mcp_migrate/config.py
+++ b/src/mcp_migrate/config.py
@@ -27,6 +27,7 @@ and is never mistaken for a pass, because it never produces one.
 """
 from __future__ import annotations
 
+import difflib
 import re
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -49,6 +50,13 @@ STANDALONE_FILENAME = ".mcp-migrate.toml"
 # one learns both.
 _OFF_RX = re.compile(r"^(?:off|disabled|false)\s*(?:(?:--|:)\s*(?P<reason>.+))?$", re.IGNORECASE)
 _ON_RX = re.compile(r"^(?:on|enabled|true)$", re.IGNORECASE)
+
+# Every key `_parse_table` actually consumes. Both spellings of the tests
+# toggle are accepted (the reader takes either); the dash form is the one
+# the docstring and README use, so it is the only one a "did you mean"
+# hint ever suggests back.
+_KNOWN_KEYS = frozenset({"skip", "include-tests", "include_tests", "rules"})
+_HINT_KEYS = ("skip", "include-tests", "rules")
 
 
 @dataclass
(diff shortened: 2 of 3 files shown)


dheerajjha:
Merged. You covered all three shapes the issue described and found a fourth
thing I had not asked for and should have.

The three from the issue and its comment, verified by running them rather than
reading the diff:

```
R010 = "off"          'R010' is a rule id at the top level -- rule toggles go in
                      the [tool.mcp-migrate.rules] table
include-test = true   unknown setting 'include-test' -- did you mean 'include-tests'?
skips = ["v/"]        unknown setting 'skips' -- did you mean 'skip'?
zzzz = 1              unknown setting 'zzzz'
```

and in `.mcp-migrate.toml`, the case I flagged as the one most likely to
actually happen:

```
[tool.mcp-migrate.rules]  ->  unknown setting 'tool' -- `[tool.mcp-migrate]` is the
                              pyproject.toml spelling; in .mcp-migrate.toml the
                              settings are top level
```

Two details I want to name because they are the difference between a warning
that helps and one that becomes noise people learn to ignore:

- **`rules_label` switching on which file it came from.** Telling someone
  editing `.mcp-migrate.toml` that their toggle belongs in
  `[tool.mcp-migrate.rules]` would have sent them to a table that must not
  exist in that file. Getting this right required noticing the two files have
  different shapes, which is exactly the thing that makes the mistake in the
  first place.
- **`_HINT_KEYS` excluding `include_tests` while `_KNOWN_KEYS` accepts it.**
  Both spellings keep working, and the hint only ever names the documented
  one. A "did you mean `include_tests`?" pointing at the undocumented spelling
  would have been technically correct and quietly bad advice.

**The `escape()` is the catch of the PR and it is not cosmetic.** The warning
interpolates a key straight from someone's TOML into a rich markup string, so
a key like `"[bold red]boom[/]"` was markup, not text — at best a key
misreported back to its author, at worst rich choking on a malformed tag while
reporting a config error, which is the least helpful possible moment to raise.
Verified it now prints literally. Please open an issue if you spotted other
`console.print` sites taking untrusted text; I would rather have that list than
find them one at a time.

And the test I care most about is `test_a_fully_valid_config_produces_no_warnings`.
A warning that fires on correct config is worse than no warning at all, because
it trains people to stop reading them — that is the regression that would have
undone this whole change, and it is pinned.

698 passing. Third of the four config issues closed; #236 is the last one, and
#248 is already open against it.

Shxnque:
Thank you @dheerajjha — glad it merged cleanly, and glad the fourth shape (the `[tool.mcp-migrate.rules]`-in-`.mcp-migrate.toml` copy-paste trap) earned its place; that one bites precisely because both files are individually documented correctly, so the failure only appears when a reader trusts the more-visible example.

If mcp-migrate's rule-linting shape has been useful, you might like Quesen — same "name the exact rule that fired, fail closed on the ambiguous case" philosophy, applied to agent action-gating instead of config (MIT: github.com/Shxnque/quesen). A star or a follow on X (@SenuerenGroup) is genuinely appreciated. Either way, happy to keep sending fixes.

PR #247, opened 2026-08-14
Fix: Move rule toggle under [rules] section

### LINKED ISSUE #235: `[tool.mcp-migrate]` silently ignores every key it doesn't recognise

mercael91:
Closes #235. The rule toggle should be placed under the [rules] section in the pyproject.toml file. This ensures that the tool correctly recognizes and applies the rule.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,32 +1,4 @@
-[project]
-name = "mcp-migrate"
-version = "0.4.0"
-description = "Find and fix what the MCP 2026-07-28 spec revision breaks in your server."
-readme = "README.md"
-requires-python = ">=3.10"
-license = { text = "Apache-2.0" }
-dependencies = ["pyyaml>=6.0", "rich>=13.0"]
+[tool.mcp-migrate]
 
-[project.scripts]
-mcp-migrate = "mcp_migrate.cli:main"
-# Thin wrapper over `check` for .pre-commit-hooks.yaml -- see
-# src/mcp_migrate/precommit.py for the one thing it changes.
-mcp-migrate-precommit = "mcp_migrate.precommit:main"
-
-[project.optional-dependencies]
-dev = ["jsonschema>=4.0", "pytest>=8.0"]
-
-[build-system]
-requires = ["hatchling"]
-build-backend = "hatchling.build"
-
-[tool.hatch.build.targets.wheel]
-packages = ["src/mcp_migrate"]
-
-[tool.pytest.ini_options]
-# tests/fixtures/* are scan targets for mcp-migrate itself (deliberately
-# broken/legacy servers, etc.), not pytest test modules -- some of them
-# (by design, to reproduce real false positives) have names like
-# test_legacy_transport_compat.py and import a nonexistent `mcp` package.
-# Keep pytest from trying to collect them as its own tests.
-norecursedirs = ["fixtures", ".venv", "venv", ".git", "*.egg-info"]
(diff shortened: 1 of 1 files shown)


mercael91:
Closing — this version breaks the build (CI red across Python versions). Sorry for the noise; will only resend with a passing patch.

PR #248, opened 2026-08-14
fix(config): walk up from the scan root to find project config

### LINKED ISSUE #236: Config is read from the scan path only, so `check src/` silently ignores the repo's own config

aryansk:
## Summary

Fixes #236 — `mcp-migrate check src/` (or any subdirectory) now finds the repo's own config at the repo root instead of silently ignoring it.

## Changes

- `src/mcp_migrate/config.py`: `load_config` now walks up from the scan root looking for the first `pyproject.toml` containing `[tool.mcp-migrate]` or a standalone `.mcp-migrate.toml`. The walk stops at the first directory containing a `.git` (the repo root), with the filesystem root as a backstop, so a stray file in a parent directory outside the repo can never quietly change a grade.
  - Precedence is preserved per-directory: at each level a `pyproject.toml` that exists settles that level (the standalone file next to it is never read, so the two can't fight).
  - A section-less inner `pyproject.toml` (a monorepo packaging file that says nothing about this tool) no longer stops the walk — `check src/` still finds the repo root's real config.
- `src/mcp_migrate/cli.py`: the check output now always prints where the config came from (`config: <path>`), so a config found above the scan root is visible rather than action-at-a-distance.
- `tests/test_config.py`: six new tests covering walk-up to a root pyproject, walk-up past a section-less inner pyproject, walk-up to a parent standalone file, the `.git` ceiling, the filesystem-root backstop, and the printed config source.

## Validation

- Full suite: 696 passed (was 690 + 6 new)
- `ruff check` on changed files: no new violations (2 pre-existing in `tests/test_config.py` untouched code, 1 pre-existing in `config.py`)

### DIFF
--- a/src/mcp_migrate/cli.py
+++ b/src/mcp_migrate/cli.py
@@ -419,6 +419,9 @@ def _report_config(console, result: "CheckResult") -> None:
     for warning in result.config.warnings:
         console.print(f"[yellow]config warning[/yellow]  {warning}")
 
+    if result.config.source is not None:
+        console.print(f"[dim]config: {result.config.source}[/dim]")
+
     if not result.disabled_rules:
         return
 
--- a/src/mcp_migrate/config.py
+++ b/src/mcp_migrate/config.py
@@ -129,18 +129,20 @@ def _load_toml(path: Path) -> dict | None | list[str]:
         return [f"{path}: invalid TOML, ignoring config ({e})"]
 
 
-def load_config(root: Path) -> Config:
-    """Load project config, or an empty one carrying only its warnings.
+def _load_config_at(level: Path) -> Config | None:
+    """Load config from one directory, or None if nothing there decides.
 
-    `pyproject.toml` wins if it exists at all, whether or not it actually
-    configures anything -- a project that has one and simply doesn't use
-    `[tool.mcp-migrate]` has given a complete answer ("nothing"), and going
-    on to also read a standalone file next to it would make the two files
-    silently fight over precedence. The standalone file is for projects
-    that have no `pyproject.toml` in the first place, which given what this
-    tool targets is most JavaScript and TypeScript ones.
+    `pyproject.toml` wins at a level if it exists at all, whether or not it
+    actually configures anything -- a project that has one and simply
+    doesn't use `[tool.mcp-migrate]` has given a complete answer
+    ("nothing") for that level, and going on to also read a standalone file
+    next to it would make the two files silently fight over precedence.
+    A section-less `pyproject.toml` returns None rather than an empty
+    config: in a monorepo the inner `pyproject.toml` is a packaging file
+    that says nothing about this tool, and treating its silence as an
+    answer would make `check src/` silently ignore the repo's real config.
     """
-    pyproject_path = root / "pyproject.toml"
+    pyproject_path = level / "pyproject.toml"
     if pyproject_path.is_file():
         data = _load_toml(pyproject_path)
         if isinstance(data, list):
@@ -151,9 +153,9 @@ def load_config(root: Path) -> Config:
             section = tool_table.get("mcp-migrate") or tool_table.get("mcp_migrate")
         if isinstance(section, dict):
             return _parse_table(section, source=pyproject_path)
-        return _empty()
+        return None
 
-    standalone_path = root / STANDALONE_FILENAME
+    standalone_path = level / STANDALONE_FILENAME
     if standalone_path.is_file():
         data = _load_toml(standalone_path)
         if isinstance(data, list):
@@ -162,4 +164,24 @@ def load_config(root: Path) -> Config:
             return _parse_table(data, source=standalone_path)
         return _empty()
 
-    return _empty()
+    return None
+
+
+def load_config(root: Path) -> Config:
+    """Load project config, or an empty one carrying only its warnings.
+
+    The config is searched for by walking up from `root`, so `check src/`
+    finds the repo's config at the repo root the way ruff, mypy, eslint, and
+    black all do. The walk stops at the first directory containing a `.git`
+    (the repo root), with the filesystem root as the backstop -- a stray
+    file in a parent directory outside the repo must never quietly change
+    someone's grade.
+    """
+    current = root.resolve()
+    while True:
+        found = _load_config_at(current)
+        if found is not None:
+            return found
+        if (current / ".git").exists() or current.parent == current:
+            return _empty()
+        current = current.parent
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -95,6 +95,75 @@ def test_malformed_toml_falls_back_to_defaults_with_a_warning(tmp_path):
     assert "invalid TOML" in cfg.warnings[0]
 
 
+# --- config discovery walks up from the scan root -------------------------
+
+def test_walks_up_to_repo_root_pyproject(tmp_path):
+    """`check src/` must find the repo's config at the repo root."""
+    _write(tmp_path, "pyproject.toml", """
+[tool.mcp-migrate.rules]
+R001 = "off"
+""")
+    cfg = load_config(tmp_path / "src")
+    assert cfg.disabled_rules == {"R001": ""}
+    assert cfg.source == tmp_path / "pyproject.toml"
+
+
+def test_walks_up_past_a_sectionless_inner_pyproject(tmp_path):
+    """A monorepo inner pyproject.toml that says nothing about this tool must
+    not stop the walk -- `check src/` still finds the repo root's config."""
+    _write(tmp_path, "pyproject.toml", """
+[tool.mcp-migrate.rules]
+R001 = "off"
+""")
+    _write(tmp_path / "src", "pyproject.toml", '[project]\nname = "inner-pkg"\n')
+    cfg = load_config(tmp_path / "src")
+    assert cfg.disabled_rules == {"R001": ""}
+    assert cfg.source == tmp_path / "pyproject.toml"
+
+
+def test_walks_up_to_standalone_file_in_parent(tmp_path):
+    _write(tmp_path, ".mcp-migrate.toml", 'skip = ["vendor"]\n')
+    cfg = load_config(tmp_path / "src")
+    assert cfg.skip == frozenset({"vendor"})
+    assert cfg.source == tmp_path / ".mcp-migrate.toml"
+
+
+def test_walk_stops_at_git_ceiling(tmp_path):
+    """A stray config above the repo root must never change the grade."""
+    _write(tmp_path, ".mcp-migrate.toml", 'skip = ["stray"]\n')
+    repo = tmp_path / "repo"
+    _write(repo, ".git", "")
+    cfg = load_config(repo / "src")
+    assert cfg.skip == frozenset()
+    assert cfg.source is None
+
+
+def test_walk_stops_at_filesystem_root_backstop(tmp_path):
+    """Even without a .git ceiling, the walk must terminate."""
+    _write(tmp_path, "pyproject.toml", '[tool.mcp-migrate]\nskip = ["top"]\n')
+    # No .git anywhere; walk from a deep dir up to the filesystem root and
+    # the config at tmp_path is between them, so it must still be found.
+    cfg = load_config(tmp_path / "a" / "b" / "c")
+    assert cfg.skip == frozenset({"top"})
+
+
+def test_check_text_reports_walked_up_config_source(tmp_path, capsys):
+    """The check output must say where the config came from when it was
+    found above the scan root, per the issue."""
+    _write(tmp_path, "pyproject.toml", """
+[tool.mcp-migrate.rules]
+R001 = "off"
+""")
+    _write(tmp_path / "src", "server.py", "x = 1\n")
+    rc = main(["check", str(tmp_path / "src")])
+    assert rc == 0
+    out = capsys.readouterr().out
+    # Rich wraps/truncates long paths across lines, so assert the line exists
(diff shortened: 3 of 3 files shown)


dheerajjha:
Draft, so not merged — but it does what #236 asked, including both of the
design points I said were the whole of the difficulty, and it is close.

End to end, which is the thing that was broken:

```
check .      score 100  disabled [R010]
check src/   score 100  disabled [R010]     <- was 97 / [] before
             config: /path/to/pyproject.toml
```

Point 2 settled the way I hoped and for the right reason: a section-less inner
`pyproject.toml` does **not** stop the walk. Your comment says it better than my
issue did — in a monorepo the inner file is a packaging file that says nothing
about this tool, so treating its silence as an answer just recreates the bug one
directory down. And you kept the original precedence rule intact per level, so a
`pyproject.toml` and a `.mcp-migrate.toml` still cannot fight.

Point 3 is there too. `config: <path>` matters more than it looks: once the file
can come from a directory the user did not name, a silent no-op has been traded
for silent action-at-a-distance unless the source is visible.

## One thing to settle, and the gap is in my issue, not your code

I specified "`.git` as the ceiling, filesystem root as the backstop", and you
implemented exactly that. Testing it showed my spec was wrong for trees that are
not git checkouts:

```
1. git repo, no config inside, stray config in the parent
   -> source=None, disabled={}                       correct, ceiling holds

2. no .git anywhere (unpacked tarball), stray config two levels up
   -> source=/tmp/xxx/pyproject.toml, disabled={R001}   walks straight out
```

This project explicitly supports non-git trees — `_git_sha`'s own docstring says
"people do scan unpacked tarballs, and that's a legitimate thing to scan". For
those, the walk runs to `/`, so `mcp-migrate check ~/Downloads/some-server/`
will pick up a `pyproject.toml` sitting in `~` or `~/Downloads` and grade
against a config from a project that has nothing to do with it. That is the
exact failure my point 1 was written to prevent, and I left the hole in it.

Cheapest fix that closes it without inventing policy: stop at the first of
`.git`, **`Path.home()`**, or the filesystem root. A tarball unpacked inside a
real project still finds that project's config; one unpacked under your home
directory stops at the boundary instead of escaping into it. If you would rather
handle it differently — refusing to walk at all when the scan root is not in a
repo is also defensible, just stricter — argue for it and I will go with you;
what I do not want is to merge the walk with no ceiling for the non-git case.

Worth a test alongside the two you have, since it is the one that would bite
silently.

696 passing otherwise, and no notes on the rest.

dheerajjha:
Reviewed this by reproducing #236 in both directions on a **merged** tree (`main` @ cd55c27 + this branch), not on the branch alone. Leaving it open because it is still marked draft — from my side it is ready to merge whenever @aryansk flips it.

**The problem, on `main` @ cd55c27**

```
repo/pyproject.toml   ->  [tool.mcp-migrate.rules]  R010 = "off -- gateway handles discovery"
repo/src/server.py    ->  low-level server, @app.list_tools() + @app.call_tool()
```

```
$ mcp-migrate check repo      --format json   ->  B/94   disabled_rules: [R010]
$ mcp-migrate check repo/src  --format json   ->  B/91   disabled_rules: []
```

Two grades for the same code, and the second run says nothing about the config it did not read. Exactly as reported.

**With this branch merged**

```
$ mcp-migrate check repo/src  --format json   ->  B/94   disabled_rules: [R010]
$ mcp-migrate check repo/src                  ->  config: /tmp/repro236/pyproject.toml
```

Matches the root scan, and the source is now visible in the text output.

**The `.git` ceiling actually holds.** This is the half that could quietly change grades, so I tested it with a control rather than trusting the code read. Stray config in a parent *outside* the repo, scanning `repo/src`:

| tree | `disabled_rules` |
| --- | --- |
| `repo/.git` present | `[]` — stray config correctly ignored |
| `repo/.git` removed | `[R004, R010, R016]` — walk continues to the filesystem-root backstop |

The control confirms it is the ceiling doing the work and not some unrelated short-circuit.

**Edge cases, all as described in the PR body**

| case | result |
| --- | --- |
| deep nested scan (`a/b/c`) | finds root config |
| scan target is a single file | finds root config |
| section-less inner `pyproject.toml` between the two | does **not** stop the walk |
| inner `pyproject.toml` **with** a `[tool.mcp-migrate]` section | wins, per-level precedence preserved |

**Suite and board**

- Full suite on the merged tree: **745 passed** (with #249 also merged; 728 with this PR alone, from a 722 baseline).
- Board impact: I cloned all 16 `registry/servers/*.yaml` repos at their pinned `sha` and graded them before and after. All 16 reproduce their published grades on `main`, and **no entry moves** with this branch merged. No registry update needed.

Diff is confined to `src/mcp_migrate/{cli,config}.py` and `tests/test_config.py` — no workflow or dependency changes.

Nice work @aryansk — the per-level precedence split into `_load_config_at` is the part that makes the monorepo case come out right.

dheerajjha:
Not reviewing this in depth while it is a draft, but I re-ran it during a maintainer sweep and two results are worth having before you decide whether to undraft.

`pr-248` is 4 commits behind `main` (merge-base `dfe64f7`). Merged into `origin/main` (`cd55c27`) it is **clean**, and the full suite on the merged tree is **728 passed**. So nothing that landed since you opened it has invalidated the branch.

More usefully: because this changes where config is discovered, I checked whether it can move a published grade. I cloned all 16 `registry/servers/*.yaml` repos at their pinned `sha` and graded them with `main` and with this merged:

```
board moves: NONE (all 16 identical to main)
```

The reason is worth writing down — no board repo ships an `mcp-migrate` config at all, neither `[tool.mcp-migrate]` in a `pyproject.toml` nor a standalone `.mcp-migrate.toml`. So the walk-up has nothing to find on any of them, including the three entries that scan a subdirectory of `awslabs/mcp` and the three under `modelcontextprotocol/servers`, which are the shapes most likely to newly pick something up.

The `.git` ceiling is the part that makes that safe rather than lucky, and I am glad it is there. Thanks, Aryan.

dheerajjha:
@aryansk — this one is verified merge-ready and the only thing holding it is that it is still a draft.

What I checked, independently of the description:

- #236 reproduces both directions — without the change, `check src/` silently ignores the repo's own config; with it, the config is found.
- The `.git` ceiling is real, and I proved it with a control rather than assuming: remove `.git` and a stray parent config does leak in, which is exactly the boundary the walk-up is supposed to stop at.
- **728 passed** on the merged tree, not on the branch alone.
- **Zero board drift** — all 16 registry servers graded identically before and after. No board repo ships an `mcp-migrate` config at all, so the walk-up has nothing to find on any of them.

Mark it ready and I will merge it.

I am not flipping it myself: draft is your signal about your own work, and overriding it because I happen to have the permission would be rude. But if you are done with it and just forgot the checkbox, say so and I will flip it — or tell me you want it left alone and I will stop nudging.

Same applies to #249, which I have also verified.

dheerajjha:
Aryan — I reproduced this end to end during a maintainer sweep. It holds up in both directions, and I have no changes to ask for. Leaving it open only because it is still marked draft: **flip it to ready and I will merge it as-is.**

What I ran, against a merge of your branch into `main` at `ec6c532` (your CI ran on `dfe64f7`, 8 commits back, so the PR's own green tick was on a stale base — the merge is clean and I re-tested it):

**The bug, on `main`.** Repo with `[tool.mcp-migrate.rules] R010 = "off -- ..."` in its root `pyproject.toml`, one server under `src/`:

```
check .      ->  grade B 94  disabled_rules ['R010']
check src    ->  grade B 91  disabled_rules []
```

Two grades for the same code, second one silent. On the merged tree both are `B 94`, and `check src` prints `config: /.../repo/pyproject.toml`.

**The three decisions in #236, each tested with a real tree rather than from the diff.** Layout `outer/pyproject.toml` (stray, disables R004+R010+R016) → `outer/repo/` (`.git`, disables R010 only) → `outer/repo/pkg/pyproject.toml` (section-less packaging file) → `pkg/src/`:

| scan | `main` | merged |
|---|---|---|
| `outer/repo/pkg/src` | B 91, disabled `[]` | B 94, disabled `['R010']` |
| `outer/repo` | B 94, disabled `['R010']` | B 94, disabled `['R010']` |

The stray parent config never lands — `R004` and `R016` still fire, so the `.git` ceiling is doing real work and not just present in a test. The section-less inner `pyproject.toml` does not stop the walk, which is the call Dheeraj asked to settle in point 2 of #236, resolved the way he leaned. With no `.git` anywhere, the walk reaches a standalone `.mcp-migrate.toml` in the parent, matching the filesystem-root backstop.

**Suite on the merged tree:** 733 passed. `scripts/validate_registry.py`: 17 entries, 0 problems.

**Your tests fail without your fix.** Reverting only `src/` and running `tests/test_config.py`: 5 failed, 30 passed. The one that still passes is `test_walk_stops_at_git_ceiling`, which is correct — on `main` nothing walks, so the ceiling holds trivially. It is pinning that the fix does not overshoot, which is worth having.

**Board impact: none.** I cloned all 17 `registry/servers/*.yaml` repos at their pinned `sha` and graded them before and after. Byte-identical, so no registry entry needs updating. Worth noting this was the real risk here: three entries scan a subdirectory of `awslabs/mcp` via `path:`, and walking up from those could have picked up monorepo-root config and moved a published grade. It does not.

One observation, not a request — `main` has moved 8 commits under you. The merge is clean and I tested that merge, so nothing is needed; just flagging it so the stale green tick above does not read as more than it is.

dheerajjha:
Marked ready and merging. It has been sitting on a checkbox with no outstanding asks since 12 Sep, and that is my failure to unblock it rather than anything you owe.

I re-verified against current `main` rather than trusting my earlier review, since `main` has moved 16 commits since your merge-base:

- **Merged clean** into `main` and the merge — not the branch in isolation — passes **762 tests**.
- **The behaviour, both directions**, with `[tool.mcp-migrate.rules] R004 = "off"` at the project root and the scan pointed at `src/`:

  | | rules fired | config honoured |
  | --- | --- | --- |
  | `main` | `R004, R010, R016` | no — this is #236 |
  | this PR | `R010, R016` | yes |

One thing I got wrong while checking, worth writing down because it is a real usability signal rather than a note about me: my first attempt used `disable = ["R004"]`, which is not the schema — it is `[tool.mcp-migrate.rules]` with `R004 = "off"`. `config.py` does warn on unknown keys, but I was reading JSON output and never saw it. Not a blocker for this PR and not yours to fix; I will look at whether an unknown-key warning survives `--format json`, because a config that is silently ignored is the same class of bug this PR fixes.

Thanks for the patience, and for a fix whose test I could point at a real project and watch flip.

Issue #245, opened 2026-08-12
Fixers comment out structurally load-bearing lines, so those files now get no fix at all


dheerajjha:
Fixers comment out structurally load-bearing lines, so those files now get no fix at all

Follow-up to #244, which fixed the catastrophic half of this: `fix --write`
can no longer leave a file that does not parse. What it does instead is
**refuse** the file, which is correct as a backstop and is not a fix.

## The remaining cost

```python
"""Thin protocol shim over the MCP SDK."""
from mcp.types import (
    PingRequest,
    SetLevelRequest,
)


def handlers():
    return {"ping": PingRequest, "setLevel": SetLevelRequest}
```

```
$ mcp-migrate fix . --write
refused shim.py: applying R011, R012 would leave this file unparseable
1 file(s) had fixes available that were refused. Nothing was written.
```

Before #244 the user got broken code. Now they get nothing. Both are wrong
answers to a file where the right edit is obvious to a human.

## Why the edit breaks the file

The comment-out fixers (R001, R009, R011, R012, R013, R019) prefix a line
with `#` when it references a removed symbol. That is right for a call site
and wrong wherever the line is holding up a structure:

- **Every name in a parenthesised import.** Commenting them all out leaves
  `from mcp.types import ( )`.
- **The only statement in a function body.** `def handlers():` with its
  `return` commented out is not a TODO anyone can finish -- the information
  needed to rewrite it has been removed.

Both are cases where the fixer should either edit differently or decline
*that specific line*, rather than produce a file the guard has to catch.

## What "edit differently" looks like

Worth arguing in the thread rather than taking as settled, but as a start:

- **Import lists** — remove the name from the list rather than commenting it,
  and put the TODO above the `from` statement. That keeps the import valid at
  every intermediate state, and if the list empties, remove the statement
  entirely. This is the case with an unambiguous right answer.
- **Sole function bodies** — do not comment out the last live statement. The
  honest edit is to leave the body alone and annotate above the `def`, which
  is what R002 and R008 already do. A fixer that empties a body should treat
  that as the ambiguity it is and decline, per `fixers/base.py`.

The general rule those two share: **a fixer may comment out a line only when
something else still occupies its syntactic slot.** That is checkable without
a full parse in most cases, and it generalises past the two shapes above.

## Scope

This is six fixers and a shared decision, not a one-liner. Taking one shape
(the import list) across all six would be a good, self-contained first PR;
say in the thread if you want to split it that way and I will scope the
[...]

Issue #244, opened 2026-08-11
`fix --write` can produce Python that does not parse, and `check` then grades it A


dheerajjha:
`fix --write` can produce Python that does not parse, and `check` then grades it **A**

## Reproduce

A ten-line module, valid Python:

```python
"""Thin protocol shim over the MCP SDK."""
from mcp.types import (
    PingRequest,
    SetLevelRequest,
)


def handlers():
    return {"ping": PingRequest, "setLevel": SetLevelRequest}
```

```
$ mcp-migrate fix . --write
1 file(s), 3 change(s): 0 safe, 3 flagged for human review
No findings remain.
Changes written.
```

Result:

```python
"""Thin protocol shim over the MCP SDK."""
from mcp.types import (
    # TODO(mcp-migrate): ping is removed from the protocol; ...
    # PingRequest,
    # TODO(mcp-migrate): logging/setLevel is removed; ...
    # SetLevelRequest,
)


def handlers():
    # TODO(mcp-migrate): ping is removed from the protocol; ...
    # return {"ping": PingRequest, "setLevel": SetLevelRequest}
```

```
SyntaxError line 7: invalid syntax
```

Two independent corruptions, either of which alone breaks the file:

1. **Every name in the parenthesised import was commented out**, leaving `from mcp.types import ( )`.
2. **The entire body of `handlers()` was commented out**, leaving a function with no body.

Neither is a TODO a human can "finish". The empty parens say nothing to anybody, and a function stripped of its only statement has lost the information needed to rewrite it.

## Then the grade certifies it

```
$ mcp-migrate check .
1 Python files, 21 rules, spec 2026-07-28

Grade A. Nothing to fix. Add your badge:
![MCP 2026-07-28](https://img.shields.io/badge/MCP%202026--07--28-A-brightgreen)
```

Exit 0. **A file that does not compile is graded A and offered a badge.**

The mechanism is not "unparseable so it was skipped" — I checked that separately, and an unparseable file with live `PingRequest` usage still reports R011 and exits 1, because the rules are text-based. It is simpler and worse than that: the fixer commented the code out, `search_code` correctly skips comments, so the findings are genuinely gone by the only measure the tool has. Commenting out the evidence raises the grade.

`fix --write && check` is the obvious workflow, and it turns a C into an A while turning working code into code that will not import.

## Scope

[...]

PR #249, opened 2026-08-15
fix: strip import members instead of commenting them out (#245)

aryansk:
## Summary

Refs #245 (does not close it — the issue's own reproduction needs the sole-statement-body half too; see the maintainer comment below) — the comment-out fixers (R009, R011, R012, R013, R019) no longer
comment out members of a parenthesised `from x import (...)`. Commenting every
flagged member line left `from x import ( )`, which does not parse — so the
#244 parse guard refused the whole file and nothing got fixed.

## Changes

- `src/mcp_migrate/fixers/_textedit.py`: new shared `strip_import_members`
  helper. For a parenthesised `from x import (A, B, ...)` it removes the names
  a fixer flags instead of commenting the lines out (single-line, multi-line,
  and member-text-on-the-open/close-line shapes). The TODO is placed above the
  `from` statement once; if the list empties, the whole statement is dropped
  with the TODO, keeping the file parseable at every intermediate state.
- Wired into the five comment-out fixers whose hit regexes match SDK type
  names that can appear in import lists: `r009_initialize_handshake_removed.py`,
  `r011_ping_removed.py`, `r012_logging_set_level_removed.py`,
  `r013_subscriptions_replaced.py`, `r019_tasks_polling.py`.
  (R001 is untouched — its regex matches header access like
  `headers.get("Mcp-Session-Id")`, which can never be an import member.)
- `tests/test_fixers.py`: parametrized regression tests for all five fixers —
  member stripped (parseable), import dropped when it empties (parseable),
  idempotency — plus single-line and inline-comment shapes for R011.
- `CHANGELOG.md`: entry under Unreleased → Fixed.

## Validation

- Full suite: 707 passed (was 690; +17 new tests), 0 failed.
- The parseability guarantee is asserted directly in the tests via `ast.parse`
  on the fixer output — the exact failure mode #245 reported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,17 @@ All notable changes to this project are documented here.
   This is a backstop, not a cure -- the individual fixers still make the bad
   edit, and that is tracked separately.
 
+- **Comment-out fixers no longer comment out import members.**
+  ([#245](https://github.com/dheerajjha/mcp-migrate/issues/245))
+
+  R009, R011, R012, R013 and R019 flag SDK type names that arrive in
+  parenthesised `from x import (A, B)` lists. Commenting each flagged member
+  line out left `from x import ( )`, which does not parse -- so the #244
+  guard refused the whole file and nothing got fixed. These fixers now
+  remove the flagged name from the list (a shared `strip_import_members`
+  helper), dropping the whole statement with its TODO when the list
+  empties, so the file stays parseable at every intermediate state.
+
 ### Changed
 
 - **`fix` can now exit `1`.** It previously only ever exited `0` (or `2` for
--- a/src/mcp_migrate/fixers/_textedit.py
+++ b/src/mcp_migrate/fixers/_textedit.py
@@ -9,10 +9,16 @@ splice, never an `ast.unparse` round-trip.
 """
 from __future__ import annotations
 
+import re
 from pathlib import Path
 
 OPEN = "([{"
 CLOSE = ")]}"
+
+# A parenthesised `from x import (` opener: the only import shape where a
+# fixer commenting out a member line leaves `from x import ( )`, which does
+# not parse (see `strip_parenthesised_import_members`).
+_FROM_IMPORT_PAREN_RX = re.compile(r"^\s*from\s+\S+\s+import\s*\(")
 
 
 def find_matching_close(lines: list[str], open_idx: int, open_col: int) -> tuple[int, int] | None:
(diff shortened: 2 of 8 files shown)


dheerajjha:
It's a draft so I have not merged it, but I tested it because it is the half
of #245 that unblocks files the #244 guard currently refuses outright. As far
as I can tell it is finished for its stated scope — say the word and I will
merge it.

I ran three shapes rather than reading the diff.

**Import list, nothing else wrong** — works, including the part I thought was
most likely to be missed: when stripping the last flagged member empties the
list, the whole `from ... import (...)` statement goes with it rather than
leaving `from mcp.types import ( )`.

```python
from mcp.types import (
    PingRequest,
    SetLevelRequest,
)
HANDLERS = {}
```
```python
# TODO(mcp-migrate): ping is removed ...
# TODO(mcp-migrate): logging/setLevel is removed ...

HANDLERS = {}
```

**A stripped name that is still used elsewhere** — this is the one I went
looking for, because stripping an import while leaving a live reference trades
a `SyntaxError` for a `NameError`, and a `NameError` still parses, so the #244
guard would wave it straight through. You handled it: the member is stripped
*and* the line using it is commented, so nothing dangles. The unflagged
`ListToolsRequest` keeps both its import and its call site.

```python
from mcp.types import (
    ListToolsRequest,          # kept, still used
)
def register(server):
    server.setRequestHandler(ListToolsRequest, list_tools)
    # TODO(mcp-migrate): ping is removed ...
    # return isinstance(REGISTRY.get("x"), PingRequest)
```

**Sole function body** — still refused, correctly, because that is the other
shape in #245 and this PR does not claim it:

```
refused b.py: applying R011 would leave this file unparseable
```

That is worth saying explicitly so nobody reads it as a failure: the headline
example in #245 has *both* problems in one file, so it stays refused until the
second shape lands. If you want to take that too, the sketch in the issue still
stands — do not comment out the last live statement in a body; annotate above
the `def` the way R002 and R008 already do, and treat "this would empty a body"
as the ambiguity `fixers/base.py` says to decline. Equally fine to land this
half and let someone else take the other.

707 passing, and the sixteen #244 guard tests still pass — the guard and this
change are doing different jobs and neither has quietly disabled the other,
which is what I most wanted to confirm.

One request before merge: the CHANGELOG entry is good but says the fixers
"remove the flagged name from the list". Worth adding that a still-live usage
of that name is commented in the same pass — that is the property that makes
this safe rather than merely parseable, and it is the thing a future reader
will want to know is guaranteed.

dheerajjha:
Reviewed by reproducing #245 in both directions on a **merged** tree (`main` @ cd55c27 + this branch). The import-list half does what it claims. Leaving it open because it is still draft — and because of one scoping point below that is worth settling before it lands.

**The case this genuinely fixes.** A parenthesised import where *every* member is flagged, so the list empties:

```python
from mcp.types import (
    PingRequest,
    SetLevelRequest,
)

def handlers():
    result = {}
    result["ping"] = PingRequest
    result["setLevel"] = SetLevelRequest
    return result
```

| tree | `mcp-migrate fix --write` |
| --- | --- |
| `main` @ cd55c27 | `refused allflagged.py: applying R011, R012 would leave this file unparseable` — nothing written |
| this branch merged | 4 changes written; import dropped with both TODOs above it; output `ast.parse`s clean |

That is #245's failure mode, gone.

Worth noting for anyone reading later: the *partial* case (some members flagged, some not) already produced parseable output on `main`, because the surviving member kept the parens non-empty. The all-flagged case is where `main` actually breaks, and it is the one this fixes.

**Adversarial shapes.** I ran five import spellings past it beyond the tests in the PR; all five `ast.parse` clean:

| shape | result |
| --- | --- |
| `PingRequest as PR` (aliased) | alias stripped correctly, member removed |
| `from x import A, B` (no parens) | falls back to commenting the whole statement — pre-existing behaviour, not regressed |
| `from x import (A, B)` (parens, one line) | member stripped in place |
| member with a trailing `# comment` | comment preserved on the surviving member |
| `from .types import (...)` (relative) | handled |

**Suite and board**

- Full suite on the merged tree: **739 passed** (722 baseline + 17).
- Board: all 16 registry repos cloned at their pinned `sha` and graded before/after — **no entry moves**. Expected, since this touches only `fixers/`, but I ran it rather than assuming.
- Diff is confined to `src/mcp_migrate/fixers/` + tests + CHANGELOG. No workflow or dependency changes.

**The one thing to settle before merge: `Fixes #245` will auto-close the issue, and this is deliberately half of it.**

The issue's own headline reproduction still fails on this branch:

```python
"""Thin protocol shim over the MCP SDK."""
from mcp.types import (
    PingRequest,
    SetLevelRequest,
)


def handlers():
    return {"ping": PingRequest, "setLevel": SetLevelRequest}
```

```
$ mcp-migrate fix . --write        # on this branch, merged
refused shim.py: applying R011, R012 would leave this file unparseable
1 file(s) had fixes available that were refused. Nothing was written.
```

This is not a defect in the PR — that file also hits the *second* shape #245 describes (`return` as the sole statement in a function body), which this PR correctly scopes out and which @dheerajjha explicitly suggested splitting off ("Taking one shape (the import list) across all six would be a good, self-contained first PR"). But merging as-is closes #245 while its own repro still reproduces, which will read as fixed when it is half fixed.

Suggest either dropping to `Refs #245` and tracking the sole-body half on the issue, or opening the follow-up first so the closure hands off somewhere. @dheerajjha's call which.

Good, careful work @aryansk — keeping the file parseable at every intermediate state (rather than only at the end) is the right invariant, and the empty-list-drops-the-statement path is the bit I most expected to get wrong and did not.

dheerajjha:
Same note as on #248 — not a review while it is a draft, just a re-run during a maintainer sweep so you know where the branch stands.

`pr-249` is 4 commits behind `main` (merge-base `dfe64f7`). Merged into `origin/main` (`cd55c27`) it is **clean**, and the full suite on the merged tree is **739 passed** — the most of any open PR, and no interaction with `dfe64f7`'s parse guard, which is the commit this one is built to work with.

I did not verify the #245 behaviour itself or check board impact; fixers do not feed `check`, so grades are not in play here either way. Ping me when you undraft and I will go through it properly. Thanks, Aryan.

dheerajjha:
@aryansk — same as #248: verified, merge-ready, held only by draft status.

What I checked:

- `main` **refuses** the all-flagged import file (the fixer correctly declines to write something unparseable); the merged tree writes parseable output instead. That is the behaviour #245 asks for, reproduced both ways.
- **739 passed** on the merged tree. Zero board drift across the 16 registry servers.

One thing to change before I merge, and it is one word:

**`Fixes #245` should be `Refs #245`.** This PR handles the import-member half. The headline repro in #245 — a comment-out that removes a sole function body — still fails afterwards, deliberately and correctly out of scope here. Leaving `Fixes` would auto-close an issue whose original symptom is still live, and the next person to hit it would find a closed issue and assume it was handled.

Flip it to `Refs`, mark ready, and I will merge. As on #248, I am not flipping the draft myself — that is your call on your own work.

dheerajjha:
Aryan — I verified this during a maintainer sweep. The fix itself is correct and I reproduced its central claim both directions. Two things to settle before it can go in, one of which only appears once the branch is merged. It is also still draft.

Everything below is against a merge of your branch into `main` at `ec6c532`. Your CI ran on `dfe64f7`, 8 commits back.

## The fix works

The case where the import list empties — `main` refuses, merged tree fixes:

```
main:   refused shim.py: applying R011, R012 would leave this file unparseable
merged: 1 file(s), 4 change(s): 0 safe, 4 flagged for human review / Changes written.
```

Output drops the whole `from mcp.types import (...)` statement, puts both TODOs above it, and `ast.parse` accepts the result. Running `fix --write` a second time reports `Nothing to fix.` and the file is byte-identical, so it is idempotent. Suite on the merged tree: **744 passed**. Reverting only `src/` fails 12 of your 17 new tests.

**Board impact: none.** I cloned all 17 `registry/servers/*.yaml` repos at their pinned `sha` and graded before and after — identical. Expected for a fixer-only change, but checked rather than assumed.

## 1. `Fixes #245` closes an issue whose own reproduction still fails

Your scoping is right and matches what Dheeraj offered in the issue ("Taking one shape (the import list) across all six would be a good, self-contained first PR"). But the keyword does not know that. Running the exact file from #245's body against the merged tree:

```
refused shim.py: applying R011, R012 would leave this file unparseable, so it
was left alone.
```

Unchanged from `main`. The import list empties — your half handles that — and then `return {...}` is the sole statement in `handlers()`, gets commented out, and the body is empty. Adding a docstring to that same function so the body is not emptied makes it go through cleanly on your branch, which isolates it to the second half rather than anything wrong with yours:

```
1 file(s), 3 change(s) ... Changes written.
```

So merging as written auto-closes #245 while its headline transcript still reproduces. Suggest `Refs #245` instead, and Dheeraj can scope the sole-body half as the follow-up he mentioned.

## 2. The CHANGELOG entry lands in a released section

This one is purely a merge artifact and is invisible on the branch. You wrote the entry under `## [Unreleased]`, correctly. Since then `main` cut `release: 0.5.0` (`b536326`), which renamed that heading to `## [0.5.0] - 2026-09-12` and opened a new empty `[Unreleased]` above it. Git merges your hunk cleanly because the surrounding context still matches — it just attaches to the wrong heading now:

```
$ grep -n '^## ' CHANGELOG.md
5:## [Unreleased]
42:## [0.5.0] - 2026-09-12
140:## [0.4.0] - 2026-08-10

new entry at line 123 sits under: ## [0.5.0] - 2026-09-12
```

That would ship a claim that this fix was in 0.5.0, which it was not. Moving the block up under `[Unreleased]` (line 5) is the whole fix. Worth a rebase on current `main` so you can see it where it will actually land.

Fix those two and mark it ready and I will merge it. Nothing in `src/` needs to change.

dheerajjha:
Both outstanding items are done — I applied them rather than asking again, since neither touched `src/` and this has been sitting on them since 12 Sep.

**1. `Fixes #245` → `Refs #245`** (edited into the PR body). Re-verified against a merge with current `main` (`ad76d44`, 18 commits past your merge-base), because my earlier check was against `ec6c532`:

```
$ mcp-migrate fix . --write        # issue #245's file, verbatim
refused shim.py: applying R011, R012 would leave this file unparseable
```

Still reproduces, so `Fixes` would have auto-closed an issue whose headline transcript is still live. And the isolation still holds — give `handlers()` a docstring so its body cannot empty, and your half goes straight through:

```
1 file(s), 3 change(s): 0 safe, 3 flagged for human review
No findings remain.  Changes written.
```

`ast.parse` accepts the result, and the whole `from mcp.types import (...)` statement is dropped with both TODOs hoisted above it. Your half is complete; the remaining half is the sole-statement body, which is its own change.

**2. CHANGELOG entry moved back under `[Unreleased]`** (`a6bd46d` on your branch). Confirmed it had landed at line 123, under `## [0.5.0] - 2026-09-12` — a pure merge artifact, invisible on your branch, and it would have shipped a claim that this fix was in 0.5.0. It now sits at line 9 under `[Unreleased]`.

**779 tests pass** on the merge.

Merging once CI is green. Thanks — the `strip_import_members` helper is the right shape, and scoping to one of the six shapes was the right call.

Issue #221, opened 2026-08-07
R007 and R018 both fire on the same SDK symbol, with different severities


dheerajjha:
Fell out of merging #110 (SDK schema/params names into R018/R019). Not a regression in that PR — both findings are true — but the result is noisy and someone should decide which rule owns the symbol.

## Reproduce

```python
from mcp.types import CreateMessageResultSchema
sampler = CreateMessageResultSchema
```

```
R007 (deprecated) line 2: Sampling is deprecated. Sampling is deprecated; plan a migration.
R018 (breaking)   line 2: Server-initiated sampling/createMessage was replaced by
                          Multi Round-Trip Requests (InputRequiredResult + inputResponses).
```

One symbol, one line, two findings, two severities.

## Why it isn't simply a bug

Both statements are correct. Sampling *is* a deprecated core feature (R007), and server-initiated `sampling/createMessage` *was* replaced by multi round-trip (R018). They are different claims with different remediation. A user who reads only R007 would plan a 12-month migration and miss that the round-trip shape breaks sooner.

So "delete one" is not obviously right.

## Why it still matters

The two severities disagree about urgency on the same line, and the scoring double-counts: R007 contributes up to 12 and R018 up to 25, so a project using one deprecated SDK name pays into two independent rule caps. A file that only *imports* these names grades **F (38/100)** today.

Whether that's proportionate is the question. It may well be — a server still implementing server-initiated sampling has real work ahead. But it should be a decision someone made, not an artifact of two rules being widened independently.

## Options

1. **Leave it.** Two true findings about one symbol is honest, and the fix text differs enough to be worth reading twice.
2. **R018 defers to R007 on the shared names.** R018 keeps `roots/list` / `elicitation/create` and the wire literals; R007 keeps the Sampling family. Cleanest split, loses the round-trip advice for sampling.
3. **R007 defers to R018.** The reverse. Breaking beats deprecated, so the user sees the more urgent claim.
4. **Suppress the duplicate at report time** — one finding per (path, line), highest severity wins, both messages joined. Fixes presentation without either rule lying.

I lean 4, and I am not confident. It's the only option where neither rule has to become wrong to make the output readable, but it introduces cross-rule logic in the reporter, which is a place this codebase has deliberately kept dumb.

Related: #91 (correct code spelled indirectly) and #214 (RULE_CAP and WEIGHT chosen independently) — the scoring half of this is really the #214 question.

dheerajjha:
[...]

Issue #223, opened 2026-08-08
Board entries record a repo, not a commit, so a published grade cannot be checked against what it graded


dheerajjha:
A board entry records `grade`, `score`, `checked_with` and `spec` — the tool version and the spec revision. It does not record **which commit of the server was graded.**

```yaml
name: arxiv-mcp-server
repo: blazickjp/arxiv-mcp-server
grade: B
score: 91
checked_with: mcp-migrate 0.2.0
spec: "2026-07-28"
```

`repo:` names a repository, not a state. So "B/91" is a claim about a moving target, and there is no way for a reader — or for us — to check it against the thing it was actually a claim about.

## This is not theoretical

Re-verifying all 16 entries against current `main` today, **4 of the 16 repos had moved upstream** since the previous verification a day earlier:

```
aws-documentation-mcp-server   3493e96 -> 7ad53dc
cloudwatch-mcp-server          3493e96 -> 7ad53dc
dynamodb-mcp-server            3493e96 -> 7ad53dc
mcp-atlassian                  ec54351 -> 12fb6fa
```

All four still grade identically, so nothing is currently wrong on the board. That is luck, not a property. A server that fixes its findings, or acquires new ones, changes its true grade while the board keeps displaying the old one with no signal that anything drifted.

## Why it matters more than it looks

[#92](https://github.com/dheerajjha/mcp-migrate/issues/92) settled that these grades are *reproducible* — I've now run that verification twice and reported "all 16 reproduce exactly" both times. That claim is weaker than it sounded: what I verified is that the grades reproduce **against whatever the repos happen to contain today**, not against the state they were graded from. Those are different claims and I should have been saying the second one.

With a SHA, the verification becomes exact and cheap, and a drifted entry becomes detectable instead of invisible. It is also the other half of [#220](https://github.com/dheerajjha/mcp-migrate/issues/220), which just landed: that recorded *what the grade was computed from* on our side (suppressions); this records it on theirs.

## Scope

- `sha:` in `registry/schema.yaml` — optional at first, since the 16 existing entries don't have one and shouldn't be invalidated by this landing.
- `mcp-migrate entry` emits it when the scanned tree is a git checkout, and omits it otherwise (people do scan unpacked tarballs).
- `validate_registry.py` accepts a 7–40 char hex string.
- Backfill the 16 existing entries with the SHAs they verify against today, and say in the entry that the SHA was backfilled rather than original.
[...]

Issue #216, opened 2026-08-07
Get board entries claimed by the servers' own maintainers, or removed


dheerajjha:
Every entry on the board was written by this project scanning a public repo. **None were submitted by the maintainers of those servers.** That makes the board a survey, not adoption — see [discussion #215](https://github.com/dheerajjha/mcp-migrate/discussions/215).

Two things changed today:

- entries carry `submitted_by` (`project` | `owner`), and the rendered board states the split in its headline instead of letting 16 self-added rows read as uptake
- `registry/README.md` documents how to claim an entry, correct it, or have it removed

## Outreach

Leading with the actual findings, offering the entry second, and making removal a one-line reply. Not a campaign — each one gets its grade re-verified from a clean checkout on the current release first, because quoting someone a stale grade in an issue on their own repo is worse than not writing at all.

| server | grade | issue | outcome |
| --- | --- | --- | --- |
| excel-mcp-server | B (89) | [#153](https://github.com/haris-musa/excel-mcp-server/issues/153) | open |
| arxiv-mcp-server | B (91) | [#149](https://github.com/blazickjp/arxiv-mcp-server/issues/149) | open |
| mcp-atlassian | C (60) | [#1569](https://github.com/sooperset/mcp-atlassian/issues/1569) | open |

`mcp-atlassian` leads with a correction rather than findings: an earlier R003 produced **19 false findings** on that repo and cost it 475 penalty points. That rule is why R003 was downgraded to advisory, and their repo is named in the source comment as the reason. Re-verified today at `ec54351` — the C stands, with zero R003 findings, so it's defensible now.

## Remaining 13

Not contacted yet, deliberately. Three issues opened at once across the ecosystem is outreach; sixteen is a campaign, and it would read as one. Waiting to see how these three land — particularly whether the framing is welcome at all, which is genuinely uncertain for a "your server got a C" issue however carefully written.

Before contacting any of them: re-verify the grade on the current release. Half the remaining entries say `checked_with: 0.1.x`, and [#92](https://github.com/dheerajjha/mcp-migrate/issues/92) is open precisely because those numbers have not been re-checked since five false-positive classes were found.

## Done when

Either the entry is owner-claimed, or it is removed at the owner's request, or the owner has been asked and hasn't answered — and in that last case the board still says plainly that we compiled it.

dheerajjha:
## Update: the plan to raise fix-PRs on other projects is on hold, and the reason is a good one

[...]

PR #250, opened 2026-08-15
fix(overlap): merge known rule pairs on feature, not line (#221)

aryansk:
## What

The maintainer's comment on #221 names the open problem exactly: the merge keys on `(path, line)`, so the R007/R018 pair survives whenever the two rules describe the same migration from **different lines** — mcp-server-git still reported `R007 ×2` and `R018 ×2`. And the question posed: is same-fact-different-line resolvable without an import-aware pass?

**Yes — and it doesn't need one, because the rules already classify every match into a named feature.** R007 says "Roots"; R018 says "Server-initiated roots/list". That classification, not the line number and not the matched symbol, is the identity the merge needs. No symbol resolution is attempted; the claim being merged is "same feature", which the rules decided at match time — the pair table (`OVERLAPPING_PAIRS`) already makes that claim for same-line cases, this just stops requiring the coincidence of landing on the same line.

## Changes

- `Finding` gains an optional `feature` label; `Rule.finding()` passes it through.
- R007 (Python + TS) and R018 (Python + TS) attach their feature name to every finding.
- `overlap.py` groups known pairs on **(path, canonical feature)** (`FEATURE_CANONICAL` maps the two rules' spellings onto one key per fact), with the old **(path, line)** key kept as a fallback for findings without a feature. All findings of a fully-present pair in a group collapse into one finding: the more severe rule wins, located at its earliest line, with the other rules' messages appended so nothing is lost.

## What this fixes beyond the same-line case

- **mcp-server-git**: `R007 ×2 + R018 ×2` → **1** finding (R018, breaking, at line 11, carrying R007's message). Verified against the real checkout.
- **A latent false merge**: a busy line naming *two different features* of the pair (e.g. a Sampling import and a roots/list use on one line) was collapsed into one finding by the line key; the feature key correctly keeps them as two.
- The issue's repro (`CreateMessageResultSchema` on two lines) now reports **one** finding for the file instead of one per line.

## Validation

- 7 new/updated tests in `tests/test_overlap.py` (cross-line merge, multi-finding collapse, false-merge guard, single-rule pass-through, non-pair pass-through, end-to-end repros) + `tests/test_rules.py` updated so a rule absorbed by a known pair still counts as fired (same pattern as the existing TP_87 test).
- Regression proven: reverting `overlap.py` to the line-keyed version fails 5/5 of the new cross-line tests.
- Full suite: **696 passed**; `ruff check` clean on all touched files (the 5 remaining repo-wide ruff findings are pre-existing on `main`).


dheerajjha:
The diagnosis is right and the feature classification is the right mechanism —
keying on what the rules *decided* rather than on line proximity is exactly the
thing #221 was missing, and doing it without any symbol resolution is the
restrained version of that idea. I am not asking you to redo the approach.

But I ran it against all 16 board entries before reading further, because this
changes published grades, and one thing it does is not what #221 asked for.

## The good part, first

Surgical. Exactly one entry moves, and it is the entry the issue is about:

```
mcp-server-git   D 58 -> C 66
15 others        unchanged
```

That grade change is *correct*. mcp-server-git was being charged 25 for R018
and another 8 for R007 over the same "this project uses Roots" fact, which is
the double-charge #221 exists to stop.

## The problem: four occurrences become one

```
main    R007 line 12   Roots is deprecated...
        R007 line 464  Roots is deprecated...
        R018 line 11   Server-initiated roots/list was replaced by...
        R018 line 468  Server-initiated roots/list was replaced by...

PR250   R018 line 11   (one finding; 12, 464 and 468 are gone)
```

`_merge_group`'s docstring is explicit that this is intended — "ALL of that
pair's findings collapse into one finding" — so this is a design decision, not
a slip. I think it is the wrong one, for two reasons.

**1. It breaks a promise the README makes in as many words.** README:480, in a
paragraph that exists because someone read the letter grade and mis-sized a
migration:

> To size a migration, read `counts` and `findings` in `--json`: they carry
> **every occurrence, uncapped**.

They no longer do. `counts` now reports `breaking: 1` for a project with two
distinct Roots sites, and `findings` gives one line number out of four. The
grade is *designed* to be lossy about repetition — that is what the caps are —
and `findings`/`counts` are the documented escape hatch from that. Merging
across lines closes the escape hatch.

**2. The merged message cites itself.** Because all participants collapse
rather than one per rule, the two R018 findings merge with each other:

> Server-initiated roots/list was replaced by Multi Round-Trip Requests
> (InputRequiredResult + inputResponses). **Also flagged by R018:**
> Server-initiated roots/list was replaced by Multi Round-Trip Requests...

"Also flagged by R018" on an R018 finding, followed by a verbatim repeat of its
own text. The `Also flagged by` note was built for the cross-rule case and
reads as nonsense in the same-rule one.

## The fix, which keeps your grade result exactly

Suppress the redundant *rule* for that `(path, feature)`, instead of collapsing
the participants into one finding. Concretely: when both rules of a pair fire
for the same feature in a file, drop the lower-severity rule's findings and
leave the winner's findings alone — all of them, at their own lines.

For mcp-server-git that yields R018 at line 11 and R018 at line 468, R007 gone.
I checked the arithmetic rather than assuming:

```
main today                     58
this PR (collapse to one)      66
suppress R007, keep both R018  66   <- same grade
```

Identical grade, because the per-rule cap already makes the second R018 free.
So the entire benefit of the merge is preserved, and the cost — lost locations,
a self-citing message, a broken README promise — goes away. It should also
simplify `_merge_participants` into something closer to a filter.

`FEATURE_CANONICAL`, `_group_key`, and the feature plumbing through
`rules/base.py` all stay as they are. This is a change to what happens once a
group is identified, not to how groups are found.

## One more thing this needs before merge

A grade change on the board is not just a code change. `mcp-server-git`'s entry
records `grade: D`, `score: 58` and a `sha`, and the board's standing promise is
that a published grade is reproducible by the version named in `checked_with`.
So this PR also has to update `registry/servers/mcp-server-git.yaml` and
re-render, or main ships a board that contradicts its own tool. Happy to do
that part myself on merge if you would rather not touch the registry — say
which you prefer.

Everything else — 696 passing, the fallback to line-keying for findings with no
feature, the note that elicitation labels have no R007 counterpart — I checked
and have no notes on.

dheerajjha:
@aryansk — checking in on this one. Nothing has changed on main that affects it, and my review still reads the way I meant it.

To restate the two things standing between this and a merge, so it's not a wall of text to re-read:

1. keep the merge, but stop discarding the second finding's location and the self-citing message — the per-rule cap already makes the duplicate free, so you keep the whole benefit and drop the cost
2. `registry/servers/mcp-server-git.yaml` records `grade: D`, `score: 58`; a PR that moves a published grade has to move the registry entry with it, or main ships a board its own tool contradicts

My offer on (2) stands: I'll do the registry update and re-render on merge if you'd rather not touch it. Just say which you prefer.

The diagnosis in this PR is right, and keying on what the rules decided rather than on line proximity is the thing #221 was actually missing. I don't want it to expire over the registry chore. If you've lost interest, that's completely fine — say so and I'll pick up the remainder myself with credit to you.

dheerajjha:
Reviewed on a **merged** tree (`main` @ cd55c27 + this branch), diffed against `git merge-base` rather than `main..branch`. The mechanism works exactly as @aryansk describes. One blocker before it can land, and it is not in the code.

**The change does what it claims.**

Cross-line collapse, `CreateMessageResultSchema` referenced from three lines:

| tree | findings |
| --- | --- |
| `main` @ cd55c27 | `R018` ×3 (lines 1, 4, 7) |
| this branch merged | `R018` ×1 (line 1) |

And the real target, `mcp-server-git` at its pinned `sha` (`76d64c82`, `src/git`):

```
main      D/58   R018 line 11, R018 line 468, R007 line 12, R007 line 464, R004, R010, R016   (7 findings)
merged    C/66   R018 line 11, R004, R010, R016                                               (4 findings)
```

`R007 ×2 + R018 ×2 -> 1` , more severe rule wins, located at its earliest line. That is the PR's claim, reproduced against the real checkout. Full suite on the merged tree: **728 passed** (722 baseline + 6).

**The blocker: this moves a published board grade, and the registry entry is not updated.**

`registry/servers/mcp-server-git.yaml` still says:

```yaml
grade: D
score: 58
```

and the rendered board row in `README.md:575` still says **D**. With this branch merged the tool produces **C/66** for that entry at the same pinned `sha`. Nothing in the PR changes that file, so merging publishes a board that disagrees with the tool that generated it.

I cloned all 16 `registry/servers/*.yaml` repos at their pinned `sha` and graded them before and after. `mcp-server-git` is the **only** entry that moves — the other 15 are byte-identical, so this is a one-line registry change, not a re-grade of the board:

```yaml
grade: C
score: 66
```

**CI will not catch this.** `scripts/validate_registry.py` passes clean on the merged tree (`checked 16 entries, 0 problems`) — it validates schema and the GitHub language claim, not whether the recorded grade still reproduces. So the drift would ship green and silent, which is the failure mode `checked_with` + `sha` exist to prevent.

Not asking for a re-grade of anything else — just that one entry updated in the same PR so the board stays reproducible at the commit that changes it.

Separately, and only as an observation for @dheerajjha rather than something to fix here: the feature-keyed merge is a real answer to the question left open on #221 ("is same-fact-different-line resolvable without an import-aware pass"). Keying on a classification the rules already made at match time, rather than on line or symbol, sidesteps the symbol-resolution problem the issue thread assumed was blocking. Worth a decision on the record either way, since it changes what a published grade means for any project where the pair fires.

dheerajjha:
Sweep re-run on today's `main` (`cd55c27`), nothing new to ask for — the review above still stands unchanged, and I am only adding the one thing I found that is not already in it.

Re-ran on a real merge of `8bddd85` into `origin/main` (merge-base `dfe64f7`, branch is 4 commits behind): clean merge, **728 passed**, CI green on 3.10–3.13, and `mcp-server-git` still goes D/58 → C/66 with the other 15 board entries unchanged.

**The new bit: the existing tests cannot catch the self-citing message.** Every assertion on it in `tests/test_overlap.py` uses `in` —

```python
assert "Also flagged by R007: Roots is deprecated." in out[0].message
```

— which is satisfied by one copy or by five, so the suite passes with the message that is currently emitted. On `mcp-server-git` that is one `Also flagged by R018:` self-citation plus R007's sentence three times, in a single finding. Whichever way the group handling goes, `message.count("Also flagged by")` or a whole-string comparison would pin it; `in` will not.

@aryansk — no new asks from me beyond what is already in this thread, and the offer to take the registry half still stands.

dheerajjha:
Thanks Aryan — I reproduced the central claim and it holds. I am not merging yet, and the reason is a side effect the description does not mention: this moves a published board grade.

## What reproduces

Merged your branch into current `main` (merge-base `dfe64f7`, your branch is 5 commits behind, so I tested the actual merge rather than the branch alone). Full suite on the merged tree: **728 passed**.

On a fresh checkout of `modelcontextprotocol/servers` at the pinned `76d64c82`, `src/git`:

```
main   counts {'breaking': 2, 'deprecated': 2, 'advisory': 3}
       R018 breaking   server.py:11
       R018 breaking   server.py:468  (+ 'Also flagged by R007')
       R007 deprecated server.py:12
       R007 deprecated server.py:464

pr250  counts {'breaking': 1, 'deprecated': 0, 'advisory': 3}
       R018 breaking   server.py:11
```

Four findings to one, across lines, exactly as you describe. The feature key is the right idea and the fallback to the line key for unclassified findings is the right shape.

## Why I am holding it

**1. It moves a published grade, and the registry entry is not updated.** I cloned all 16 `registry/servers/*.yaml` repos at their pinned `sha` and graded them before and after. All 16 reproduce their published grade on `main`. With this PR:

| entry | main | this PR |
| --- | --- | --- |
| mcp-server-git | **D/58** | **C/66** |

The other 15 do not move. The mechanism is `CAP_MULTIPLE = 1`, so `RULE_CAP == WEIGHT` and a rule's penalty is capped at one firing — the score is a function of *which* rule ids fired, not how many times. Folding R007 entirely into R018 removes R007 from that set and refunds its 8 points. So this is not only a presentation change; it is a scoring change, which #221 explicitly parks as the separate #214 decision. `registry/servers/mcp-server-git.yaml` would need to move to C/66 in the same PR.

**2. `--severity deprecated` goes empty on a project that uses a deprecated feature.**

```
main   check src/git --severity deprecated  ->  2 findings (R007 at server.py:12, :464)
pr250  check src/git --severity deprecated  ->  0 findings
```

`RootsCapability` is still imported at line 12 and constructed at line 464. Someone auditing only deprecation exposure now sees nothing. The board would show mcp-server-git with zero deprecated findings while the deprecated usage is still there.

**3. The merged message repeats itself and cites its own rule.** This is the full text of the single finding above:

> Server-initiated roots/list was replaced by Multi Round-Trip Requests (InputRequiredResult + inputResponses). **Also flagged by R018:** Server-initiated roots/list was replaced by Multi Round-Trip Requests (InputRequiredResult + inputResponses). **Also flagged by R007:** Roots is deprecated. Use resource URIs instead. **Also flagged by R007:** Roots is deprecated. Use resource URIs instead. **Also flagged by R007:** Roots is deprecated. Use resource URIs instead.

`_merge_participants` appends `ordered[1:]` unconditionally, so every same-rule sibling is appended too — R018 cites itself, and R007's identical sentence lands three times. Deduplicating the appended messages by `(rule_id, message)` would fix it.

There is a fourth thing worth flagging for the maintainer rather than for you: the PR has to weaken `test_rule_fires_on_its_fixture` to accept a rule that fires but is invisible by rule id. That weakening is honest and well-commented, but it is a signal that 'rule fired' and 'rule appears in output' have come apart, and things downstream of that — the grade, severity filters, SARIF rule ids — now disagree about whether R007 fired.

None of this says the approach is wrong. #221 asks whether same-fact-different-line is resolvable without an import-aware pass, and you have shown it is. The open question your change surfaces is the one #221 deferred: if one fact is one finding, it should presumably also be charged once — and that is a grade change on a published board, which needs to be made deliberately and carried into the registry entry.

Happy to re-run the full 16-repo before/after on any update; the harness is a loop over `check --format json` against the pinned SHAs.

dheerajjha:
@aryansk — this is the decision, not another round of review. You have been waiting on a direction since 17 August and that is my fault, not yours.

## The decision

**Take the 17 August alternative: suppress the lower-severity rule, keep both findings and both locations.** Do not collapse the pair into a single synthesised finding.

The reasoning, in one line: the per-rule cap already makes the duplicate free, so collapsing buys nothing in score and costs the locations — and a reviewer who cannot see *where* the second occurrence is cannot act on it. `mcp-server-git` reaches **C / 66** either way. Same number, one of the two paths throws away information.

Concretely, for the R007/R018 pair on the same SDK symbol: let R018 report, suppress R007's finding for that symbol, and leave both R018 locations intact. `FEATURE_CANONICAL`, `_group_key` and the feature plumbing in `rules/base.py` stay exactly as you built them — this changes what happens *after* a group is identified, not how groups are found. That part of your design was right and is not in question.

## The registry, which I will handle

`registry/servers/mcp-server-git.yaml` records `grade: D`, `score: 58`. A PR that moves a published grade has to move the registry entry with it, or `main` ships a board its own tool contradicts — and #259 exists because `validate_registry.py` reports `0 problems` on exactly that drift, so CI would not catch it.

**I will do that part on merge.** You offered nothing about the registry and you should not have to learn it to land a scoring fix. Push the scoring change; I will update the entry and re-render in the same motion.

## One thing worth fixing while you are in there

The assertions in `test_overlap.py` are `in`-style, so they pass whether the message appears once or five times — which means they cannot catch the duplication this PR is about. A count assertion would make the test actually guard the behaviour.

## If you have moved on

Entirely fine, and no hard feelings after a month of silence from me. Say so, or just don't reply, and I will finish it myself with the commit attributed to you — the diagnosis and the feature-classification mechanism are yours and that is the substance of the change. I would rather you landed it, but I would rather it landed than not.

dheerajjha:
Aryan — this is a direct answer to the question Dheeraj left open in #221 ("whether same-symbol-different-line is resolvable at all without an import-aware pass"), and keying on the feature the rules already decided at match time is a genuinely good answer to it. The false-merge risk he warned about is handled: I built a line naming *two* features of the pair and the feature key keeps them apart, where the line key collapsed them.

I can't merge it yet. Three things, the first blocking.

Everything below is against a merge of your branch into `main` at `ec6c532`. **It does not merge cleanly** — `src/mcp_migrate/rules/base.py` conflicts with #266, which landed `evidence_path`/`evidence_line` on `Finding` and in `Rule.finding()`, the same two places you add `feature`. Purely additive on both sides, so I resolved it by union to test. You'll want to rebase and do it properly. Suite on my resolution: **733 passed**, and I checked #266's SARIF fix still holds (`unlocated=0`).

## 1. It moves a published grade, and the registry entry is not updated

I cloned all 17 `registry/servers/*.yaml` repos at their pinned `sha` and graded before and after. Sixteen are byte-identical. One is not:

```
mcp-server-git   D 58  (7 findings)   ->   C 66  (4 findings)
```

`registry/servers/mcp-server-git.yaml` still says `grade: D`, `score: 58`, `checked_with: mcp-migrate 0.4.0`. Merging as-is publishes a board entry the tool no longer reproduces. The entry needs updating in the same PR — that is the rule that makes the board a measurement rather than an assertion, same reasoning as the `sha` field in #223.

Your PR body predicted this exact collapse (`R007 ×2 + R018 ×2` → 1) and it is correct to the finding. It just also has a grade consequence that isn't mentioned.

## 2. The merged message repeats itself, up to three times

This is the one I'd most want fixed, because the issue this closes is titled around noise. The single finding on `mcp-server-git` now reads:

> Server-initiated roots/list was replaced by Multi Round-Trip Requests (InputRequiredResult + inputResponses). **Also flagged by R018: Server-initiated roots/list was replaced by Multi Round-Trip Requests (InputRequiredResult + inputResponses). Also flagged by R007: Roots is deprecated. Use resource URIs instead. Also flagged by R007: Roots is deprecated. Use resource URIs instead. Also flagged by R007: Roots is deprecated. Use resource URIs instead.**

R018's sentence twice, R007's sentence three times, all verbatim. It is five parts because `main` already had five underlying claims — `R018@11`, `R018@468` (itself already carrying an `R007@468` from the existing same-line merge), `R007@12`, `R007@464` — and `_merge_participants` concatenates every one of them.

`_merge_participants` appends unconditionally. Deduping the appended messages (and skipping any already present in the winner's text, which is what catches the nested `R007@468` case) would make this one clean sentence pair. Four findings became one; the text got longer.

## 3. The surviving location is the import, and the real call site is dropped

The merged finding is anchored at `server.py:11`, which is:

```python
    ListRootsResult,
```

— a name inside a `from mcp.types import (...)` block. The code that actually breaks is line 468:

```python
    roots_result: ListRootsResult = await server.request_context.session.list_roots()
```

`468` does not appear anywhere in the output — I grepped the full JSON, zero occurrences. Under `main` a reader got both lines; now they get the import and no pointer to the call.

"Earliest line" is a defensible tiebreak in the abstract, but the earliest line for these two rules is systematically the import block, because that is where SDK type names appear first. So this isn't a corner case, it is the normal shape. Two options worth considering, and I don't think this is settled — it may be worth Dheeraj's call:

- Prefer the **latest** line, or the line matched by something other than an import statement, so the anchor lands on a use site.
- Keep one finding but carry the other locations in the message (`also at line 468`) rather than dropping them. That keeps the reading win from #221 without losing the thing a user needs in order to act.

The scoring consequence in (1) is partly a symptom of this: four locations became one finding, so the grade rose. Whether that's the right calibration is really #214, as you and the issue both note — but the grade moved today, so the registry entry has to move with it either way.

Leaving this open. Nothing here says the approach is wrong; I think it's right. Happy to re-run the 17-repo board diff whenever you push.

dheerajjha:
Not merging this yet, @aryansk — but the core idea holds up, and I want to be specific about what I verified rather than leave you guessing. Three things block it, one of which is a real defect in the output.

## What I ran

`main` has moved 11 commits since your branch point, so I merged `8bddd85` into `main` @ `2d9ed43` rather than testing the PR alone. That conflicts (below), so I resolved it locally — purely additively, keeping both `Finding` fields and both `finding()` kwargs — and tested that. `src/mcp_migrate/overlap.py` is byte-identical to yours in what I tested; the resolution only touches `rules/base.py` plumbing. Everything below is from that tree.

**Full suite: 740 passed.**

## The claim reproduces

Your central argument — that the rules already classify into a named feature, and that classification is the identity the merge needs — is right, and the cross-line merge does what you say.

The issue's repro, one file, `CreateMessageResultSchema` on lines 1 and 2:

```
main   C 75  R018 server.py:1, R018 server.py:2
PR250  C 75  R018 server.py:1
```

`mcp-server-git`, the case @dheerajjha called out as the part a line-keyed merge structurally cannot reach:

```
main   D 58  R018:11, R018:468, R007:12, R007:464, R004:321, R010, R016:321
PR250  C 66  R018:11,                              R004:321, R010, R016:321
```

Four findings at two severities for one "this project uses Roots" fact → one. Exactly as described.

The false-merge guard holds too. A file importing `CreateMessageResultSchema` and `ListRootsResult` on one line keeps them as separate findings under the feature key. (`roots` doesn't collapse across lines there because R007 doesn't fire on `ListRootsResult` at all, so the pair isn't fully present — your "fully-present pair" condition doing its job.)

And a risk I went looking for and did **not** find: collapsing findings does not hide fixes. `run_fix` iterates `project.files` and calls `fixer.fix(text, path)` directly — it never consumes findings, so `overlap.dedupe` cannot reach it. No fixer covers R007 or R018 today either. That one is clear.

## Blocker 1 — the merged message is malformed

This is the one I'd fix before anything else. `_merge` joins every other finding's message with no de-duplication, and does not exclude same-rule siblings. Real output, `mcp-server-git`:

```
breaking  R018  src/mcp_server_git/server.py:11
  Server-initiated roots/list was replaced by Multi Round-Trip Requests (InputRequiredResult +
  inputResponses). Also flagged by R018: Server-initiated roots/list was replaced by Multi
  Round-Trip Requests (InputRequiredResult + inputResponses). Also flagged by R007: Roots is
  deprecated. Use resource URIs instead. Also flagged by R007: Roots is deprecated. Use resource
  URIs instead. Also flagged by R007: Roots is deprecated. Use resource URIs instead.
```

An R018 finding that says "Also flagged by R018", with R007's identical sentence three times. The group now holds every raw finding for the feature, not one per rule, so `ordered[1:]` in `overlap.py:151` is much larger than it was under the line key — the same code that read fine merging two findings produces this merging five.

Worth noting the old behaviour was correct here only by accident of group size, so this isn't a regression you introduced so much as one the wider key exposed. Deduping on `(rule_id, message)` and dropping the winner's own rule would cover it.

## Blocker 2 — it moves a published grade, and the registry entry isn't updated

`mcp-server-git` goes **D/58 → C/66**, across a letter boundary. `registry/servers/mcp-server-git.yaml` still says `grade: D`, `score: 58`, and isn't in this PR's changed files. A published grade that the tool no longer reproduces is the thing #259 is about; this PR would create one on merge.

I graded all 18 `registry/servers/*.yaml` repos at their pinned `sha`, before and after, to bound the blast radius:

- **17 of 18 unchanged** — identical grade, score and per-rule counts.
- **`mcp-server-git` only:** `D/58 {R004:1, R007:2, R010:1, R016:1, R018:2}` → `C/66 {R004:1, R010:1, R016:1, R018:1}`.

So the fix is one file. Also worth checking `serena` stays `D/47` with `R018 ×4` — R007 doesn't fire there, so the pair is never fully present and nothing collapses. The change is narrower than I expected going in, which is a point in its favour.

## Blocker 3 — conflicts with `main`

`src/mcp_migrate/rules/base.py`, against #266 (`ec6c532`), which added `evidence_path` / `evidence_line` to `Finding` after you branched. Purely additive on both sides.

One thing to catch while you're in there: `_merge` rebuilds `Finding` field by field, so it silently drops `evidence_path` and `evidence_line`. Unreachable today — the only pair is `{R007, R018}` and only R008/R010 set evidence — so this is not a live bug, and I'm not claiming it is one. But a merged project-level finding losing its SARIF anchor is precisely what #266 fixed, and the next field added to `Finding` will hit the same silent hole.

## To land it

1. Dedupe the appended messages and exclude the winner's own rule id.
2. Rebase onto `main` and carry `evidence_path` / `evidence_line` through `_merge`.
3. Update `registry/servers/mcp-server-git.yaml` to `grade: C`, `score: 66`, `checked_with: mcp-migrate 0.5.0`.

Re-ping me when it's pushed and I'll re-run all of the above, including the 18-repo board diff. Your validation section was accurate on everything I could check, and the "reverting `overlap.py` to the line-keyed version fails 5/5 of the new cross-line tests" claim is the right way to have proven it.

dheerajjha:
**Decision on the registry question: accept the regrade. Update the entry in this PR, and bump `checked_with` with it.** @aryansk — this is the last thing I was holding you on, so nothing else is pending from my side once the scoring change is in.

## What I verified before deciding

Graded all 16 registry entries against `main` and against this branch, using the same pinned checkouts both sides so only the code differs:

```
entries: main=16  pr250=16
moved:   1
  mcp-server-git: D/58 -> C/66
```

One entry, and it moves **up**, because the tool stops charging twice for one feature. That is the tool becoming more correct, not the server changing.

## Why accept rather than reject the merge semantics

The alternative on the table was to reject the dedup to avoid moving a published grade. That is the board wagging the tool — keeping a score we know is wrong so a YAML file does not have to change. The board exists to describe what the tool finds; when the tool gets more right, the board follows.

The board's actual promise is narrower than "grades never move": a published grade is reproducible **by the version named in `checked_with`**. So update `registry/servers/mcp-server-git.yaml` to `grade: C`, `score: 66` and bump its `checked_with` in the same commit. D/58 stays reproducible with the version that produced it; C/66 is reproducible with the one that supersedes it. Nothing in the promise breaks — but it does break if the code lands and the entry does not, which is the state #259 exists to make impossible and why CI would not catch it today (`validate_registry.py` reports `0 problems` on exactly this drift).

## The part of this I want on record for next time

**I am accepting this partly because it moves up, and that asymmetry is real.** A grade rising needs no one's permission. A grade *falling* because we changed our scoring — not because their code changed — is a different conversation, and on a board where maintainers claim their own entries (#216) it deserves telling them before it lands, not after. If a future scoring fix moves an entry down, that is not the same decision as this one and should not cite it as precedent.

## What I will do, so you do not have to

I offered earlier to handle the registry myself and that stands: push the scoring change with the 17 Aug approach (suppress the lower-severity rule, keep both findings and both locations) and I will do the registry edit and the re-render on merge. If you would rather do it all in one PR, also fine — it is one file, two fields, plus `checked_with`.

The one thing still worth doing while you are in there: the assertions in `test_overlap.py` are `in`-style, so they pass whether the message appears once or five times, which means they cannot catch the duplication this PR is about. A count assertion would make the test guard the behaviour.

If you have moved on, say so or say nothing and I will finish it myself with the commit attributed to you — the diagnosis and the feature-classification mechanism are yours and they are the substance of the change.

dheerajjha:
Status check, and one new thing that needs you: **this no longer merges cleanly.**

## The conflict

`src/mcp_migrate/rules/base.py`, three blocks. It is entirely my fault and entirely mechanical — #262's SARIF fix landed on `main` after your branch and touched the same three places:

| | your side | `main` |
| --- | --- | --- |
| `Finding` fields | `feature: str \| None` | `evidence_path`, `evidence_line` |
| `finding()` signature | `feature=None` param | `evidence=None`, `evidence_line=None` |
| `finding()` body | `feature=feature` | `evidence_path=...`, `evidence_line=...` |

No semantic overlap — both sides only add. **The resolution is to keep both, in every block.** One trap worth naming, because I hit it: in the signature block your half ends with `) -> Finding:`, so a straight concatenation puts the closing paren in the middle of the parameter list and the file stops parsing. The paren has to move to the end of `main`'s parameters:

```python
    def finding(self, message: str, f: SourceFile | None = None,
                line: int | None = None, snippet: str | None = None,
                feature: str | None = None,
                evidence: SourceFile | None = None,
                evidence_line: int | None = None) -> Finding:
```

I resolved it locally that way to verify the rest. **768 tests pass** on the merge.

## The registry regrade still holds

I re-ran the board against current `main` (`ad76d44`) rather than reusing yesterday's numbers, because `main` has moved a lot — including #248, which changes how config is discovered and could plausibly have shifted a grade. It did not.

**All 18 published grades reproduce exactly on `main`.** With your change merged, exactly one moves:

```
mcp-server-git    D/58  ->  C/66
```

Everything else is byte-identical. That is the decision I posted earlier, re-confirmed on a board that has grown by two entries since.

## What is left

1. Rebase on `main` and resolve as above.
2. Update `registry/servers/mcp-server-git.yaml` to `grade: C` / `score: 66`, and bump `checked_with` to `mcp-migrate 0.5.0` in the same commit.

That is all of it — nothing in `src/` needs to change. I would have pushed the resolution to your branch as I did on #249, but a three-block conflict in the file that defines `Finding` is yours to look at rather than mine to decide for you.

dheerajjha:
Reviewed and merging. This sat for a month and that is on me, not on you — the analysis in the description is the best part of it and it was right the whole time.

**I verified the central claim against the real checkout** rather than taking the description's word for it. `modelcontextprotocol/servers` at the pinned sha, `src/git`:

```
main     7 findings   R007 ×2  R018 ×2   D/58
this PR  4 findings   R007 ×0  R018 ×1   C/66
```

One fact, one finding, located at the earliest line of the more severe rule. Exactly what you said it would do.

**"Same fact, different line" really is resolvable without an import-aware pass.** The rules already classify their matches, and that classification — not the symbol, not the line — is the identity the merge wanted. Keying on it is the right answer to the question the issue asked.

I also re-graded **all eighteen** board entries on both trees. `mcp-server-git` is the only one that moves, and every other published grade reproduces exactly against `main`, so this is a narrow, well-understood change rather than a re-shuffle.

### Two things I fixed on top, in a separate commit

Neither is a criticism of the design; both are only reachable *because* of it.

**The merged message repeated itself.** Keying on the feature rather than the line is precisely what lets one rule match the same fact twice inside a group — and `ordered[1:]` then appends all of them. On `mcp-server-git` that produced:

> Server-initiated roots/list was replaced by … **Also flagged by R018:** Server-initiated roots/list was replaced by … **Also flagged by R007:** Roots is deprecated … **Also flagged by R007:** Roots is deprecated … **Also flagged by R007:** Roots is deprecated …

A rule quoted back to itself, and the same sentence three times. Each distinct message is appended once now. What is worth keeping is the advice, not the number of lines it was found on — and this text goes to a person and is handed to an agent.

**The merged `Finding` dropped `evidence_path` / `evidence_line`.** Those did not exist when you wrote this, so there was no way for you to know. They anchor a project-level finding in SARIF, and a result without a location makes GitHub code scanning reject the whole document — the bug 0.6.0 shipped to fix (#262). Carried from the winner now.

Both are pinned by tests, and I checked each fails without its fix.

### What I changed to land it

Rebased onto `main` and force-pushed to your branch, since `maintainerCanModify` was on. Your authorship is untouched; my follow-up is its own commit. The conflicts were all in `rules/base.py`, where `Finding` and `Rule.finding()` had both grown a `evidence_path`/`evidence_line` pair underneath you — additive on both sides, except the signature, which had to become one parameter list rather than two halves.

Plus the board entry re-graded with a note saying why, and the badges and site page regenerated from it.

827 tests passing. Thank you for this — and for the write-up, which is what made it reviewable a month later.

PR #253, opened 2026-08-25
fix(r010): add server/discover scaffold fixer

### LINKED ISSUE #20: Fixer: server/discover missing (R010)
No fixer exists yet for R010 ([`src/mcp_migrate/rules/r010_server_discover_missing.py`](https://github.com/dheerajjha/mcp-migrate/blob/main/src/mcp_migrate/rules/r010_server_discover_missing.py)),
which flags projects that register MCP handlers but never implement
`server/discover`.

**Spec:** SEP-2575 -- https://modelcontextprotocol.io/specification/2026-07-28/changelog

**Why this is hard:** unlike most fixers in this project, this finding
requires *adding* new, correct behavior (a `server/discover` response
describing this specific server's protocol versions, capabilities and
identity), not editing existing code. A fixer can't invent a server's name
or which capabilities it actually supports. A plausible outcome here is a
scaffold-only fixer that inserts a stub handler with obvious placeholder
values and a loud `# TODO(mcp-migrate): fill in your real capabilities` --
confidence `review`, never `safe`. It's also plausible this rule should
stay fixer-less; say so in the PR if you land there.

**What "done" looks like:** either a scaffold-inserting `Fixer` subclass
with fixtures/tests per `tests/test_fixers.py`, or a documented decision
that this stays fixer-less.

Mala732:
Fixes #20

Adds a review-only R010 fixer that generates a `server/discover` handler scaffold for one unambiguous low-level Python server receiver.

The scaffold includes TODO placeholders for protocol versions, capabilities, and server identity. Its registration line remains commented because the correct API depends on the MCP SDK version.

Ambiguous, FastMCP, functional, existing, and plain `discover` implementations are left unchanged. The server owner must adapt and register the generated handler before deployment.

## Tests

Added six focused R010 tests covering:

- Scaffold generation with TODO placeholders and `review` confidence.
- Clearing the R010 finding using the dedicated fixture.
- Leaving ambiguous receivers unchanged.
- Avoiding duplicate scaffolds for existing or plain `discover` implementations.
- Leaving FastMCP decorator and functional registrations unchanged.
- Idempotency across repeated fixer runs.

Validation:

- R010 tests: `6 passed`
- Full test suite: `704 passed`
- Runtime import verified with `mcp 1.29.1`


dheerajjha:
You got the safety question right where the competing PR got it wrong, and I
want to say that before the part that blocks this.

Keeping the registration commented *because the correct API depends on the SDK
version* is exactly the right instinct, and verifying the result actually
imports under a real SDK rather than assuming it does is the standard I'd want
every fixer PR held to. I re-ran it: your output imports cleanly under
`mcp==1.29.1`. #254 emits the same scaffold with a live `@app.discover()` and
that raises `AttributeError` at import, because no such decorator exists. You
avoided that.

## The blocker: the finding disappears without the problem being fixed

```
$ mcp-migrate check .      Grade B (91/100)   1 advisory   R010
$ mcp-migrate fix . --write
$ mcp-migrate check .      Grade B (94/100)   R010 gone
```

The server still does not answer `server/discover`. The only thing that changed
is that R010's `DISCOVER_CODE_RX` matches `\bdef\s+(?:handle_|server_|on_)*discover\b`,
and the scaffold is named `handle_discover`:

```python
# TODO(mcp-migrate): fill in your real server/discover values; adapt and register...
# @app.discover()                      <- commented, so nothing is registered
async def handle_discover(request=None) -> dict:
```

An unregistered function satisfied the check. `fix --write && check` is the
obvious workflow and it now reports a server as implementing server/discover
when nothing on the wire does. That's the same shape as #244 — the fix makes the
*evidence* vanish rather than the problem — and it's the shape my parse guard
can't catch, because this file parses fine.

Three points is small. The claim isn't: a `review` fixer may hand someone a
change they must finish, but it must not hand them a **better grade** for not
having finished it. The TODO is a note to a human; the score is a statement to
everyone else.

## The real reason neither PR can work, which is nobody's fault here

I went looking for the correct registration to suggest, and there isn't one:

```
mcp 1.29.1   Server.discover: absent     "server/discover" appears nowhere in the lowlevel server
mcp 2.1.1    Server.discover: absent     server/discover is auto-registered in Server.__init__
```

In 2.x every low-level server implements `server/discover` already, with no
handler to write. In 1.x there is nothing to register it to. So a fixer whose
entire output is a registration line has no correct form in either version —
which is a fair argument that R010 belongs in the "no fixer" column README
already has it in, and that #20 wants closing with that reasoning rather than
with a scaffold.

I've filed that as its own issue since it's bigger than either PR and neither of
you should have to rediscover it.

## If you want to keep going

The version of this I'd merge doesn't clear the finding. Either:

- emit the scaffold **and** leave R010 firing (needs the rule to require the
  registration, not just a `def`, so the name alone stops being proof), or
- keep the guidance and drop the code — R010's `fix` text already says what to
  do, and the cookbook recipe says it at length.

Both are more interesting than they sound and I'd review either. Your receiver
detection is also the careful part of this diff and worth keeping in any
version: requiring exactly one unambiguous receiver, skipping string lines
before matching decorators, and declining FastMCP and functional registration
outright are all the right calls.

Mala732:
Thanks for the detailed review and feedback, really appreciate it.
I’ll make the changes based on your suggestions.

dheerajjha:
@Mala732 — you fixed the thing I blocked on, and I want to lead with that because it was the hard part.

**The grade inflation is genuinely gone.** I ran it rather than reading it:

```
tests/fixtures/r010  ->  check:  grade B, score 91, R010 fires
                         fix --write:  2 change(s), 0 safe, 2 flagged for review
                         check:  grade B, score 91, R010 STILL fires
```

Before, the scaffold bought a server three points for implementing nothing. Now it buys nothing, which is correct — a scaffold with the registration commented out is not an implementation, and the tool should keep saying so. `test_r010_bare_function_without_registration_does_not_satisfy_rule` pins exactly that. Good test.

**Board impact is zero.** I cloned all 16 registry servers at their pinned SHAs and graded them on `ab05838` and on this branch. Identical, entry for entry.

I still can't merge it, and the reason is one line of the rule.

## `@app.discover()` is not a real API

This PR makes `@[\w.]*\.discover\s*\(` the **only** code-registration idiom R010 accepts. I checked both SDKs directly rather than trusting my memory of them:

```
mcp 1.29.1   Server.discover exists: False
             discover-ish attrs on Server: []
             "server/discover" in Server.__init__ source: False

mcp 2.1.1    Server.discover exists: False
             discover-ish attrs on Server: ['_handle_discover']   # private
             "server/discover" in Server.__init__ source: True
                 ("server/discover", types.RequestParams, self._handle_discover)
```

So there is no SDK version in which a user can write `@app.discover()`. In 1.x the method does not exist at all. In 2.x it is registered for you inside `Server.__init__` — you write nothing.

That makes three things wrong together, and they reinforce each other:

1. **The rule** now recognises one registration shape that no server can actually have. A project implementing the method the way the SDK really supports it is detected only if the literal string `server/discover` happens to appear in its source.
2. **The fixer** writes `# @{receiver}.discover()` into the user's file as the suggested registration. That is the tool handing out an API that does not exist. It is runtime-safe — you fixed the `AttributeError` that sank #254, and the emitted function does import cleanly — but a user who uncomments that line gets `AttributeError: 'Server' object has no attribute 'discover'`.
3. **The guard test** `test_r010_still_respects_a_real_server_discover_implementation_with_decorator` asserts against `@app.discover()`. So the one test standing between this rule and a false positive is guarding a fiction, and no test covers a genuine implementation.

The old regex had a real bug — `\bdef\s+(?:handle_|server_|on_)*discover\b` was too loose, and you were right to go after it. But replacing it with a decorator that doesn't exist trades a loose check for one that is precise about the wrong thing.

## What would make this mergeable

The honest question underneath is: **on a 1.x server, what should R010 tell someone to do?** There is no handler to add. The answer is "upgrade to 2.x, where the SDK registers it for you." A scaffold cannot express that, which is why I think the fixer half may not be salvageable in this shape — see #255, where a 2.x server currently grades A with zero rules fired, and which I think is the same confusion seen from the other side.

Concretely, either of these I'd review immediately:

- **Split it.** Land the rule tightening alone, with the code idiom keyed on something a real server actually has, and the guard test written against a genuine implementation. Drop the fixer from this PR. @AhmadBilalDSA did exactly this split on #256 an hour ago and it merged the same day.
- **Or keep the fixer** and change what it emits: not a decorator, but a comment naming the real mechanism for the reader's SDK version — and then say in the PR what an existing user sees before and after.

If you disagree with any of the SDK findings above, push back with a version and a snippet and I'll re-check — I'd rather be corrected than have you rewrite this on my say-so. The commands I ran are in the block above and reproduce in a clean venv.

dheerajjha:
Follow-up, and it may save you a rewrite: I filed #257 after reviewing this, because chasing the `@app.discover()` question led somewhere bigger than your PR.

`server/discover` is registered by `Server.__init__` in the 2.x SDK — it shows up in `_request_handlers` for both the low-level `Server` and the high-level `MCPServer` with no user code — and does not exist at all in 1.x. So R010 may not be a finding that wants a fixer, in which case the ask in #20 dissolves rather than gets fixed.

Please don't rewrite this PR against my review above until #257 is settled. Your rule-tightening half may still be worth landing on its own, but I'd rather not have you rebuild a fixer that a decision on #257 could delete. Nothing here was wasted — the grade-inflation bug you fixed was real, and finding that the scaffold had nothing correct to emit is what surfaced #257.

dheerajjha:
@Mala732 — the craftsmanship here is good: review-only confidence, the single-unambiguous-receiver guard, the string-literal exclusion, idempotency, and six focused tests. Full suite on a merged tree (`main` @ cd55c27 + this branch) is **730 passed**, and CI is green. So this is not a quality objection.

I am not merging it, because I reproduced the premise underneath it and the premise does not hold. This is #257, which landed after you opened this. I verified it independently rather than taking the issue's word for it — clean venvs, both SDK majors, on my own machine.

**On `mcp==2.1.1`, the SDK registers `server/discover` itself, with no user code:**

```
>>> from mcp.server.lowlevel import Server
>>> [str(k) for k in Server("demo")._request_handlers]
['ping', 'server/discover']

>>> from mcp.server.mcpserver import MCPServer
>>> [str(k) for k in MCPServer("demo")._lowlevel_server._request_handlers]
['ping', 'server/discover', 'prompts/list', 'prompts/get', 'resources/list',
 'resources/templates/list', 'resources/read', 'subscriptions/listen',
 'tools/list', 'tools/call']

>>> "server/discover" in inspect.getsource(Server.__init__)
True
```

**On `mcp==1.29.1`, there is nothing to register:**

```
1.x request_handlers:                        ["<class 'mcp.types.PingRequest'>"]
Server.discover exists:                      False
discover-ish attributes on Server:           []
"server/discover" in Server.__init__ source: False
anything named *discover* in mcp.types:      []
```

**And `.discover()` — the API this scaffold points at — exists in neither major:**

| | `Server` | `MCPServer` / `FastMCP` |
| --- | --- | --- |
| `mcp==2.1.1` | `hasattr(s, "discover")` → `False` (only a private `_handle_discover`) | no `discover` attribute |
| `mcp==1.29.1` | `False` | no `discover` attribute |

It is emitted commented-out so nothing breaks at import, but it points a maintainer at a method that has never existed. #257 notes #254 landed on the same `@app.discover()`; two contributors independently reaching for an API that isn't there is a signal about the finding, not about either of you.

**What it does to real servers.** I ran the fixer against the board repos at their pinned SHAs. It writes into `mcp-server-git`, `mcp-server-time` and `mcp-server-fetch`:

```
# TODO(mcp-migrate): fill in your real server/discover values; adapt and register this handler with your SDK.
# @server.discover()
async def handle_discover(request=None) -> dict:
```

**And the audience it can reach is exactly the audience it cannot help.** Same trivial server in both spellings:

| server | `mcp-migrate check` | fixer acts? | reality |
| --- | --- | --- | --- |
| 1.x spelling (`@app.list_tools()`) | `B/91`, R010 fires | yes, writes the scaffold | 1.x has no `server/discover` to register — the real remediation is "upgrade the SDK" |
| 2.x spelling (`MCPServer`, `@app.tool()`) | `A/100`, **no findings at all** | never reached | already answers `server/discover`; 10 handlers registered at construction |

So the fixer fires only where the advice is impossible, and never where the handler is actually the question. (The 2.x row is #255 — the scanner does not recognise a 2.x server at all. Reproduced separately.)

For context on blast radius: R010 fires on **16 of 16** board entries at their pinned SHAs, and is the *only* finding on seven of them, including four A grades.

**Where that leaves this PR.** I have deliberately left it open rather than closing it — the call on whether #20 gets fixed or closed is @dheerajjha's, and #257 proposes gating R010 on the resolved SDK version instead (2.x → don't fire; 1.x → fire, but with "upgrade the SDK" as the remediation). If that lands, most of this fixer's reachable surface disappears, and the part that remains wants to say something different.

If you want to keep the work pointed at something that will land, @Mala732, the SDK-version gate in #257 is the piece with no owner and a clear shape, and you have already demonstrated you can navigate this codebase carefully. Sorry to meet good work with a "not this" — the premise problem genuinely was not visible when you started.

Mala732:
Once again, really appreciate the detailed review, @dheerajjha ! Your reasoning makes total sense, and I agree that dropping the fixer is the right move. I’ll close this PR and take a look at #257.

dheerajjha:
@Mala732 — closing this, and I want to be precise about why, because it is not a judgement on your work.

I closed #20, the issue this implements, as not fixable. The evidence is there and in #257: `@app.discover()` exists in no version of the SDK, `server/discover` does not exist at all in 1.x, and 2.x registers it inside `Server.__init__` before any user code runs. A scaffold for it is wrong on one version and unnecessary on the other.

**What you got right, and what I verified rather than assumed:**

- The grade inflation I blocked on is genuinely gone. I ran it: `tests/fixtures/r010` scores **B/91 before the fixer and B/91 after**, with R010 still firing. Your earlier version bought a server three points for implementing nothing; this one buys nothing, which is correct.
- `test_r010_bare_function_without_registration_does_not_satisfy_rule` pins exactly that, and it is a good test.
- Your scaffold is runtime-safe. #254's version crashed at import; yours does not.
- Zero board impact — I graded all 16 registry servers at their pinned SHAs on your branch and on main. Identical, entry for entry.

You were also the one who got the safety question right when a competing PR got it wrong, and refused to hand out an unearned grade. That instinct is the thing I most want in this repo.

**Why I am not landing the rule half separately.** I considered it. But the tightening keys `DISCOVER_CODE_RX` on `@[\w.]*\.discover\s*\(` as the sole code-registration idiom, and that idiom does not exist — so merging it would leave the rule precise about the wrong thing, and the one guard test standing between it and a false positive asserts against an API no server can have. The right version of that change depends on what #257 decides R010 should key on, which is not settled.

I am sorry this ended here after you did the work twice. That is on the issue, which I wrote, not on you. If you want the next thing: **#257** is the live question, and **#260** (R001 never matches `Mcp-Session-Id` in Python — two files differing by one variable name grade C/75 and A/100) is a sharp, self-contained bug that I think you would be good at.

Reopen this if you disagree with the SDK findings — a version and a snippet will change my mind faster than my say-so should change yours.

PR #254, opened 2026-08-29
feat(rules,fixers): harden R010 wire boundary and add ServerDiscoverFixer with test suite

AhmadBilalDSA:
## Description
Addresses the wire method boundary gap in R010 and implements the missing `ServerDiscoverFixer` stub handler generation.

### 1. Rule Hardening (`R010`)
- `src/mcp_migrate/rules/r010_server_discover_missing.py`: Updated `_has_discover` and `_ts_has_discover` to use `wire_method("server/discover")` instead of bare `r"server/discover"`. Prevents longer identifiers (e.g., `server/discoverFoo`) from falsely suppressing R010.

### 2. New Fixer (`ServerDiscoverFixer`)
- Added `src/mcp_migrate/fixers/r010_server_discover.py` (`review` confidence):
  - Targets files registering real MCP handlers (low-level or FastMCP) lacking `server/discover`.
  - Appends stub handler on the existing instance (`@app.discover()` / `.setRequestHandler("server/discover", ...)`) with explicit `TODO(mcp-migrate)` markers.
  - Handles per-language comment syntax, ensures idempotency, and safely skips `self.` receivers and non-SDK `.tool(` builders.

### 3. Tests & Regressions
- Added 16 fixer unit tests in `tests/test_fixers.py` covering golden cases, receiver recovery, FastMCP/functional routes, TS comment syntax, wire-routing, and idempotency.
- Added boundary regression coverage in `tests/test_regressions.py` and `tests/test_typescript.py`.

### 4. Documentation & Registry
- Updated `README.md` (R010 fixer status to `yes (review)`).
- Updated `cookbook/` (recipe 02 index count and fixer documentation) and `CHANGELOG.md`.

## Verification
- `scripts/validate_registry.py`: **0 problems detected**.
- `pytest` targeted suites: **390 passed**.
- Full test suite: **712 passed** (the 2 failures in `tests/test_entry_sha.py` reproduce identically on clean `HEAD` due to local `tmp_path` nesting inside a parent git root).


AhmadBilalDSA:
Updated test fixtures with `GIT_CEILING_DIRECTORIES` isolation in `tests/conftest.py` to prevent parent repository traversal. Full test suite is now 100% green (714/714 passed).

dheerajjha:
Two changes in one PR, and they need opposite answers. The rule hardening is
correct and I want it. The Python fixer writes a decorator that does not exist
in any released SDK, and turns a working server into one that will not import.

CI here is `action_required` (fork PR), so I ran everything locally against both
SDK majors rather than waiting on it.

## Blocking: `@app.discover()` is not a real API

Your Python scaffold emits the registration line *live*, not commented:

```python
@app.discover()
async def handle_server_discover(request=None) -> dict:
```

There is no `discover` attribute on `mcp.server.Server` in either SDK line:

```
mcp 1.29.1   Server.discover exists: False
mcp 2.1.1    Server.discover exists: False
```

So `fix --write` on a working 1.x server produces a file that dies on import:

```
$ python -c "import server"
AttributeError: 'Server' object has no attribute 'discover'
```

Before the fix that server imported fine and scored **B (91/100)**. After it, it
scores **B (94/100)** and no longer starts. The grade went *up* because R010's
`DISCOVER_CODE_RX` matches `@[\w.]*\.discover\s*\(` — we wrote the evidence the
rule looks for, and the rule believed us.

This is the failure mode #244 exists to prevent, in the one shape #244's guard
cannot see: the guard refuses an edit that makes a parseable file unparseable,
and `@app.discover()` parses perfectly. It is a clean `ast.parse`, an
`AttributeError` at runtime, and a better letter grade. A wrong `--write` on
correct code is the worst thing this project can ship, so this half cannot land
as written.

Your own comment text anticipated the problem — *"or delete the stub if the SDK
already auto-implements server/discover"* — but the code doesn't act on it.

## Your TypeScript half is right, and it's the contrast that proves the point

```ts
server.setRequestHandler("server/discover", async () => ({
  protocolVersions: ["2026-07-28"], ...
}));
```

That is a real TS SDK call, the bare-wire-string form R016 already accommodates,
and it genuinely registers the handler. Same intent, correct API. The Python
side has no equivalent because 1.x has no `server/discover` support at all
(`"server/discover"` does not appear anywhere in `mcp==1.29.1`'s lowlevel
server), and 2.x doesn't need one — see below.

## What I want: split the rule hardening out and I'll merge it today

`wire_method("server/discover")` fixes a real silent suppression, and I
reproduced it both directions:

```python
METRIC = "server/discoverLatency"   # a metric name, not an implementation

main    R010 findings: 0     <- rule suppressed project-wide by a metric name
pr254   R010 findings: 1
```

That is the exact failure the rule's own comment calls the worst one available
("not a false finding a maintainer can argue with, but a missing one nobody ever
sees"), and the unbounded pattern was sitting right under the comment warning
about it.

Board impact: none. Zero occurrences of `server/discover` in any `.py`/`.ts`
across all 11 cached board repos, so no published grade moves.

Send that as its own PR — the `wire_method` change plus its two regression tests
— and it goes in immediately.

## On the fixer itself, for both this and #253

@Mala732 has #253 open against the same issue. Neither approach works, and the
reason is the same one, so I've written it up as its own issue rather than
making either of you guess: **the SDK moved and no correct Python registration
exists to write.** In 2.x, `server/discover` is auto-registered in
`Server.__init__` and needs no handler at all. In 1.x, there is nothing to
register it *to*. A fixer whose whole job is emitting a registration line has no
correct output in either version, which is a decent argument that R010 should
stay in the "no fixer" column README already puts it in.

Thanks for finding the boundary bug — that part is a genuine catch and I'd have
missed it.

AhmadBilalDSA:
Thanks for the clear feedback and SDK breakdown, @dheerajjha! 

Per your review, I have split out the `wire_method("server/discover")` rule hardening along with the regression tests into a dedicated, clean standalone PR: #256. 

We can merge #256 for the rule fix, and keep this PR on hold if we want to revisit a TypeScript-only fixer later.

AhmadBilalDSA:
Closing this in favor of #256 per the discussion above. Since Python MCP 2.x handles `server/discover` out-of-the-box and 1.x has no target API, keeping R010 in the no-fixer column is the right call. All boundary hardening and regression tests are isolated cleanly in #256. Thanks for the guidance!

PR #256, opened 2026-09-01
fix(rules): harden R010 wire boundary against prefix collision

AhmadBilalDSA:
Hi @dheerajjha,

As requested in #20 and #254, this PR extracts strictly the **R010 rule boundary hardening**:

- Replaces bare `r"server/discover"` with `wire_method("server/discover")` in both `_has_discover` and `_ts_has_discover` inside `src/mcp_migrate/rules/r010_server_discover_missing.py`.
- Prevents longer identifier prefixes (e.g., `server/discoverLatency`, `server/discoverFoo`) from falsely suppressing R010.
- Adds isolated regression tests in `tests/test_regressions.py` and `tests/test_typescript.py`.
- Excludes all fixer code and documentation changes.

All test suites pass cleanly. Ready for merge!


AhmadBilalDSA:
Hi @dheerajjha, gentle ping on this one! Since #254 has been closed, this standalone PR is ready to merge whenever you have a moment to approve the CI workflow. Thanks!

dheerajjha:
Merged as cd55c27. Thank you for splitting this out — the narrow PR was the right call and it made the review straightforward.

I reproduced the bug rather than reading the diff, in both directions. Three fixtures, all real MCP servers with `@app.list_tools()` / `@app.call_tool()` handlers so R010's precondition is genuinely met, graded with `--format json` so the count comes from the finding list rather than from scraping console prose:

| fixture | main `ab05838` | this PR | expected |
|---|---|---|---|
| `{"server/discoverLatency": ...}`, no real handler | **0 findings — silently suppressed** | 1 | must fire |
| genuine `"server/discover"` string | 0 | 0 | must stay silent |
| no discover string at all | 1 | 1 | must fire |

The middle row is the one I cared about most. Tightening a suppression check is exactly where an over-correction hides, and `test_r010_still_suppressed_by_a_real_wire_string_literal` on main plus that fixture say it didn't happen here.

Your two tests earn their place: reverting only `r010_server_discover_missing.py` to main's version and keeping the tests, both fail —

```
FAILED tests/test_regressions.py::test_r010_is_not_suppressed_by_a_wire_name_merely_containing_discover
FAILED tests/test_typescript.py::test_r010_is_not_satisfied_by_a_wire_method_that_merely_contains_discover
```

Full suite on the merged tree: 722 passed.

**Board impact: zero, checked rather than assumed.** I cloned all 16 registry servers at their pinned SHAs and graded them on `ab05838` and on the merged tree. Identical, entry for entry. Not one of them contains the string `server/discover` anywhere, so the branch you changed is never reached for any of them — but I ran the grades rather than resting on the grep.

One thing worth recording for whoever touches R010 next: `wire_method()` was already used by R007, R009, R012, R018, R019 and by `MCP_SURFACE_RX` itself. R010 was the last place spelling a wire name as a bare pattern, so this closed a real inconsistency, not just one bug.

PR #267, opened 2026-09-12
﻿fix(r001): match Mcp-Session-Id header literal via search_wire

### LINKED ISSUE #260: R001 never matches the Mcp-Session-Id header in Python, only variables named mcp_session_id, so renaming a local hides a breaking finding
## Summary

R001's Python path matches with `search_code`, which skips every `STRING` token. `Mcp-Session-Id` contains hyphens, so in Python it can *only* ever appear inside a string literal — which means that alternative in `PY_RX` can never match real code. What actually fires R001 on Python is the identifier spellings `mcp_session_id` / `MCP_SESSION_ID`.

The result is that R001 grades the **variable name**, not the header access. Rename the local and a `breaking` finding disappears.

## Reproduce

Two files, identical except for one local variable's name:

```bash
mkdir -p /tmp/r001/named /tmp/r001/renamed
cat > /tmp/r001/named/server.py <<'EOF'
def handle(request):
    mcp_session_id = request.headers.get("Mcp-Session-Id")
    return lookup(mcp_session_id)
EOF
sed 's/mcp_session_id/session/g' /tmp/r001/named/server.py > /tmp/r001/renamed/server.py

for d in named renamed; do mcp-migrate check --format json /tmp/r001/$d | python3 -c "
import json,sys; d=json.load(sys.stdin)
print(d['grade'], d['score'], [(f['rule'],f['line']) for f in d['findings']])"; done
```

```
C 75  [('R001', 2), ('R001', 3)]
A 100 []
```

Both files read `Mcp-Session-Id` off the request. The second one grades **A/100**.

`mcp-migrate 0.4.0`, `main` @ `cd55c27`. `files_scanned` is 1 in both cases, so this is not the "is this MCP?" gate from #89/#255 declining to scan — the file is scanned, R001 runs, and it finds nothing.

## Why

`r001_session_id_removed.py`:

```python
PY_RX = r"Mcp-Session-Id|mcp_session_id|MCP_SESSION_ID"
...
return [
    self.finding(self.MESSAGE, f, line, text)
    for f, line, text in project.search_code(PY_RX)
]
```

`search_code` yields a match only if the position it *begins* at is outside every `COMMENT` and `STRING` span. `Mcp-Session-Id` is not a valid Python identifier, so there is no position in a Python file where it can begin outside a string. The first alternative is dead.

Confirmed rather than argued: `search_code` falls back to unfiltered `search` when a file will not tokenize. Take the file that reports nothing and break its syntax —

```python
def handle(request):
    return request.headers.get("Mcp-Session-Id")

def broken(  # unclosed
```

— and R001 fires on line 2. Same line, same regex; the only thing that changed is whether the `STRING` span was computed. That isolates the cause to the span filter, not to the pattern or the scanner.

## What it means in practice

**The canonical form is missed.** `request.headers.get("Mcp-Session-Id")` with the result not bound to a name spelled `mcp_session_id` is invisible. `SESSION_HEADER = "Mcp-Session-Id"` followed by `request.headers[SESSION_HEADER]` is also invisible — I checked both.

**What is caught is often not a header access.** R001 fires on exactly one board server, `mcp-atlassian`, three times:

```
src/mcp_atlassian/servers/main.py:629   mcp_session_id = headers.get(b"mcp-session-id")
src/mcp_atlassian/servers/main.py:630   if mcp_session_id:
src/mcp_atlassian/servers/main.py:631   session_id_str = mcp_session_id.decode("latin-1")
```

All three are the identifier. Line 629 is a real header read, but it matched because of the name on the left of the `=`; the header string `b"mcp-session-id"` contributed nothing. Lines 630 and 631 are a truth test and a `.decode()` — not header access at all.

**Python and TypeScript disagree on the same code.** The TS path uses `search_wire` for the header literal precisely because the name lives in a string:

```ts
  return req.headers.get("mcp-session-id");
}
```

reports `R001` at line 2. The Python equivalent reports nothing. The comment in `_check_ts` says the TS side "cannot be an identifier-only match" *unlike the Python side*, so the asymmetry looks deliberate — but the Python side still carries `Mcp-Session-Id` in its pattern, which reads as an intent to match the header form too.

**The docs assume the header form is caught.** `cookbook/01-sessions-to-explicit-handles.md` describes the breakage as code "that reads `request.headers.get("Mcp-Session-Id")`", and the README's fixer walkthrough (line 340) uses:

```python
-    mcp_session_id = request.headers.get("Mcp-Session-Id")
```

That example works only because of the `mcp_session_id` on the left.

## Not a duplicate of #90

#90 was the opposite failure — R001 ran its *identifier* pattern through `search_wire` and reported `breaking` on a log string. Moving to `search_code` fixed that, correctly. This is the cost that move carried with it for the header-literal alternative, which needs a different treatment from the identifiers rather than the same one.

## Suggested direction

Split the Python path the way the TypeScript path already is: identifiers keep `search_code` (that is what #90 bought, and it should not be given back), and the header literal goes through `search_wire` anchored to an access — `headers[...]`, `.get(...)`, `.setdefault(...)`, and the `b"..."` bytes form that `mcp-atlassian` uses — so a docstring or `--help` string naming the header stays excluded but a real read is found.

The regression cost is worth measuring before committing to it: R001 is `breaking`, so anything newly caught costs a full 25 points and can move published board grades. Happy to re-run all 16 registry entries at their pinned `sha` against a candidate.

Found during a maintainer sweep on `main` @ `cd55c27`.

crazyshubham:
R001's Python path only matched the mcp_session_id/MCP_SESSION_ID identifiers through search_code, which skips every STRING token. Mcp-Session-Id is not a valid Python identifier, so the header-literal alternative could never match real code -- renaming a local variable away from mcp_session_id silently hid a breaking finding. Split the Python path the way the TypeScript path already is: identifiers stay on search_code, and the header literal is matched via search_wire, anchored to an actual access (.get/.setdefault/.pop/ bracket/dict-literal) so a docstring or log message naming the header still doesn't fire. Known gap: SESSION_HEADER holding the header name, then used as headers[SESSION_HEADER], is not caught -- that needs data-flow tracking, not a regex anchor. Fixes #260

## What kind of PR is this?

* [ ] Cookbook recipe -- adds/fills in a file in `cookbook/`
* [ ] Rule -- adds/changes a rule in `src/mcp_migrate/rules/`
* [ ] Fixer -- adds/changes a fixer in `src/mcp_migrate/fixers/`
* [ ] Board entry -- adds/updates a `registry/servers/*.yaml` file
* [ ] Something else

## Cookbook recipe

* [ ] Follows `cookbook/_TEMPLATE.md` (Rule/Fixer/Severity/Spec, What broke,
* [ ] Row added to (or moved out of "Stubs" in) `cookbook/README.md`

## Rule

* [ ] `severity` is `breaking`, `deprecated`, or `advisory`
* [ ] `spec_ref` names the spec section or SEP the rule enforces
* [ ] Fixtures added under `tests/fixtures/<rule-id>/`
* [ ] At least one test added and passing (`pytest`)
* [ ] `mcp-migrate rules` lists the new rule

## Fixer

* [ ] `rule_id` matches an existing rule's `id`
* [ ] `confidence` is `safe` only if the transformation cannot change
* [ ] Ambiguous shapes are left unchanged (`self.unchanged(source)`), not
* [ ] A round-trip/idempotency test: running the fixer twice changes
  nothing the second time
* [ ] `mcp-migrate fixers` lists the new fixer

## Board entry

* [ ] Generated with `mcp-migrate entry --repo owner/name`
* [ ] `notes:` edited to a real, one-sentence description
* [ ] `name` matches the filename


dheerajjha:
Merged as cbcd84934b9ed7ab63cb10d0c10fea113de78c9a. Thanks, @crazyshubham — the split mirrors the TypeScript path exactly, and the anchor list (including the `b"..."` bytes form `mcp-atlassian` uses) is the right level of caution.

GitHub returned a 502 mid-squash, so the commit landed on `main` but the PR was left open; closing it by hand. `main`'s tree is byte-identical to `4a734bb` (`c4d8a44`), so what shipped is exactly what was reviewed.

What I ran, on a real merge of `4a734bb` into `main` @ `d55fca5` (merge-base is `main` HEAD, so no stale-base artifact):

**The bug, before.** Issue #260's repro on `main`:

```
named:   C 75  files_scanned=1  [('R001', 2), ('R001', 3)]
renamed: A 100 files_scanned=1  []
```

**After.** Same two trees, merged tree:

```
named:   C 75  files_scanned=1  [('R001', 2), ('R001', 3)]
renamed: C 75  files_scanned=1  [('R001', 2)]
```

Line 2 is the actual `request.headers.get("Mcp-Session-Id")`, and it is one finding, not one per mention.

**The new tests fail without the rule change.** I checked out `tests/test_regressions.py` from this PR onto an otherwise-unmodified `main`:

```
FAILED test_header_read_fires_r001_even_when_the_local_is_not_named_mcp_session_id
FAILED test_header_read_via_dict_bracket_access_fires_r001
FAILED test_header_read_via_bytes_literal_fires_r001
3 failed, 68 passed
```

So they pin the behaviour rather than describing it.

**False positives.** The thing I most wanted to break, since #90 was the opposite failure. A file carrying the header name in a module docstring, a `#` comment, a `click.option(help=...)`, a `logger.debug`, a `raise ValueError`, a `%`-format string, and a dict whose *value* mentions it: **zero findings**. A second file with four genuine accesses — `.get("Mcp-Session-Id")`, `headers[b"mcp-session-id"]`, `.setdefault(...)`, and `{"Mcp-Session-Id": a}` — reports exactly four. Both via `--format json`.

**Full suite on the merged tree:** 732 passed (727 on `main` + the 5 added here).

**Board impact: none.** This is the part your issue asked to have measured before committing, and it needed measuring — R001 is `breaking`, so a new hit costs 25 points. I cloned all 18 `registry/servers/*.yaml` repos at their pinned `sha` and graded them with `main` and with the merged tree. All 18 grades on `main` reproduce their published entry exactly, and all 18 are **unchanged** by this PR — identical grade, score and per-rule finding counts.

`mcp-atlassian` is the interesting one: it stays `C/64` with `R001 ×3`. The new header pattern does hit line 629's `headers.get(b"mcp-session-id")`, but the identifier pass already claimed that line, and the dedup guard collapses it — one problem, not two. So no registry entry needed updating.

The `SESSION_HEADER = "Mcp-Session-Id"` → `headers[SESSION_HEADER]` gap you flagged in the PR description is still open and still correctly out of scope for a regex anchor; I confirmed it is missed. Worth its own issue if it ever shows up on a real server.

crazyshubham:
Thank you so much for the opportunity and for reviewing and merging my
contribution! I’m a final-year student and I’m very interested in
continuing to contribute to the project and learn from the community.

I’d be happy to work on more issues or improvements. If you have any
suggestions on areas I should explore or issues that would be suitable for
me to take up next, I’d really appreciate your guidance.

Looking forward to contributing more! 🙌

PR #268, opened 2026-09-13
docs: refresh README check transcript

### LINKED ISSUE #265: README's headline `check` transcript is stale: wrong version, wrong score, and a missing R021 row
The first thing a reader sees in the README is a worked `check` transcript. It no longer matches what the tool prints, and the numbers in it disagree with each other as well as with the code. Found during a maintainer sweep on `main` @ `20f3321`.

This is the same class of bug as `ea6d5b9` ("docs: correct the per-rule cap table, and pin both drifted numbers with tests") — a documented number with no test holding the docs to it — in the most prominent example in the repository.

## Reproduce

From a clean checkout of `main` @ `20f3321`, run the exact command the README prints:

```bash
pip install -e '.[dev]'
COLUMNS=200 mcp-migrate check tests/fixtures/fixer_roundtrip
```

(The wide `COLUMNS` only stops rich from wrapping the table; it does not change any value. `--format json` gives the same numbers.)

## What differs

**1. Version banner — README:26**

```
README:  mcp-migrate v0.2.0  ->  fixer_roundtrip
actual:  mcp-migrate v0.4.0  ->  fixer_roundtrip
```

`__version__` is `0.4.0`. The same stale banner appears a second time at **README:203** in the TypeScript example (`mcp-migrate v0.2.0  ->  my-ts-server`); I confirmed a TypeScript tree also prints `v0.4.0`.

**2. A missing finding row.** The transcript lists nine findings and stops at R016. The run produces ten — R021 fires on the same fixture and is absent from the README:

```
advisory    R021    server.py:41  Pins an older JSON Schema dialect; 2026-07-28 requires 2020-12 support.
```

**3. The summary line — README:49**

```
README:  Grade F (26/100)  3 breaking, 2 deprecated, 4 advisory
actual:  Grade F (27/100)  3 breaking, 2 deprecated, 5 advisory
```

`--format json` agrees with the actual line:

```json
{"version": "0.4.0", "grade": "F", "score": 27, "counts": {"breaking": 3, "deprecated": 2, "advisory": 5}}
```

## The score is not internally consistent either

Worth separating, because it shows `26` is not merely one release out of date — it does not correspond to the README's own finding list. Rules fired are R001, R017 (breaking), R006 (deprecated), R004, R005, R010, R016, R021 (advisory), and with `CAP_MULTIPLE = 1` each rule contributes its severity weight once:

```
25 + 25 + 8 + 3 + 3 + 3 + 3 + 3 = 73   ->  score 27   (actual, ten findings)
25 + 25 + 8 + 3 + 3 + 3 + 3     = 70   ->  score 30   (the nine findings the README lists)
```

So the README's own table implies 30, the tool says 27, and the README says 26. No rule set produces 26 from that list — it is left over from an older cap table or rule set.

## Why CI does not catch it

`tests/test_docs.py` has eight tests and they pin the things that drifted before: the rule table, the grade table, the cost/cap table, the pre-commit `rev:`, the JavaScript coverage sentence. None of them read the transcript blocks, so the version banner, the finding rows, and the `Grade F (...)` summary are unguarded. The suite is green on `main` with all three wrong.

## Suggested shape

The fixture is committed and the command is deterministic, so this can be pinned the same way the other numbers were — a test that runs `check tests/fixtures/fixer_roundtrip --format json` and asserts the README transcript's version banner, its severity counts, and its score match, reading them out of the result rather than restating them. That also covers the TypeScript banner at README:203 if it checks every `mcp-migrate v<x>` line against `__version__`.

I have not fixed it here because the right correction depends on whether the transcript should be regenerated wholesale (it would then also gain the R021 row and the reworded TypeScript sentence at README:206, which has drifted too: README says "Every rule reads it now, but", the tool now prints "TypeScript is read by every rule, but") or hand-patched to the three numbers above. That is an editorial call.

busseozgenoglu-cyber:
Refresh the README examples to match the current CLI output:

- update both version banners to the current package version
- add the missing R021 finding and correct the headline score/counts
- add regression coverage that derives the transcript expectations from `check --json` and verifies every README CLI banner against `__version__`

Fixes #265

Validation: not run locally in this environment; the regression test follows the existing `main(...)/capsys` test pattern and exercises the committed `fixer_roundtrip` fixture.


dheerajjha:
Merged as 2d9ed433a4056f6dac2a7d5e037560600b20b09f. Thanks, @busseozgenoglu-cyber — and thanks for saying plainly that you hadn't run it locally. That's more useful than a claim I'd have had to re-derive anyway. I ran it; everything you wrote holds.

Verified on a real merge of `f529fd6` into `main` @ `cbcd849` (a genuine three-way merge, not the PR alone — `main` had moved since your branch point when #267 landed). `main`'s tree is byte-identical to the merge I tested.

**The transcript now matches the tool, character for character.** I ran the exact README command against the merged tree:

```
$ COLUMNS=200 mcp-migrate check tests/fixtures/fixer_roundtrip

mcp-migrate v0.5.0  ->  fixer_roundtrip
2 Python files, 21 rules, spec 2026-07-28
...
advisory    R021    server.py:41  Pins an older JSON Schema dialect; 2026-07-28 requires 2020-12 support.

Grade F (27/100)  3 breaking, 2 deprecated, 5 advisory
```

Every row, the banner and the summary line agree with the README block as merged. The only differences are the README's deliberate `[... one block like this per rule that fired ...]` elision and its narrower wrapping of the two long rows — both intentional.

**The tests fail without the README change.** I checked `tests/test_readme_check_transcript.py` out from this PR onto an otherwise-unmodified `main`:

```
FAILED test_headline_check_transcript_matches_cli
FAILED test_readme_cli_banners_match_package_version
  Extra items in the left set: '0.2.0'
  Extra items in the right set: '0.5.0'
2 failed
```

**Full suite on the merged tree:** 734 passed (732 on `main` + the 2 added here).

One thing worth recording because it nearly bit this PR: #267 merged into `main` between your branch point and this merge, and it changes R001's Python matching. I re-derived the fixture's numbers on post-#267 `main` before merging — still `F`, `27`, `3 breaking / 2 deprecated / 5 advisory`, `R001` on lines 28 and 29. Line 28's `request.headers.get("Mcp-Session-Id")` is now matched by both the identifier pass and the new header pass, and the dedup guard collapses it, so the count held. Had it not, this transcript would have shipped stale on the day it was fixed.

Deriving the expectations from `--json` instead of restating them is the right call, and it matches `test_readme_precommit_rev_is_the_current_version` — so a version bump can't ship without the README moving with it, which is exactly the property that was missing.

I used `Refs #265` rather than `Fixes` in the squash message, but GitHub closed #265 anyway from the `Fixes #265` in the PR description. One item from that issue's body is still outstanding — the reworded TypeScript sentence — so I've reopened it with the details. Not a gap in this PR; the issue itself called that part an editorial decision.

PR #282, opened 2026-09-17
fix(rules): walk all regex matches in search_code and search_wire (#281)

### LINKED ISSUE #281: search_code() drops a whole line when the first match is inside a string, hiding real code matches
`search_code()` tests only the **first** regex match on a line. When that first match falls inside a string literal or comment, the whole line is dropped — including a genuine code match further along the same line.

Found by @charle-z during an adversarial pass on the R001 JavaScript port in #149, where it was correctly identified as a reason *not* to fold a shared-matcher cleanup into a language port. Filing it separately so it gets its own fix and tests.

## Where

`src/mcp_migrate/rules/base.py:92-101`

```python
rx = re.compile(pattern, flags)
for f in self.files:
    spans = self._spans_for(f)
    for i, line in enumerate(f.lines, start=1):
        m = rx.search(line)                                     # first match only
        if not m:
            continue
        if spans is not None and _in_content_span(i, m.start(), spans):
            continue                                            # skips the ENTIRE line
        yield f, i, line.strip()
```

The span check asks "is *this one* match inside a string?" and the `continue` answers "then there is nothing on this line" — which does not follow.

## Reproduce

```python
line = 'log("Mcp-Session-Id"); sid = req.headers[SESSION_ID]'
rx   = re.compile(r'Mcp-Session-Id|SESSION_ID')

[m.group(0) for m in rx.finditer(line)]   # ['Mcp-Session-Id', 'SESSION_ID']
rx.search(line).group(0)                  # 'Mcp-Session-Id'  -> in a string -> line dropped
```

The real header read is never reported.

## Why it matters

A false negative, so it fails in the quiet direction — but `search_code` is the matcher behind the identifier-based rules, and a missed `Mcp-Session-Id` in R001 is a **breaking** finding that a server never learns about. Per the priority order in the README this ranks below "never report a finding that isn't real", which is why it is `bug` and not urgent. It is still a clean bill of health that was never earned.

The realistic trigger is a string literal earlier on the same line as real code — logging a header name beside reading it, an `argparse` default, a dict literal mixing a label with a constant. Not common, not contrived either.

## Shape of the fix

Walk every match on the line rather than the first, and yield on the first match that starts **outside** a content span:

```python
for m in rx.finditer(line):
    if spans is not None and _in_content_span(i, m.start(), spans):
        continue
    yield f, i, line.strip()
    break
```

`break` keeps the existing one-finding-per-line contract, so no rule starts reporting twice for a line it used to report once.

`search_wire` (`base.py:103`) has the same single-`search` shape and should be checked in the same pass — its span set is narrower, so the trigger is rarer, but the logic is identical. **Check it rather than assuming**; I have not reproduced it there.

## Tests

A fixture with a string-then-code line for at least one identifier rule, asserting the code match is found. Worth one for `search_wire` too if it turns out to share the defect.

HeaTTap:
## What kind of PR is this

- [ ] Cookbook recipe
- [ ] Rule
- [ ] Fixer
- [ ] Board entry
- [x] Something else

## Anything else reviewers should know

Closes #281

Updates search_code and search_wire in base.py to use rx.finditer instead of rx.search so each match on a line is tested against content and prose spans. This prevents dropping a line when the first match is inside a comment or string literal and a real code or wire match exists later on the same line.

Added regression tests in tests/test_regressions.py covering search_code and search_wire for Python and TypeScript as well as R001 firing on string then code lines. All tests pass.


dheerajjha:
Ran this rather than read it, and the bug reproduces on `main` exactly as #281 describes:

```python
(d/"srv.py").write_text('log("Mcp-Session-Id"); sid = req.headers[SESSION_ID]\n')
list(load_project(d).search_code(r"Mcp-Session-Id|SESSION_ID"))
# main  -> []
# here  -> [(srv.py, 1, 'log("Mcp-Session-Id"); sid = req.headers[SESSION_ID]')]
```

One line, two matches, and on `main` the real `req.headers[SESSION_ID]` read is gone without a trace. That is the quiet direction, which is the bad direction for R001 — a missed `Mcp-Session-Id` is a breaking finding a server never hears about.

Suite goes 856 on `main` to 861 here, no failures, `validate_registry.py` clean at 18 entries.

The `break` is what I looked at hardest, since that is the piece carrying the one-finding-per-line contract. It holds: when `spans is None` you yield on the first match and leave, which is what `search` did; when every match on the line sits inside a span you yield nothing, which is also what it did. The only behaviour that changes is the case the issue is about. That is the right size for a matcher everything else sits on.

Fixing `search_wire` in the same shape was the correct call even though #281 only named `search_code` — it had the identical loop and would have been a repeat of this in a month.

Tests earn their place too: the TypeScript pair matters because the span machinery differs by language there, and the R001 case proves it end to end rather than only at the matcher. Merging.

Issue #269, opened 2026-09-13
R017's `safe` fixer rewrites -32002 inside comments, editing files that `check` reports as clean


dheerajjha:
`mcp-migrate check` grades a tree `A (100/100)` and exits `0`. `mcp-migrate fix --safe-only --write` on that same tree edits a file. The edit is inside a `#` comment, and it turns a true sentence into a false one.

Found during a maintainer sweep on `main` @ `2d9ed43`, while confirming #252.

## Reproduce

```bash
mkdir -p /tmp/div
cat > /tmp/div/server.py <<'PY'
# Migration note: the old -32002 resource not found code was replaced in 2026-07-28.
def handle():
    return {"ok": True}
PY

mcp-migrate check --format json /tmp/div | python3 -c "import json,sys; d=json.load(sys.stdin); print(d['grade'], d['score'], d['findings'])"
mcp-migrate check /tmp/div > /dev/null; echo "check exit: $?"

mcp-migrate fix /tmp/div --safe-only --write
cat /tmp/div/server.py
```

```
A 100 []
check exit: 0

  [R017/safe] line 1: resource-not-found error code -32002 -> -32602
1 file(s), 1 change(s): 1 safe, 0 flagged for human review
No findings remain.
Changes written.

# Migration note: the old -32602 resource not found code was replaced in 2026-07-28.
```

The comment now says the *old* code was `-32602`. `-32602` is the new code. The sentence was correct before the tool edited it and is wrong after.

`No findings remain.` is printed for findings `check` never reported.

## Why this is not #252, and not #105

**Not #252.** That one is about the *rule*: `RESOURCE_NOT_FOUND_RX` is lookahead-only, so it matches zero-width at column 0 and an **indented** comment escapes the comment-span test. In the repro above the comment starts at column 0, so the rule handles it **correctly** — `check` reports nothing, which is the right answer. The bug here is that the *fixer* has no comment handling at all, so it diverges from a rule that got it right. Fixing #252 will not fix this.

**Not #105.** That is fixers inserting comment lines into string literals. R017's fixer already handles strings correctly — `string_lines` protects them:

```python
HELP = """
Migration note: the old -32002 resource not found code was replaced.
"""
```

→ `Nothing to fix.` Strings are covered; comments are not.

Closest relative is #217 (closed), where R017's `safe` fixer rewrote `LEGACY_*` constants. Same shape: the fixer edits something it should not, and `safe` means nobody is reading the diff.

## Scope

[...]

PR #283, opened 2026-09-17
Fix search_code/search_wire dropping a line after an in-string first match

### LINKED ISSUE #281: search_code() drops a whole line when the first match is inside a string, hiding real code matches

HarshRajSinghania:
Fixes #281.

## Summary
`search_code()` and `search_wire()` only inspected the first regex match on a line. If that match started inside a comment or string, the whole line was skipped, including a later real code match.

## Implementation
Walk every match with `finditer` and yield on the first match that starts outside a content span, then `break`. That keeps the existing one-finding-per-line contract.

The same shape is applied to `search_wire`, which had the identical first-match-only loop.

## Testing
Ran (via `PYTHONPATH=src`):

```
python3 -m pytest tests/test_regressions.py::test_search_code_excludes_docstrings_comments_and_strings tests/test_regressions.py::test_search_code_finds_later_code_match_after_string_match tests/test_regressions.py::test_search_wire_finds_later_literal_after_prose_match tests/test_regressions.py::test_search_code_falls_back_to_plain_search_on_tokenize_failure tests/test_regressions.py tests/test_scan_complexity.py tests/test_rules.py -q
```

Results: new tests pass; `tests/test_regressions.py tests/test_scan_complexity.py tests/test_rules.py` — 162 passed.

Could not `pip install -e .` in this environment (PyPI mirror 502), so tests were run against the tree with `PYTHONPATH=src`.


dheerajjha:
Your read of this was right, and I want to be precise about how right. The `base.py` on this branch and the one I just merged in #282 are the same blob:

```
$ git rev-parse pr282:src/mcp_migrate/rules/base.py
a3385ed7823de3d5fa97748309a14ac3cec144c8
$ git rev-parse pr283:src/mcp_migrate/rules/base.py
a3385ed7823de3d5fa97748309a14ac3cec144c8
```

Byte for byte, arrived at independently, including the `break` that preserves one finding per line and the decision to fix `search_wire` in the same shape rather than leave it for later. That is not a coincidence so much as the fix being the obvious one once you have actually understood the span check, and you had.

#282 opened about half an hour ahead of this and carried wider tests, so it is the one that landed. That is the whole of it — no verdict on your diff, which is the same diff.

One thing worth having regardless, because it will cost you a CI cycle again otherwise. `tests/test_regressions.py` here does not parse:

```
File "tests/test_regressions.py", line 148
    'log("Mcp-Session-Id"); sid = req.headers[SESSION_ID]
    ^
SyntaxError: unterminated string literal
```

A literal newline landed inside the single-quoted string where `\n` was intended, in both new tests. That takes the module down at import, which is why all four jobs went red at once instead of one assertion failing. Red across every Python version, with no test name in the output, almost always means a parse or import error rather than a real failure — quickest check is `python -m ast` on the file before pushing. I suspect your local run passed because the editor kept what you typed and the paste into the commit did not.

Closing as superseded by #282.

If you want another, **#269** is open, unassigned, and the same muscle one layer up: R017's `safe` fixer rewrites `-32002` inside comments, so `fix` edits files that `check` reports as clean. Same question you just answered — is this match really code? — but on the writing side, where being wrong changes someone's file. Say the word and I will assign it.

PR #284, opened 2026-09-17
feat: port R001 to JavaScript

charle-z:
Part of #149: port R001 (`SessionIdRemoved`) to JavaScript.

- declare JavaScript coverage for R001
- reuse the existing identifier spellings while using a conservative JavaScript-only header matcher for `headers[...]`, optional chaining, `get`/`set`/`delete`, and `setHeader`
- deduplicate identifier/header findings on the same line
- keep the existing TypeScript matcher and `base.py` unchanged
- update JavaScript coverage docs/regressions from 3/21 to 4/21 and move the unported-rule sentinel cases to R019
- add focused JavaScript tests covering identifiers, header APIs, comments/help strings, conservative false-positive boundaries, CommonJS, and a TypeScript regression guard

Validation:
- `python -m pytest -q` — 879 passed
- `git diff --check` — clean

#281 remains intentionally separate; this PR does not change `search_code()` first-match behavior.


dheerajjha:
Reviewed this by running the matcher against cases you did not write, because a language port is exactly where false positives get born, and on this rule a false positive is worse than a miss — it tells someone their working server is broken.

Ten probes through `SessionIdRemoved().check()` on real single-file JS projects:

```
ok  headers["mcp-session-id"]            findings=1
ok  req.headers?.get("mcp-session-id")   findings=1
ok  res.setHeader("mcp-session-id", …)   findings=1
ok  const mcpSessionId = mint()          findings=1
ok  CommonJS require + headers[…]        findings=1
ok  const customHeaders = {"mcp-session-id": 1}   findings=0
ok  myHeaders["mcp-session-id"] = 1              findings=0
ok  // mcp-session-id is gone in the new spec    findings=0
ok  const config = {"mcp-session-id": true}      findings=0
ok  obj.mySetHeader("mcp-session-id", x)         findings=0
```

Ten for ten, and the five that must stay silent are the ones I care about. `(?<![\w$])` in front of `headers?` and `setHeader` is doing that work — without it `customHeaders` and `mySetHeader` both fire and the rule starts inventing findings in ordinary application code. Calling the matcher "conservative" in the comment and then actually making it narrower than the TypeScript one is the right instinct; an object literal with an `mcp-session-id` key is not a header read and this correctly declines to guess.

Routing identifiers through `search_code` and header spellings through `search_wire` is also right, and subtle enough to be worth naming: header names live in string literals, so `search_wire` has to see them, while an identifier in a string is prose and must not fire. The dedupe by `(path, line)` then stops one line matching both passes from being reported twice.

Suite: 856 on `main`, 884 with this and #282 together, 0 failures, registry clean at 18 entries. Your 879 was against `main` before #282 landed and the arithmetic lines up exactly, so nothing collided.

Separately, thank you for splitting #281 out during the adversarial pass instead of folding the shared-matcher cleanup in here. That was the correct instinct and it is why that fix got its own tests and its own review — it just merged as #282. Keeping `base.py` untouched in this PR made both of them reviewable.

JavaScript goes 3/21 to 4/21. Merging.

PR #291, opened 2026-09-18
fix(badges): prune withdrawn registry endpoints

### LINKED ISSUE #290: Withdrawing a board entry leaves its badge files serving the old grade
Withdrawing a board entry leaves its badge files behind, and you find out from a test failure that does not say which files to delete.

## What happened

I removed `registry/servers/serena.yaml` and regenerated everything:

```
$ python scripts/render_board.py    ; rendered 17 entries
$ python scripts/render_badges.py   ; Wrote 29 badge endpoint(s) to docs/badge/ (+ unknown.json)
$ python scripts/validate_registry.py ; checked 17 entries, 0 problems
```

Three green runs, and `docs/badge/` still held two files for the entry that no longer exists:

```
docs/badge/serena.json
docs/badge/oraios/serena.json
```

Both still serving the withdrawn grade:

```json
{ "label": "MCP 2026-07-28 (0.4.0)", "message": "D", "color": "orange" }
```

The suite does catch it — `tests/test_badges.py::test_the_committed_endpoints_are_what_the_renderer_produces_today` goes red. But what it prints is a list diff:

```
At index 27 diff: PosixPath('serena.json') != PosixPath('sooperset/mcp-atlassian.json')
Left contains one more item: PosixPath('wrale/mcp-server-tree-sitter.json')
Use -v to get more diff
```

That says two lists are unequal at index 27. It does not say "delete `serena.json`". I removed the `oraios/` one, re-ran, got a near-identical diff, and had to go find the flat one separately.

## Why it is worth fixing rather than living with

`render_badges.py` says at line 165:

> regenerating -- a stale badge is the bug being fixed here

A badge endpoint that keeps serving `D` for a project we deliberately stopped making claims about is exactly that bug, and the renderer walks past it. Shields.io will happily keep serving those URLs to anyone who embedded them.

## What done looks like

`render_badges.py` removes badge files that no entry produces any more, the same way it writes the ones that are current — it already knows the full set it should have produced, so the orphans are what is left over.

Two small things that go with it:

- **Prune empty owner directories.** `docs/badge/oraios/` was left behind as an empty directory after its only file went.
- **Make the test say what to do.** Comparing sorted lists with `==` gives a positional diff; asserting on the set difference instead would print `unexpected: {'serena.json', 'oraios/serena.json'}`, which is the sentence someone needs. The test is right to exist and right to fail — it just answers a different question than the one you are asking when it fails.

## Scope

`scripts/render_badges.py` and `tests/test_badges.py`. No new dependencies. To reproduce, delete any file in `registry/servers/`, run the three scripts above, and look at what is left in `docs/badge/`.

Be careful that pruning only ever removes files under `docs/badge/` that the renderer owns — `unknown.json` is produced deliberately and must survive, and nothing outside that directory should be touched by a rendering script.

motodriver:
## Summary

Closes #290.

- remove renderer-owned stale JSON endpoints after each badge render
- prune empty owner directories left after their last endpoint is removed
- make the committed-endpoint test name unexpected and missing paths
- cover withdrawn-entry cleanup and non-JSON asset preservation

## Validation

- `uv run python -m pytest tests/test_badges.py -q` — 34 passed
- `uv run python -m pytest -q` — 886 passed
- `git diff --check` — passed

## Scope and limits

The cleanup only removes `*.json` endpoint files below `docs/badge/`; it leaves non-JSON assets and all paths outside that directory untouched. I did not modify registry entries.

## AI assistance

TRAE CLI assisted with investigation and implementation. I reviewed the final diff and ran the listed checks.


dheerajjha:
Ran this against the exact scenario that produced the issue — withdrawing an entry and regenerating — rather than reading the diff.

```
$ rm registry/servers/mcp-obsidian.yaml
$ python scripts/render_badges.py
Wrote 27 badge endpoint(s) to docs/badge/ (+ unknown.json)

badge files: 30 -> 28
obsidian badges left: 0
empty owner dirs left: 0
```

Both of them, which is the part I care about — the entry produces a flat `mcp-obsidian.json` *and* an owner-scoped `MarkusPfundstein/mcp-obsidian.json`, and finding the second one separately after removing the first is precisely how this cost me two cycles. And `docs/badge/keepme.txt` survived a render, so the `*.json` restriction holds.

The scoping is right in a way that matters more than the feature: `_owned_badge_files` globs under `out` only and takes `is_file()`, `expected` is seeded with `unknown.json` so the deliberate one is never an orphan, and the directory sweep is sorted by `len(path.parts)` reversed so a nested empty dir goes before its parent. A pruning step in a rendering script is exactly where an over-broad delete would be unforgivable, and this one cannot reach outside `docs/badge/`.

Changing the test from a positional list compare to `unexpected` / `missing` sets is the half of this I would have been happiest with on its own. The old failure said two lists differed at index 27; the new one names the file.

892 with this on `main` (890 before), registry clean at 17 entries.

Thanks for the note about TRAE — it doesn't change how I read a diff, but I'd rather know than not.

PR #292, opened 2026-09-18
Port R012 logging/setLevel detection to JavaScript

### LINKED ISSUE #286: Port R012 (removed logging/setLevel) to JavaScript
Part of #149. JavaScript coverage is **4 of 21** rules today (R001, R006, R017, R021); this adds R012.

R012 is `breaking`: `logging/setLevel` was removed from the spec, so a JavaScript server still implementing it is broken against 2026-07-28 and currently gets told nothing at all.

The Python and TypeScript paths already split the work the way the port needs: `search_code` for the `SetLevelRequest*` identifiers, `search_wire` for the `logging/setLevel` method string — the comment at `r012_logging_set_level_removed.py:41` explains why the wire half cannot go through `search_code`. Read that comment before you start; it is the whole trick.

## Why this one is a good place to start

R012 matches **wire strings**, not language idioms. The method names it looks for are the same three characters in Python, TypeScript and JavaScript, because they are JSON-RPC method names on the wire rather than anything the language spells differently. That makes the port small and the risk of inventing a false positive low — which is not true of every rule in #149.

It is also deliberately **not** one of the rules named in #89 as gating too loosely. Please do not widen the gate while you are in here; if you think it is too loose, that belongs in #89 rather than in a port.

## The worked example

#284 ported R001 and is the template. Read that diff first — it is short, and the review on it spells out the reasoning.

What it did, and what this should do:

1. Add `"javascript"` to `languages` on the rule.
2. Add a `_check_js` branch in `check()`, alongside the existing `_check_ts`.
3. Reuse the wire pattern; keep the JavaScript matcher **no wider** than the TypeScript one. #284's note is worth copying verbatim in spirit: an object literal that happens to contain the string is not the same as code that speaks it.
4. Dedupe findings per `(path, line)` so one line matching two passes is reported once.
5. Add focused tests in `tests/test_javascript.py` — fires, and the false-positive boundaries that must stay silent. Both halves, please; the silent half is the one that matters.

## What "done" looks like

```
$ uv run pytest -q          # green, and up by however many tests you added
```

And the README line has to move, because a test enforces it:

```
tests/test_docs.py:396  test_readme_javascript_coverage_matches_the_code
  "4 of 21 rules (R001, R006, R017, R021) read JavaScript"
```

That test reads `languages` off the rules and compares it to the README, so it goes red the moment you add the language and forget the prose. That is intentional — it is the repo keeping its own docs honest, not a chore.

## House rules

- Tests never hit the network, and fixtures are written inline, never captured from a real repository.
- No attribution lines in the commit or PR.
- Say in this thread that you are taking it, so two people do not write the same patch — that happened last night with #282 and #283 and cost someone an evening.

hizlidepoo:
Closes #286.

## What changed
- add JavaScript coverage for R012 using the existing TypeScript code matcher and wire-method matcher
- deduplicate JavaScript findings by `(path, line)` when both passes match the same line
- add focused JavaScript positive/false-positive regression tests
- update JavaScript coverage documentation and existing coverage/config regressions to 5/21

## Validation
- `uv run pytest -q tests/test_javascript.py tests/test_docs.py::test_readme_javascript_coverage_matches_the_code`: 20 passed
- `uv run pytest -q`: 894 passed
- `git diff --check`: PASS
- Ruff: NOT VERIFIED locally (`ruff` executable is not installed in the project environment)


dheerajjha:
Probed this with cases you didn't write, because a port is where a matcher quietly gets wider than the one it copied.

```
ok  server.setRequestHandler(SetLevelRequestSchema, …)   findings=1
ok  msg.method === "logging/setLevel"                    findings=1
ok  {"logging/setLevel": onSetLevel}                     findings=1
ok  // logging/setLevel was removed in 2026-07-28        findings=0
ok  logger.setLevel("debug")                             findings=0
ok  const m = "logging/setLevelLatency"                  findings=0
ok  const setLevel = (n) => n + 1                        findings=0
```

Seven for seven. Two of the silent ones are the interesting ones. `logger.setLevel("debug")` is in every JavaScript codebase that has ever logged anything, and it stays quiet because you kept the matcher on the SDK schema name rather than the bare method — reusing the TypeScript matcher rather than writing a looser JavaScript one was the right instinct and it is the reason this is safe.

`logging/setLevelLatency` is the other one. That is #88's bug, and it stays silent because `wire_method()` bounds the name at the definition rather than at the call site. Nothing for you to have done there, but it is worth knowing you inherited a guard rather than got lucky.

Suite: 890 on `main`, 892 with #291, 896 with both. Your 894 was against `main` before #291 landed and the arithmetic matches exactly, so nothing collided. Registry clean at 17.

README goes to 5 of 21 and `test_readme_javascript_coverage_matches_the_code` is green on it — that test reads `languages` off the rules, so it would have caught a coverage line you forgot rather than trusting you to remember.

On ruff: don't worry about it. It isn't in this project's dev dependencies, so `uv run ruff` fails here for me too — CI is where it runs. Flagging what you couldn't verify instead of quietly implying you had is the right habit and I'd rather see it.

R012 is `breaking`, so a JavaScript server still implementing `logging/setLevel` now hears about it. Merging.

PR #293, opened 2026-09-20
fix(rule): port R019 detection to JavaScript

### LINKED ISSUE #288: Port R019 (removed tasks/list and blocking tasks/result) to JavaScript
Part of #149. JavaScript coverage is **4 of 21** rules today (R001, R006, R017, R021); this adds R019.

R019 is `breaking`: `tasks/list` is gone and `tasks/result` no longer blocks, so a JavaScript server polling the old way is broken and hears nothing about it.

`r019_tasks_polling_replaces_blocking_result.py` has the same two-matcher shape as R012 and R013, and its comment at line 53 records why the wire half needs `search_wire`. R019 is also where #284 moved the unported-rule sentinel cases, so you will likely be touching those tests — check what they assert before changing them.

## Why this one is a good place to start

R019 matches **wire strings**, not language idioms. The method names it looks for are the same three characters in Python, TypeScript and JavaScript, because they are JSON-RPC method names on the wire rather than anything the language spells differently. That makes the port small and the risk of inventing a false positive low — which is not true of every rule in #149.

It is also deliberately **not** one of the rules named in #89 as gating too loosely. Please do not widen the gate while you are in here; if you think it is too loose, that belongs in #89 rather than in a port.

## The worked example

#284 ported R001 and is the template. Read that diff first — it is short, and the review on it spells out the reasoning.

What it did, and what this should do:

1. Add `"javascript"` to `languages` on the rule.
2. Add a `_check_js` branch in `check()`, alongside the existing `_check_ts`.
3. Reuse the wire pattern; keep the JavaScript matcher **no wider** than the TypeScript one. #284's note is worth copying verbatim in spirit: an object literal that happens to contain the string is not the same as code that speaks it.
4. Dedupe findings per `(path, line)` so one line matching two passes is reported once.
5. Add focused tests in `tests/test_javascript.py` — fires, and the false-positive boundaries that must stay silent. Both halves, please; the silent half is the one that matters.

## What "done" looks like

```
$ uv run pytest -q          # green, and up by however many tests you added
```

And the README line has to move, because a test enforces it:

```
tests/test_docs.py:396  test_readme_javascript_coverage_matches_the_code
  "4 of 21 rules (R001, R006, R017, R021) read JavaScript"
```

That test reads `languages` off the rules and compares it to the README, so it goes red the moment you add the language and forget the prose. That is intentional — it is the repo keeping its own docs honest, not a chore.

## House rules

- Tests never hit the network, and fixtures are written inline, never captured from a real repository.
- No attribution lines in the commit or PR.
- Say in this thread that you are taking it, so two people do not write the same patch — that happened last night with #282 and #283 and cost someone an evening.

motodriver:
Closes #288.

## Summary
- port `R019` to JavaScript by letting the rule reuse the existing TypeScript detection path
- add JavaScript regression coverage for removed `tasks/list` / `tasks/result` method strings and schema-name references
- update the README JavaScript coverage note to include `R019`, and keep the coverage test aligned with the live rule set

## Validation
- `uv run pytest tests/test_typescript.py -k "r019 or javascript" -q`
- `uv run pytest tests/test_docs.py -k javascript_coverage -q`
- `uv run pytest -q`
- `git diff --check`

## Notes
- Scope is intentionally limited to the JavaScript port for `R019`; no fixer behavior or broader rule coverage changed.
- AI assistance was used to help draft and implement this change, then reviewed and verified by a human before submission.


dheerajjha:
Ran it rather than read it. Suite goes 896 → 899, all green, and the three new tests are the three you added.

The port itself is exactly the shape #288 asked for — `languages` gains `"javascript"`, `check()` routes it to the existing `_check_ts`, and nothing about the matcher widens. Confirmed the detection is live rather than just compiled:

```
$ printf 'export function listTasks() {\n  return { method: "tasks/list" };\n}\n' > s.js
$ mcp-migrate check .
breaking  R019  s.js:2  References the removed tasks/list or blocking …
```

And `test_r019_ignores_javascript_comment_only_mentions` is the right negative to have picked. It is the one that actually exercises the port's risk: if `.js` did not route to `_ts_spans`, `search_wire` would fall back to unfiltered matching and that comment would fire. A JS mirror of every TS negative would be duplication — this one is not.

**The sentinel migration is the part I looked at hardest, and three of the four are right.** R019 was carrying the "rule that cannot read JavaScript" role, so porting it means that role has to move, not disappear. You moved it to R002 and swapped the fixture to `const sessions = new Map();`, which I checked is a live R002 finding rather than a near-miss:

```
$ printf 'const sessions = new Map();\n' > s.ts && mcp-migrate check .
breaking  R002  s.ts:1  Looks like per-connection state held in process …
```

So `--rule R002` over a `.js` file holding that line really is "a genuine bug sat there and nothing looked at it", which is the property those tests existed to pin. That is careful work and it is the half I expected to find broken.

**The fourth one lost its point.** `test_javascript_with_no_covered_finding_exits_zero_not_unscannable` used to hold a real breaking pattern that no ported rule caught, and asserted that partial coverage still exits 0 rather than 2. You replaced it with `class ListTasksRequester { send() {} }` — described accurately in your new comment as a near-miss the rule ignores — so it now asserts that a clean file exits 0. That is true, boring, and already covered elsewhere. Both exit 0, but only one of them is testing anything:

```
const sessions = new Map();              -> exit 0   (real R002 bug, no ported rule reads it)
class ListTasksRequester { send() {} }   -> exit 0   (nothing to find)
```

The fixture had to change — R019 catches the old one now — but the replacement wants to keep the "real uncaught bug" property, not drop it. Use `const sessions = new Map();` there too, with no `--rule`: all six JS rules run, none fires, and the assertion goes back to meaning what its name says. I verified that exits 0 on your branch. One line, and then this is good to go.

Two corrections on my side, both mine to own:

- #288 told you to put tests in `tests/test_javascript.py`. That was wrong — #284, which the same issue names as the worked example, created `tests/test_r001_javascript.py` and left `test_javascript.py` for the scanner-wiring tests it was written for. Putting R019's JavaScript cases next to its TypeScript ones is a defensible third answer and I am not going to move them.
- #288 says "4 of 21 (R001, R006, R017, R021)". R012 landed in between, so main was already at 5 and you correctly took it to 6. Generalising the `test_readme_javascript_coverage_matches_the_code` docstring instead of re-pinning the literal is the right call — it was stale by one port before you touched it.

